PR #171, opened 2016-02-23
Add preliminary QCow2 image support. Supports backing files only with…

crawlingChaos:
… absolute paths. No compression, encryption, or snapshots.

Autodetects qcow2 images vs. raw images. Useful for creating hierarchies of parent/child disks to isolate changes to the disk by software and reducing disk usage.

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -103,10 +103,9 @@ _AM_AUTOCONF_VERSION(m4_defn([AC_AUTOCONF_VERSION]))])
 # configured tree to be moved without reconfiguration.
 
 AC_DEFUN([AM_AUX_DIR_EXPAND],
-[dnl Rely on autoconf to set up CDPATH properly.
-AC_PREREQ([2.50])dnl
-# expand $ac_aux_dir to an absolute path
-am_aux_dir=`cd $ac_aux_dir && pwd`
+[AC_REQUIRE([AC_CONFIG_AUX_DIR_DEFAULT])dnl
+# Expand $ac_aux_dir to an absolute path.
+am_aux_dir=`cd "$ac_aux_dir" && pwd`
 ])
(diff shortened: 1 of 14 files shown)


incognitte:
I tested and qcow2 files converter from raw (using qemu-img) mount and work properly, however if the qcow2 image is compressed (qemu-img -c) then DOSBox-x doesn't recognize it.

crawlingChaos:
> I tested and qcow2 files converter from raw (using qemu-img) mount and work properly, however if the qcow2 image is compressed (qemu-img -c) then DOSBox-x doesn't recognize it.

Compressed images are not supported. I never got around to writing the code for it.

rderooy:
If I try to imgmount a qcow2 file in DOSBox-X, it only works if I specify ``-fs none``.

Note how the first imgmount command fails in the screenshot. The second imgmount, confusingly also complains, but DOES mount the image using a drive number.

Lastly, to prove the image DOES work, I boot from it.

The issue with not being able to mount it a C: in the dosbox-x integrated DOS, makes it more difficult to transfer files into and out of the image.

guest os_000

```
$ qemu-img info hdd.qcow2 
image: hdd.qcow2
file format: qcow2
virtual size: 500 MiB (524288000 bytes)
disk size: 1.62 MiB
cluster_size: 65536
Format specific information:
    compat: 1.1
    lazy refcounts: false
    refcount bits: 16
    corrupt: false
```
The image was created with ``qemu-img create -f qcow2 hdd.qcow2 500M``, after which I partitioned and formatted it in DOSBox-X.

rderooy:
Doing a little bit of tracing, since there are 2 locations in the code that can generate that error message. In both occasions the error message is caused by line 4420 of dos_programs.cpp

PR #172, opened 2016-02-23
Getting ready to handle relative paths for backing files.

crawlingChaos:
Really minor changes, inlcuding passing the original file name into the QCow2Image constructor to use as a jumping off point for relative paths in backing files.

Fixed my boneheaded mistake in read_header so that we ensure an empty header is returned in failure cases.

### DIFF
--- a/include/qcow2_disk.h
+++ b/include/qcow2_disk.h
@@ -52,7 +52,7 @@ public:
 	
 	static QCow2Header read_header(FILE* qcow2File);
 
-	QCow2Image(QCow2Header qcow2_header, FILE *qcow2File);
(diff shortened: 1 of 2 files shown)


joncampbell123:
Before I merge... I made a minor C++ correction to your new code in my branch. GCC 4.6.3 does not recognize constant assignments to members of a C++ class within the declaration.

crawlingChaos:
Sure, I can go ahead and make the changes, but just want to be clear: I need to move all constant definitions into the cpp file? What about magic, which needs to be public? The compiler complains if I try to move it.

Sorry my C++ skills haven't been used in nearly 15 years.
Running gcc 4.9.2 here.

crawlingChaos:
Nevermind, I figured it out.  :)

PR #253, opened 2017-04-07
fix build on macOS

martinlindhe:
The revert fixes ./configure on macOS, which failed like this:

```
...
checking that generated files are newer than configure... done
configure: error: conditional "LINUX" was never defined.
Usually this means the macro was only invoked conditionally.
```

Also, that patch didn't make much sense to me.

Working on another patch to fix compilation on macOS

EDIT: pushed a second commit that fully fixes compilation on macOS.

### DIFF
--- a/configure
+++ b/configure
@@ -629,8 +629,6 @@ LIBOBJS
 HAVE_WINDRES_FALSE
 HAVE_WINDRES_TRUE
 WINDRES
-LINUX_FALSE
-LINUX_TRUE
 CXXCPP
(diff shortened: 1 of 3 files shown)


martinlindhe:
Upon closer inspection, I note there's been some previous osx/macOS compile fix PR:s

However, since https://github.com/joncampbell123/dosbox-x/commit/8cb3c035f82cec27f0bf3e54c9d04d58a239e659, the build on osx would again have been broken

PR #254, opened 2017-04-07
Fullscreen scale fix

martinlindhe:
This PR fixes sdl setup to enable OpenGL, which other parts of the code assumed to be around. Because sdl didn't get the SDL_OPENGL flag, later probe of max texture size always returned 0, which made scaling not working.

Using dosbox.conf settings like the following now uses the full screen with correct aspect ratio.
Without this patch, there is a lot of black border around the video in full screen mode.

```
[sdl]
fullscreen=true
fullresolution=1280x800  # your native resolution
output=opengl

[render]
aspect=true
```

Tested on macOS and Fedora 25. In order to compile on macOS, one also need #253 

Regarding sdlmain.cpp. It is a mess. I considered doing more major cleanup of it but there is just too much of it. Especially, there is not a central place to control screen resolution changes, but calls to SDL_SetVideoMode is thrown all over the place, with mixed amount of flags set. While debugging I noticed that video mode was changed maybe 7 times just while BIOS booted. It could be cleaned up to offer a better experience.

Also a unrelated commit sneaked in, which fixes file encoding to UTF8 for some Korean text

Before:
before_ultima

With this PR:
after_ultima

### DIFF
--- a/src/dos/dos_misc.cpp
+++ b/src/dos/dos_misc.cpp
@@ -79,7 +79,7 @@ static Bitu INT2F_Handler(void) {
 	for(Multiplex_it it = Multiplex.begin();it != Multiplex.end();it++)
 		if( (*it)() ) return CBRET_NONE;
    
-	LOG(LOG_DOSMISC,LOG_ERROR)("DOS:Multiplex Unhandled call %4X",reg_ax);
+	LOG(LOG_DOSMISC,LOG_ERROR)("DOS:INT 2F Unhandled call AX=%4X",reg_ax);
 	return CBRET_NONE;
 }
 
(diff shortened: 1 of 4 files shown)


joncampbell123:
There's a lot this code inherited that's very messy and disorganized, yes :)

martinlindhe:
Yea, it shows. Many many different coding styles and conventions fighting with eachother :)

joncampbell123:
This is why I gradually rewrite portions as I go. I would go faster, but a hasty rewrite would just break everything and/or make it worse.

martinlindhe:
If I get in the right mood I might have a deeper look at sdlmain.cpp, and try to build something from scratch with bits and pieces from it. No promises though.

PR #272, opened 2017-05-24
Add support for DOS's CPM entry point

AlanStern:
DOSBox does not support the legacy CPM entry point in DOS.  This adds the missing support.  Believe it or not, people still run programs that use this mechanism.


sikthehedgehog:
This gets me wondering: is the A20 line toggle emulated? Since the CP/M entry point is precisely the one thing that tends to trip it.

joncampbell123:
The A20 gate is toggled. There are two modes. The first is backwards compatible with DOSBox-X where only the first 64KB after the 1MB boundary is remapped to fake the A20 mapping. The other is a full emulation of A20 where the 20th bit is always masked off if gated.

joncampbell123:
Do you mind if I boot up an MS-DOS bootdisk to confirm that your new code sets up the CPM service request address the way your code does?

joncampbell123:
Also, can you provide me an EXE or COM executable that uses the CPM service address so I can confirm this change works?

joncampbell123:
Weird, here's what I see from a Windows 95 bootdisk. HIMEM.SYS is obviously loaded and A20 is enabled or else this pointer into the HMA wouldn't work.

At PSP:0x05:   CALL F01D:FEEE

At F01D:FEEE:   RET

I'm not sure how a far call and a near return is supposed to work.

I rebooted again, and used the step-by-step mode to prevent HIMEM.SYS from loaded, and the far call still points at F01D:FEEE

At F01D:FEEE:

0000      ADD [BX+SI],AL
...and so on

This is pure DOS mode (bootdisk, not a DOS box). Obviously Windows 95 is not a good test bed for this sort of thing, so I'll next boot up MS-DOS 6.22 and then MS-DOS 5.0

joncampbell123:
MS-DOS 6.22 has the same weird pointer at PSP:0x05.

joncampbell123:
MS-DOS 5.0 has the same weird pointer at PSP:0x05. HMA is disabled in my setup, and it points to what is apparently gibberish code.

Where did you see otherwise? Or, what am I doing wrong?

joncampbell123:
I'm not clear on how CP/M compatible DOS programs interact with DOS. Documentation is hard to find. Can you clarify?

sikthehedgehog:
I was reading that if DOS is loaded high (which means into the segment at FFFF) that then the CP/M entry point will still work with A20 enabled, especially since only programs that can only use 64KB would be using that kind of call (and wouldn't try to touch the HMA).

...also apparently they messed up with the entry point from DOS 2 onwards (target address is two bytes too early it seems?), so I wonder how many programs even use it (since it'd mean they'd likely only work on DOS 1). Then again I guess directories probably messed them up already (with paths becoming longer than usual).

joncampbell123:
> also apparently they messed up with the entry point from DOS 2 onwards (target address is two bytes too early it seems?),

Hey, you're right about that:

From Windows 95 bootdisk:

At PSP:0x05

    CALL F01D:FEEE

At F01D:FEEE:

    C3           RET
    00            (invalid)

At F01D:FEF0 (two bytes later):

    JMP         00C9:0FE4

If I decompile where INT 21h points at, I also get:

At INT 21h vector entry point 0251:04A0:

    JMP         00C9:0FE4

That's interesting. I'll verify that with MS-DOS 6.22 and MS-DOS 5.0.

joncampbell123:
MS-DOS 6.22 bootdisk:

Same thing, both INT 21h and the JMP at the call site point to the DOS kernel at 000F:40D3.

However if HIMEM.SYS is not loaded, MS-DOS 6.22 in DOSBox-X does not bother to disable A20 (and DOSBox-X has it enabled by default) so the default configuration would point to zeroed and uninitialized memory and crash. It would work on actual hardware since most systems boot with A20 disabled from the BIOS onward.

joncampbell123:
My MS-DOS 2.1 bootdisk is even weirder:

At PSP:0x05

    CALL    F01D:FEEE

Again, this only works if you've instructed DOSBox-X to boot MS-DOS with A20 disabled.

At F01D:FEF0:

    JMP     00E3:0B14

At 00E3:0B14:

    POP AX               ; offset
    POP AX               ; segment
    CS: POP [02D1]  ; ????
    PUSHF
    CLI
    PUSH AX             ; put segment back
    CS: PUSH [02D1]; ???
    ...

DEBUG.COM in this version appears to hook INT 21, but if you follow the FAR JMP it executes, the original vector is in a similar location:

At 00E3:0B0B:

    CMP    AH,57
    NOP
    JBE     0B2B
    MOV   AL,00
    IRET

    ..(CPM entry point here, which may fall through to B2B)..
    ..(INT 21h stuff)..

So I get that in the above code, MS-DOS 2.1 removes the far return address from the stack, but saves the segment value.

joncampbell123:
Finally, MS-DOS 1.10:

At PSP:0x05:

    CALL    F01D:FEF0

Note that the 2 byte error is gone!

At F01D:FEF0:

    JMP      00BF:001D

Note that INT 21h points to 00BF:0015

At 00BF:0015:

    CMP    AH,2E
    JBE     0033
    MOV   AL,0
    IRET

Immediately following at 00BF:001D:

    POP   AX
    POP   AX
    CS: POP [1690]
    PUSHF
    ...

Interesting.

So, how exactly is one supposed to use the CP/M entry point and how do you call it?

AlanStern:
On Wed, 24 May 2017, Jonathan Campbell wrote:

> Also, can you provide me an EXE or COM executable that uses the CPM
> service address so I can confirm this change works?

DEMO.COM is attached.  It uses the CPM interface to print a one-line 
message and then exit.


On Wed, 24 May 2017, Jonathan Campbell wrote:

> Weird, here's what I see from a Windows 95 bootdisk. HIMEM.SYS is
> obviously loaded and A20 is enabled or else this pointer into the HMA
> wouldn't work.
> 
> At PSP:0x05:   CALL F01D:FEEE
> 
> At F01D:FEEE:   RET
> 
> I'm not sure how a far call and a near return is supposed to work.
> 
> I rebooted again, and used the step-by-step mode to prevent HIMEM.SYS
> from loaded, and the far call still points at F01D:FEEE
> 
> At F01D:FEEE:
> 
> 0000      ADD [BX+SI],AL
> ...and so on
> 
> This is pure DOS mode (bootdisk, not a DOS box). Obviously Windows 95
> is not a good test bed for this sort of thing, so I'll next boot up
> MS-DOS 6.22 and then MS-DOS 5.0


On Wed, 24 May 2017, Jonathan Campbell wrote:

> MS-DOS 5.0 has the same weird pointer at PSP:0x05. HMA is disabled in 
> my setup, and it points to what is apparently gibberish code.
> 
> Where did you see otherwise? Or, what am I doing wrong?

You aren't doing anything wrong.  What you see is expected, even if it 
isn't quite correct.

The far pointer F01D:FEEE points to the real address 1000BE.  As Javier
points out, this value is too small by 2; it really should be 1000C0.  
Since DOS programs run with A20 disabled (by default), the system
interprets this as a pointer to real address 000C0, which is where the
CPM service handler resides.  (What you see instead at 000BE is 
gibberish, part of a different interrupt vector.)

My patch sets up the correct far pointer (F01D:FEF0) at PSP:0005.


On Wed, 24 May 2017, Jonathan Campbell wrote:

> I'm not clear on how CP/M compatible DOS programs interact with DOS. 
> Documentation is hard to find. Can you clarify?

The interface is very simple.  The function value is stored in the CL
register instead of AH.  Other registers are set the same as they would
be for a DOS INT 21h system call.  Then the program executes a near
jump to PSP:0005.  The interface only supports CL values that are <=
0x24 (I think these were the only functions provided by CPM).


On Thu, 25 May 2017, Javier Degirolmo wrote:

> I was reading that if DOS is loaded high (which means into the 
> segment at FFFF) that then the CP/M entry point will still work with
> A20 enabled, especially since only programs that can only use 64KB
> would be using that kind of call (and wouldn't try to touch the HMA).

Yes.  DOS arranges to store a JMP instruction at the 1000C0 alias 
address.

> ...also apparently they messed up with the entry point from DOS 2 
> onwards (target address is two bytes too early it seems?), so I
> wonder how many programs even use it (since it'd mean they'd likely
> only work on DOS 1). Then again I guess directories probably messed
> them up already (with paths becoming longer than usual).

In my case, the program intercepts the CPM service point interrupt
vector and jumps there directly for its own purposes.  Therefore the
address stored at PSP:0005 doesn't matter (but I included it in the
patch for thoroughness).  The program is indeed very old, although not
quite as old as DOS 2.

It is true that almost no existing software uses this interface, and it 
has been broken for a long time.  But adding it (in fixed form!) to 
DOSBox adds so little overhead that there's no real reason not to put 
it in.  (If anyone thinks the off-by-2 value used by DOS >=2 should be 
emulated instead of fixed, it would be easy to adjust the patch.)


I hope this covers all your questions.

Alan Stern

sikthehedgehog:
> (If anyone thinks the off-by-2 value used by DOS >=2 should be emulated instead of fixed, it would be easy to adjust the patch.)

To be fair the ideal thing would be to just boot into a real DOS and for the emulator to only do the hardware (especially if there's anything that wants the real DOS code in memory). We're assuming that whoever is not doing that is probably just trying to get as much stuff running as possible, so I guess the non-broken (DOS 1) behavior should be the implemented one. I don't think there's anything out there that relies on this being broken, right?

AlanStern:
On Thu, 25 May 2017, Javier Degirolmo wrote:

> > (If anyone thinks the off-by-2 value used by DOS >=2 should be emulated instead of fixed, it would be easy to adjust the patch.)
> 
> To be fair the ideal thing would be to just boot into a real DOS and for the emulator to only do the hardware (especially if there's anything that wants the real DOS code in memory). We're assuming that whoever is not doing that is probably just trying to get as much stuff running as possible, so I guess the non-broken (DOS 1) behavior should be the implemented one. I don't think there's anything out there that relies on this being broken, right?

I am not aware of any.

Alan Stern

joncampbell123:
In typical DOSBox-X style, I'm thinking of making it a dosbox.conf option to select between DOS 1.x and DOS 2.x+ behavior. DOS 2.x+ behavior would be default, since that seems to be the standard for MS-DOS. You could select DOS 1.x behavior if you need old CPM code to run. Sound like a plan?

sikthehedgehog:
I thought about that but didn't suggest it since I didn't think it was worth the effort, but if you feel like adding a setting for it, sounds fine to me.

joncampbell123:
Settings are easy to add. The difference between them is whether or not the call address is off by 2. Not much additional effort at all.

AlanStern:
On Thu, 25 May 2017, Javier Degirolmo wrote:

> I thought about that but didn't suggest it since I didn't think it was worth the effort, but if you feel like adding a setting for it, sounds fine to me.

I have no objection.

Alan Stern

joncampbell123:
Hey, just noticed something relevent as well regarding MS-DOS 5.0 at least.

I produced the previous results by just running DEBUG.COM/DEBUG.EXE in each version of DOS without specifying a file to load.

However, if you ask DEBUG.COM to load a program (either COM or EXE), then the offset is correct.

MS-DOS 5.0, "DEBUG.COM"

    At PSP:0x05:      CALL F01D:FEEE

MS-DOS 5.0, "DEBUG.COM LABEL.EXE"

    At PSP:0x05:      CALL F01D:FEF0

MS-DOS 5.0, "DEBUG.COM COMMAND.COM"

    At PSP:0x05:      CALL F01D:FEF0

This suggests that the default setting should *not* be the off-by-2 setting.

I'll check the Windows 95, MS-DOS 6.22 and MS-DOS 2.1 bootdisks.

joncampbell123:
MS-DOS 2.1 is slightly different.

MS-DOS 2.1, "DEBUG.COM"

    At PSP:0x05:     CALL F01D:FEEE

MS-DOS 2.1, "DEBUG.COM LINK.EXE"

    At PSP:0x05:     CALL F00D:FFF0

joncampbell123:
Just added code to show this from hw/dos/test.c in DOSLIB. It also shows a valid pointer there.

guest os_000

joncampbell123:
So what I'll do is apply this patch, then modify the code further to put this CP/M call at F01D:FEF0 and a JMP from F01D:FEF0 to the INT 21h entry point. If more than 1MB is there, I'll have the DOS kernel block off the first 256 bytes of HMA to make the "mirror" work if A20 is enabled. The patch is correct in using the space for INT 30h and INT 31h for the JMP instruction that the CP/M far call goes to (F01D:FEF0 is a 1MB wraparound alias of 0000:00C0).

AlanStern:
On Thu, 25 May 2017, Jonathan Campbell wrote:

> Hey, just noticed something relevent as well regarding MS-DOS 5.0 at least.
> 
> I produced the previous results by just running DEBUG.COM/DEBUG.EXE in each version of DOS without specifying a file to load.
> 
> However, if you ask DEBUG.COM to load a program (either COM or EXE), then the offset is correct.
> 
> MS-DOS 5.0, "DEBUG.COM"
> 
>     At PSP:0x05:      CALL F01D:FEEE
> 
> MS-DOS 5.0, "DEBUG.COM LABEL.EXE"
> 
>     At PSP:0x05:      CALL F01D:FEF0
> 
> MS-DOS 5.0, "DEBUG.COM COMMAND.COM"
> 
>     At PSP:0x05:      CALL F01D:FEF0
> 
> This suggests that the default setting should *not* be the off-by-2 setting.
> 
> I'll check the Windows 95, MS-DOS 6.22 and MS-DOS 2.1 bootdisks.

That's nice to know.  Evidently the "Create PSP" function in DOS works 
slightly differently from the "Load Program" function.


On Thu, 25 May 2017, Jonathan Campbell wrote:

> MS-DOS 2.1 is slightly different.
> 
> MS-DOS 2.1, "DEBUG.COM"
> 
>     At PSP:0x05:     CALL F01D:FEEE
> 
> MS-DOS 2.1, "DEBUG.COM LINK.EXE"
> 
>     At PSP:0x05:     CALL F00D:FFF0

As opposed to F01D:FEF0 for command.com.

This is because DOS adjusts the offset part of the far pointer to be
equal to the maximum space available for the program's data segment,
rounded down to a multiple of 0x10 and capped at 0xFFF0 (since no 
larger multiple of 0x10 can be represented in 16 bits).

A COM program can't exceed 64 KB, and the first 0x100 bytes of the data
segment are used by the PSP, so the available space is 0xFFF0 - 0x0100
= 0xFEF0 unless the program was loaded close to the top of memory.  
But an EXE program's data segment does not include the PSP, so it has
the full 0xFFF0 bytes available.

If you can manage to load a program near the top of memory, you'll see 
the offset value adjusted to reflect that the data segment has less 
than 64 KB available.

Alan

joncampbell123:
No, actually MS-DOS 5.0 seems to set the field to one of two values depending on whether (apparently?) 64KB or more memory is available when the program loads.

If more than 64KB is present:

    CALL F01D:FEF0

If more than 64KB is present without loading a program:

    CALL F01D:FEEE

If 64KB or less:

    CALL 0000:00C0

If 64KB or less and not loading a program:

    CALL 0000:00BE

joncampbell123:
At a quick glance however your description seems to be correct regarding MS-DOS 2.1.

If I configure DOSBox-X to emulate a PC with 64KB of RAM and load the debugger, the address seems to change such that the offset field reflects the remainder of memory available.

AlanStern:
On Thu, 25 May 2017, Jonathan Campbell wrote:

> So what I'll do is apply this patch, then modify the code further to
> put this CP/M call at F01D:FEF0 and a JMP from F01D:FEF0 to the INT
> 21h entry point.

Actually it should jump to 0000:00C0, not to the INT 21h entry point.  
They are not the same: The CP/M entry point rearranges the stack and 
moves CL into AH before falling into the INT 21h code.

Alan

> If more than 1MB is there, I'll have the DOS kernel
> block off the first 256 bytes of HMA to make the "mirror" work if A20
> is enabled. The patch is correct in using the space for INT 30h and
> INT 31h for the JMP instruction that the CP/M far call goes to
> (F01D:FEF0 is a 1MB wraparound alias of 0000:00C0).

joncampbell123:
Thought so:

MS-DOS 5.0, when HIMEM.SYS is loaded and DOS=HIGH,UMB, the CP/M JMP is written to both 0:C0 and the F01D:FEF0 address in HMA. Even though the two memory contents are clearly different around the JMP instruction, the same JMP instruction is written to both locations.

joncampbell123:
Where is DEMO.COM attached by the way? Is there a download link?

sikthehedgehog:
...it was meant to be attached but apparently wasn't. Seems like it was a reply over e-mail and those attachments don't end up in the comment.

AlanStern:
On Fri, 26 May 2017, Jonathan Campbell wrote:

> Where is DEMO.COM attached by the way? Is there a download link?


On Sat, 27 May 2017, Javier Degirolmo wrote:

> ...it was meant to be attached but apparently wasn't. Seems like it was a reply over e-mail and those attachments don't end up in the comment.

That's right; it was an email attachment.  You can download the demo 
program from:

	http://www.rowland.harvard.edu/~stern/DEMO.COM

Alan

joncampbell123:
Ah, I see!

So using the CP/M compatible entry point means doing a near call to PSP:5, where DOS places a FAR CALL.

debug_000

It works, but only if you ask DOSBox-X to turn off the A20 gate.

dosbox_000

AlanStern:
On Sat, 27 May 2017, Jonathan Campbell wrote:

> Ah, I see!
> 
> So using the CP/M compatible entry point means doing a near call to PSP:5, where DOS places a FAR CALL.

Exactly.

> debug_000
> 
> It works, but only if you ask DOSBox-X to turn off the A20 gate.
> 
> dosbox_000

Your idea of reserving the first 256 bytes above 1 MB and storing a 
copy of the INT 30/31 far jump there sounds like a good approach.  It 
should be added on to my patch.

Alan

joncampbell123:
Done. Also some sanity checks for EMS/XMS emulation that affect A20 gate and CP/M entry point.

Any more .COM files that use the CP/M entry point?

AlanStern:
On Sun, 28 May 2017, Jonathan Campbell wrote:

> Done. Also some sanity checks for EMS/XMS emulation that affect A20 gate and CP/M entry point.
> 
> Any more .COM files that use the CP/M entry point?

None that I can distribute.  Of course, it's trivial to create your 
own.

Some of the very earliest DOS programs used the CP/M emulation
mechanism.  For instance, I have a vague memory that the original
version of msbasic.com was ported to DOS using automatic translation
from the earlier CP/M source code.  But it's quite possible that it
was some other program; I can't be sure.  This was over 30 years ago.

Also, I can't remember where I read the original documentation for the 
CP/M entry point.  It was probably the DOS-1 technical reference 
manual.

Alan

PR #290, opened 2017-09-08
MMX instruction set recompiler for core=dynamic

### LINKED ISSUE #288: Adding MMX instructions to dynamic recompilator
Hi all! :)

As far as I know, DOSBox-X does falling back to interpretator while executing MMX instructions when core=dynamic is used. Obviously, this approach leads to extremely slow performance while running MMX applications (such as [some late 90s DOS demos](https://demozoo.org/productions/tagged/needs-mmx/), Impulse Tracker, etc etc) comparing with regular integer/floating point instructions.

As dynrec is targeted to x86, it is possible (AFAIK again :) to dynamically recompile MMX code for native execution without performace (and probably compatibility) issues. So, there can be at least two approaches for recompiling MMX code:

- Drop-in recompilation
In this case, MMX instructions are placed to code cache blocks as-is (with address recalculations if needed), so they behaves just as if they are running under native x86 environment. This approach should not suffer from memory alighnment issues in most cases, but probably will lead to FPU save/restore issues (since an EMMS instruction will be required to return back to FPU mode)

- MMX to SSE2 recompilation
SSE2 supports integer operations hat are equivalent to MMX instructions but operating with wider registers (128 bit vs 64). Most SSE2 integer instructions shares their opcodes with their MMX equivalents, adding operand size prefix (0x66) before actual opcode (more info [here](http://ref.x86asm.net/coder32.html)), so conversions for most MMX instructions is pretty straightforward. As SSE2 regs won't collide with FPU stuff, there is no need to worry about FPU save/restore (yet again AFAIK =). However this approach will most likely lead to memory misalignment issues and therefore to performance drops/compatibility problems.

So, I tend to the first approach, since dynrec already has FPU save/restore logic and adding MMX instructions should be pretty safe.

Thanks in advance ;)

wbcbz7:
Hello,

here it is - MMX instruction set recompiler for core=dynamic, using host MMX instructions. Since it implements whole instruction set, it does not need to interact with MMX interpreter, allowing saving MMX state between code cache blocks by saving/restoring FPU state.
Although I don't have MMX benchmarks, perfomance increase is very noticeable, as it improves speed at least by 20-40 times
Tested with protected mode stuff (demos and intros like Discloned/Fusion) and with some real mode applications (Impulse Tracker 2.14)

p.s. hope it will close #288 nicely :)

### DIFF
--- a/src/cpu/core_dyn_x86/decoder.h
+++ b/src/cpu/core_dyn_x86/decoder.h
@@ -20,6 +20,11 @@
 #define X86_DYNFPU_DH_ENABLED
 #define X86_INLINED_MEMACCESS
 
+#define X86_DYNREC_MMX_ENABLED
+
+#ifdef X86_DYNREC_MMX_ENABLED
+void dyn_mmx_restore();
+#endif
 
 enum REP_Type {
 	REP_NONE=0,REP_NZ,REP_Z
@@ -367,6 +372,7 @@ static void dyn_check_bool_exception_ne(void) {
 }
 
 static void dyn_fill_blocks(void) {
+	GenReg *gr;
 	for (Bitu sct=0; sct<used_save_info; sct++) {
 		gen_fill_branch_long(save_info[sct].branch_pos);
 		switch (save_info[sct].type) {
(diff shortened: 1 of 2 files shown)


joncampbell123:
Looks good to me. As I develop on x86_64 I have no way to test this right now. Anyone else have objections?

kekko-dos:
@wbcbz7 your MMX instruction set for the recompiler is very nice. Good job!
I updated MMX instructions for the interpreter, fixing some long-standing bugs and making it use dosbox emulated fpu registers. Dosbox-x already adopted the changes. 
It would be nice if your recompiler also save/restore host registers to/from dosbox emulated fpu registers on dynamic block enter/exit, this way you can switch back and forth between recompiler and interpreter, primarily for diagnostic purposes, by just commenting some opcodes in the recompiler in order to make it fallback to the interpreter.
How could we proceed?
Thanks in advance!

PR #324, opened 2017-11-12
Nuked OPL support for OPL emulation

alexat:
https://sourceforge.net/p/dosbox/patches/270/

From here. According to forums, it's quite more accurate.


alexat:
For the time I run it, it definitely sounds better, but it may be personal preference :)

joncampbell123:
At first glance, I thought you meant you had removed OPL emulation entirely :)

PR #325, opened 2017-11-12
Direct3D and shaders support from forums

alexat:
Direct3D/shaders patch used in gulikoza/daum builds. Good for i.e. using nice xBR scaler shaders.


alexat:
Please take note, Direct3D SDK paths are hardcoded.

joncampbell123:
Before I accept this, would it be possible to define the Direct3D SDK paths in makefile variables so that users who put the SDK elsewhere can just edit the variables to point at it?

alexat:
Added path to Include/Library paths via proper %DXSDK_DIR% envvar, MS sets it.
Please also note this makes DirectX SDK a requirement. If undesired, it may be made conditional via #define / #ifdef

alexat:
Also added a shader set to use.

joncampbell123:
Ready? I'll merge if so.

Can you provide in your next commit instructions for users who want to #ifdef out the DirectX support if they really want to?

alexat:
Yes, basically ready. I'm now checking if existing config options are enough for disabling it and building without DX SDK.

PR #326, opened 2017-11-12
Make D3D support truly conditional (using HAVE_D3D9_H define)

alexat:
Just a small followup: with this change 
#define HAVE_D3D9_H 1
really regulates D3D support, if undef or set to 0, it compiles well without DX SDK


alexat:
Btw thanks! It's just great to have D3D in.

Issue #434, opened 2017-12-31
Suggestion: Key mapping to JIS 'ロ(Ro)' key


basic2004:
Dosbox-X with PC-98 or DOS/V when host keyboard is not JP layout, can't input 'Ro(0x73)' key.
I thought alternate mapping, this key to other key.
I suggest 'Application key', this is nearer with right shift, and this isn't used before Windows 9x.
Can we make this option in [dos] section of dosbox.conf?

joncampbell123:
I have a USB keyboard with the JP layout that I will plug into Windows 10 soon to re-work the scan code handling to better map the keys properly.

joncampbell123:
@basic2004 I have reworked the code to support a JP keyboard on the host and map input to PC-98 mode using the same layout and shift states as an actual PC-98 keyboard.

I still do not have a default mapping for Ro on US keyboards, but you can go into the mapper interface and map Ro to any key you want now.

joncampbell123:
I will close this issue unless @basic2004 still considers this an unsolved problem.

basic2004:
Thanks, I confirmed but too late response... sorry.

I tell more suggestion, Application key should'nt remap to Ro when **pc-98 force ibm keyboard layout** is **true**.

And I pressed Grave(Tilde) key but this pressed like Escape maybe this mapped STOP key in PC98 layout.

PR #512, opened 2018-02-17
Windows SDL 1.x: Fix keyboard mapping on non-US keyboard layouts

fincs:
This patch removes the need to load the US English keyboard layout on Windows SDL 1.x by replacing the usage of Virtual Key codes with raw keyboard scan codes.

This change avoids corrupting system-wide keyboard layout configuration on systems which do not have the US English keyboard layout installed (for instance, my system running Windows 10 FCU only has the Spanish keyboard layout installed). When *any* SDL 1.x based application is launched, this corruption happens and the keyboard layout list is filled with garbage (duplicate layouts, etc) - a restart is required to fix this mess.

This issue seems exclusive to SDL 1.x since 2.x doesn't appear to try to load the US keyboard layout.

This change may also help with fixing #434

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -1713,9 +1713,8 @@ static LRESULT CALLBACK WinExtHookKeyboardHookProc(int nCode,WPARAM wParam,LPARA
 								wParam = WM_USER + 0x101;
 						}
 
-						// catch the keystroke, post it to ourself, do not pass it on
-						PostMessage(myHwnd, wParam, st_hook->vkCode,
-							(st_hook->flags & 0x80/*transition state*/) ? 0x0000 : 0xA000/*bits 13&15 are set*/);
+						// catch the keystroke, post it to ourselves, do not pass it on
+						PostMessage(myHwnd, wParam, st_hook->vkCode, MAKELONG(0, MAKEWORD(st_hook->scanCode,st_hook->flags)));
 						return TRUE;
 					}
 				}
--- a/vs2015/sdl/src/video/windib/SDL_dibevents.c
+++ b/vs2015/sdl/src/video/windib/SDL_dibevents.c
@@ -41,6 +41,8 @@
 #include "SDL_dibvideo.h"
 #endif
 
+#include <dinput.h> // for the DIK_ defines
+
 #ifndef WM_APP
 #define WM_APP	0x8000
 #endif
@@ -49,7 +51,7 @@
 #define NO_GETKEYBOARDSTATE
 #endif
 
-/* The translation table from a Microsoft VK keysym to a SDL keysym */
+/* The translation table from Windows scan codes to a SDL keysym */
 static SDLKey VK_keymap[SDLK_LAST];
 static SDL_keysym *TranslateKey(WPARAM vkey, UINT scancode, SDL_keysym *keysym, int pressed);
 static SDLKey Arrows_keymap[4];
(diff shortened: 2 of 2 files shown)


joncampbell123:
How well have you tested this?

Does it break US keyboard layouts?

I have future plans for the mapper to represent non-US keys as well and to display non-US keyboard layouts.

joncampbell123:
I'm concerned that it completely replaces standard VK_* keyboard handling with DirectInput keyboard handling. I would accept this patch if it were conditional (such as adding to dosbox.conf something like "use scan codes=true" however.

fincs:
To be fair, my PR does not use any DirectInput API. It just uses the Windows scancode information that was already present in the standard Win32 messages (alongside VK). It just so happens that a good place to find scan code definitions was in the DirectInput headers.

joncampbell123:
I'm happy that this patch happens to work for you but it relies on scan codes in WM_KEYDOWN that happen to match DirectInput scan codes. There's no guarantee from Microsoft that the scan codes have any fixed meaning. So I'm going to have to reject this pull request.

https://msdn.microsoft.com/en-us/library/ms646267(v=vs.85).aspx

However this discussion does bring up an important point about the keyboard layout. The first obvious fix is to remove the code that loads the US keyboard layout so that the VK_ constant handling can deal with non-US keyboards, then to look at what these keys generate and add in SDLK constants for them if they're not there already.

I also see that your patch corrects my keyboard hook proc, which in retrospect really should be filling in the lParam field with the scan code provided instead of passing zero. That might have been the cause of your problems initially since SDL does use the scan code to call MapVirtualKeyEx for extended keys.

SDL 1.x already has a DirectInput handler which then properly uses the DIK_* scancodes. I may have accidentally disabled that when disabling directx video, but it might be a good fix to SDL 1.x to allow using both the windib video driver with the DirectInput keyboard input code somehow.

sikthehedgehog:
I believe the warning about scancodes was because Windows originally could
run on non-PC platforms and hence would be probably intended to return the
native scancodes for the platform. In practice it always returns PC set 1
scancodes, even when most keyboards use set 2 nowadays. I guess too many
programs relied on it :P

That said, SDL1 (and SDL2) can handle keys by physical location so there's
no real reason to not use that.

El 20/02/2018 14:15, "Jonathan Campbell" <notifications@github.com>
escribió:

Closed #512 <https://github.com/joncampbell123/dosbox-x/pull/512>.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/joncampbell123/dosbox-x/pull/512#event-1482918871>, or mute
the thread

fincs:
I have had a chance to try the latest commits in upstream DOSBox-X, and I can confirm that now that the US keyboard layout code is removed, I no longer have the corruption problem I noted in my OP. However there's a different problem now; keypresses inside DOSBox-X no longer match what the emulated system expects (so for example I press the `ñ` key on my keyboard and instead `º` pops up -
 both host and guest keyboard layouts are set to Spanish). This is the problem I anticipated happening here:

>  Actually, using VKs while a non-QWERTY layout is active would produce all sorts of weird behaviour when trying to type certain keys - this is why the original SDL 1.x code loaded the US keyboard layout

joncampbell123:
In order to handle non-US layouts the code will have to be updated add SDLK constants representing anything that cannot be properly represented on a US keyboard layout. By that I mean the non-shifted interpretation of the key. This is where it becomes important for DOSBox-X to know the keyboard layout and to keep track of notifications from Windows of layout changes.

joncampbell123:
I would like DirectInput keyboard input as well. Currently in the SDL codebase it's only enabled if you're willing to go all in and use DirectX video output as well, so the work to do is to get windib to cooperate with the DirectInput code when asked to by the user.

fincs:
Well actually, the `ñ` key is supposed to take the same place as the `;` key on US keyboards. I've just tested what happens if I press this key if the US layout is active within DOSBox-X (i.e. `keyb us`). Looks like I get a backtick (\`) instead. Confusingly enough, if I press the backtick key (actually the grave accent dead key) I get the semicolon...

joncampbell123:
It's also possible that you may have set your mapper up around the US layout. Try going into the DOSBox-X mapper and re-configuring your keymap. I added debug information at the bottom right corner to aid with this, by showing you the scan code and SDL information.

fincs:
Thank you for pointing me towards mappers. Turns out the default mapper configuration is only suitable for US keyboard layouts. I've went ahead and remapped all keys that were wrong on my Spanish layout. Now everything works fine :)

PR #540, opened 2018-02-26
Show menu bar toggle

Shane32:
Currently there does not seem to be a way to hide the menu from the gui.  This feature change makes it so that the 'show menu bar' selection in the system menu is a toggle, allowing the menu bar to be turned on or off at runtime.


joncampbell123:
Merged.

I think what you meant is that the system menu dropdown (not the main menu bar) should turn the "show menu bar" into a toggle, which is fine by me.

PR #541, opened 2018-02-26
Add feature: memory-backed disk images

Shane32:
This new feature lets a user use imgmount to mount a memory-backed virtual hard drive (aka a ramdrive).  Syntax is simple: imgmount d -t ram -size 200   (where the number is the amount of MB to allocate).  Actual memory allocaton is performed as needed.  Also included is a help feature for imgmount; just type imgmount for the syntax (also supports -? or -help).

Copying files to the ramdrive is quite slow (by today's standards), probably due to the many layers of emulation.  Most likely it would be faster if the ramdrive emulation was at the file-system level, whereas my code emulates a raw image and relies on the rest of dosbox to provide the file-system emulation layers.

The code consists of:
(a) new class based on imageDisk that uses ram behind the Read_AbsoluteSector and Write_AbsoluteSector functions
(b) code copied from imgmake to format the blank ramdisk upon initial load
(c) modification to the imgmount program to support "-t ram"

Notes:
(a) There seems to be an existing memory leak where images mounted with imgmount automatically get mounted in the virtual bios as well as in the virtual dos.  imgmount -u is specifically not supported for images mounted that way.  Because of this bug (or "missing feature"), ramdrives do not mount in the bios at all at this time.  Running imgmount -u on a ramdrive is verified to release all memory allocated.
(b) Since a ramdrive cannot be accessed through the bios, specifying c/h/s for a ramdisk seemed pointless, so the -size parameter has been changed (for "-t ram" only) to accept a number that equals a number of megabytes to allocate.  This number is rounded up based on the c/h/s chosen by the underlying code.
(c) For some reason, each file copied to the ramdrive consumes 100k+ on the ramdrive even if it's a very small file.  I'm guessing that there is some edit to the formatting code (adapted from imgmake) that would adjust the FAT16 cluster size to fix the problem.  Another possibility is changing the method of calculating c/h/s, as right now heads always equals 255 and sectors always equals 63.
(d) You may update the copyright notice to match the rest of the repository if you wish.

### DIFF
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -137,6 +137,30 @@ public:
     std::vector<vfdentry> dents;
 };
 
+class imageDiskMemory : public imageDisk {
+public:
+	virtual Bit8u Read_AbsoluteSector(Bit32u sectnum, void * data);
+	virtual Bit8u Write_AbsoluteSector(Bit32u sectnum, void * data);
+	virtual Bit8u GetBiosType(void);
+	virtual Bit8u Format();
+
+	imageDiskMemory(Bit32u imgSizeK);
+	imageDiskMemory(Bit32u cylinders, Bit32u heads, Bit32u sectors, Bit32u sectorSize);
+	imageDiskMemory(diskGeo floppyGeometry);
+	virtual ~imageDiskMemory();
+
+private:
+	void init(Bit32u cylinders, Bit32u heads, Bit32u sectors, Bit32u sectorSize);
+
+	Bit8u * * MemMap;
+	Bit32u sectors_per_ramcluster;
+	Bit32u ramcluster_size;
+	Bit32u total_ramclusters;
+	Bit32u total_sectors;
+
+	Bit8u bios_type;
+};
+
 void updateDPT(void);
 void incrementFDD(void);
 
--- a/src/dos/dos_programs.cpp
+++ b/src/dos/dos_programs.cpp
@@ -1659,7 +1659,7 @@ restart_int:
 		return;
 	}
 	void printHelp() { // maybe hint parameter?
-		WriteOut(MSG_Get("PROGRAM_IMGMOUNT_SYNTAX"));
+		WriteOut(MSG_Get("PROGRAM_IMGMAKE_SYNTAX"));
 	}
 };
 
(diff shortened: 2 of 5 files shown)


joncampbell123:
When you add new source files please add them under the src/ directory, not vs2015.

joncampbell123:
Found some issues with the new code.

As long as you don't hit them, it works fine.

https://github.com/joncampbell123/dosbox-x/issues/545

Shane32:
Will do; thanks! Glad you like the feature but I won’t feel bad if you reject anything I upload! I should be able to fix the bug tomorrow and maybe next week I’ll have time to go back and add detailed comments as well as track down the cluster size issue. 

Of course if you prefer kB that’s fine. Since the code now only works in 8mb increments I just did mb. Ideally the size parameter would accept c,h,s or ##kb or ##mb, etc but I’m not a C programmer really at all so I was struggling with the text parsing.

Shane32:
If you look at that parsing code, by the way, it did seem that a number longer than 20 digits given to the size parameter (for c,h,s parsing) would cause a buffer overrun, but then again I’m not up on C string parsing. The code I added shouldn’t be vulnerable to the buffer overrun, but I didn’t fix the old code.

joncampbell123:
For the time being, taking the size in KB and rounding up to the next 8MB is perfectly fine.

Just curious, what's the reason behind the 8MB increments? I don't remember the FAT filesystem being that picky.

Shane32:
I was lazy with the chs code and it just maxes out the h and s (63 and 255) and calculates c. I saw some better code elsewhere within dosbox that I could adapt that would work better. (I forgot where at the moment.) I think it increased h only as necessary to keep c < 1024. Since the ramdrive doesn’t use ram unless you copy files to it, it didn’t seem to be worth the effort to be able to make smaller ramdrives.

Shane32:
The underlying ramdrive code fully supports any chs parameters that won’t overflow internal variables, and should reject any invalid parameters nicely, and it supports floppy drives. But the formatting code requires drives with 512 byte sectors and so on. The formatting code should work with RAM disk floppies also but I didn’t test it, as I didn’t code imgmount to mount ram based floppies.

Shane32:
Once imgmount -u can unmount bios-mounted disks properly, I could enable advanced RAM disk mounting options within imgmount (otherwise, ramdisk memory can’t be released). But at the moment I’m not sure what needs to be done.

PR #565, opened 2018-02-28
Imgmount ramdrive updates

Shane32:
Three commits:
- Added lots of comments
- Renamed 'ramcluster' to 'chunk' to avoid confusion between FAT clusters
- Updated ramdisk formatting code to better match FAT12/FAT16 specifications and limits

Possible future features:
- consolidate ramdisk and imgmake formatting code
- implement GetLabel code for images (volume label)
- fix/write/verify boot sector code for formatting code, or disable boot sector signature being written


Shane32:
My code still doesn't meet the spec perfectly....I'm still working on it.  Of course, it always estimates in favor of FAT16, which means the worst-case scenario is that there is extra space allocated to the FAT than necessary

Shane32:
Specifications I'm looking at now:
http://download.microsoft.com/download/1/6/1/161ba512-40e2-4cc9-843a-923143f3456c/fatgen103.doc

joncampbell123:
Got a suggestion from the Vogons forums:

https://www.vogons.org/viewtopic.php?f=41&t=31881&p=654055#p654055

PR #595, opened 2018-03-10
Imgmount ramdrive progress

Shane32:
Work progress on ramdrive code.
- rewrote the formatting code nearly from scratch, to exactly compute the amount of data sectors that can fit on a given disk
- formatting code can format floppies to match original specifications more closely (2 sectors per cluster for certain disk types)
- ramdrives can mask existing drives, for copy-on-write disk images

Todo:
- fix the bug (memory leak) where imgmounted disks do not get released from bios or ram when unmounted.
- enable mounting ramdisks into the bios
- check/update boot command to allow booting into a mounted ramdisk
- verify formatted floppies works with early versions of PC-DOS and MS-DOS (assuming a compatible type of floppy)
- determine if boot signature should not be written to the disk (because we are not writing a boot sector)
- move Format() to imageDisk, and add format command to dos emulation, so any mounted image can be formatted
- add -cow (?) option to imgmount to mount a copy-on-write image

And maybe:
- add command to save a mouted image (particularly for ramdisks)
- replace imgmake implementation of formatting with new Format() implementation
- enhance imgmount help further (probably pause between pages)
- clean up imgmount / imgmake code
- rename imageDisk variables to match a design pattern, and scope them properly

### DIFF
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -41,6 +41,9 @@ struct diskGeo {
 	Bit16u cylcount;  /* Cylinders per side */
 	Bit16u biosval;   /* Type to return from BIOS */
     Bit16u bytespersect; /* Bytes per sector */
+	Bit16u rootentries;  /* Root directory entries */
+	Bit8u sectcluster;   /* Sectors per cluster */
+	Bit8u mediaid;       /* Media ID */
 };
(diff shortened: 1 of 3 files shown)


Shane32:
not sure why all my commits are in there twice, but i'm going to close the pull request and fix it

PR #597, opened 2018-03-10
XMS message fix

Shane32:
Fixed bug where a static error message was performing a lookup in a string table


Shane32:
Also added -? help for loadfix.

PR #607, opened 2018-03-11
Refactored imgmount (no functional changes)

Shane32:
Separated code for el-torito floppies, cdroms, and fat drives into subroutines.  Added a few comments.  Didn't change any logic, so it should perform identical to before.


Shane32:
On second thought, maybe I should have moved each subroutine in a separate commit.... a little late now....

Shane32:
You're welcome!  There's bugs in here I'd like to fix but it was too messy to understand.

PR #608, opened 2018-03-12
Misc code cleanup & refactoring

Shane32:


joncampbell123:
Before I accept this, I have a few questions:

- Does this patch preserve the HDI/FDI disk image support?
- StartSector was the means by which hard disk image mounting could pick the partition, and work from that partition, in the FAT driver. Does the new code preserve the ability to mount a hard disk image and the DOS partition within?

joncampbell123:
The only other problem I see in the patch is the assumption that all FAT filesystems have the boot signature 0x55AA at offset 510 of the sector. That is only true if the boot sector size is 512 bytes per sector. PC-98 disk images do NOT have that signature, especially in the 1024 bytes/sector disk format.

joncampbell123:
I can offer some FDD, HDI and FDI disk images for you to test your code against if you like.

joncampbell123:
I'm testing the code so far.

- FDI works.
- FDD works.
- HDI works.

Booting FDI and FDD works.

Booting Windows 95 from a disk image still works.
Booting Windows 98 SE from a disk image still works.

Okay, I'll merge the changes.

Shane32:
startSector was overwritten on every code path.  So that's why I removed it from the constructor.  I didn't change the magic byte detection algorithms either.

I'll try to inform you if I think there's any potential problems I foresee.  I'd appreciate the test images - thanks!

Where is the best spot to chat about a change before I post a pull request?  Create an 'issue'?  vogons?

joncampbell123:
A discussion on vogons sounds good to me. Let me know where you open the conversation. I think the rule is that once I respond to a thread, the site will notify me of other postings on that thread, correct?

Issue #593, opened 2018-03-09
Segfault when mounting readonly floppy image files under Linux OR booting from a disk image with odd file size


frank-deng:
When mounting floppy image files with only reading permission using `IMGMOUNT` under Linux, DOSBox-X will crash with Segmentation fault.

joncampbell123:
Do you mean mounting with the -ro option, or mounting disk images marked read only by your host filesystem?

frank-deng:
I tried to mount disk images marked read only on my Linux host filesystem.

joncampbell123:
Ah, classic. Code that uses fopen() and doesn't check the return value against NULL, probably inherited from the original DOSBox project.

I'll check to see if this is the case.

frank-deng:
Thanks!

joncampbell123:
I'm not able to reproduce the issue, even if I make the floppy image read-only and immutable. It correctly errors out if the file is immutable and does not crash. Can you try the latest commit?

frank-deng:
I tried the latest commit of DOSBox-X, the same problem occurs with the following dosbox command:
`imgmount 0 /home/frank/.dosbox/boot.img -t floppy -fs none`

joncampbell123:
I still can't reproduce the issue, but I have come across this bizarre bug where mounting a disk image who's size doesn't match any geometry causes DOSBox-X to segfault.

Shane32:
@frank-deng has the problem been solved with the latest commits?

frank-deng:
Thanks. I'll try the latest commit some days later.

frank-deng:
Thanks! It no longer segfaults when loading floppy disks with readonly permission under Linux.

PR #613, opened 2018-03-13
Misc bug fixes

Shane32:
- validation check in el-torito mounting bug fix
- assignment in IDE_Auto for 3rd and 4th drive (which was never used since BIOS currently only supports 2 HDDs)
- ensure checks against imageDiskList to use MAX_DISK_IMAGES
- ramdrive formatting minimum size check validation update

### DIFF
--- a/src/dos/dos_programs.cpp
+++ b/src/dos/dos_programs.cpp
@@ -2425,7 +2425,7 @@ public:
 				return;
 			}
 			drive=temp_line[0];
-			if ((drive<'0') || (drive>3+'0')) {
+			if ((drive<'0') || (drive>=MAX_DISK_IMAGES+'0')) {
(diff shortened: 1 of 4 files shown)


Shane32:
- log message cleanup in the ramdrive formatting code

Shane32:
- Print success message when mounting el torito floppy image from cdrom

Shane32:
the imageDisk name is only used in swapInDisks, but it IS a public member....

Shane32:
Pushed potential fix for #593 ---- added checking for failure to open files when -fs none

joncampbell123:
Nice!

Perhaps the message name could be PROGRAM_MOUNT_STATUS_ELTORITO instead of PROGRAM_MOUNT_STATUS_2 in the near future?

PR #617, opened 2018-03-15
Mt32 update to libmt32emu_2_3_0

heydojo:
I couldn't get mt32 in dosbox-x to work so I created a branch and rebased it using the code at https://github.com/munt/munt and this is the result.
I don't have a build environment set up so I don't know if this code even builds.
I will commit a patch file into the mt32 branch on my fork for anyone who wants to play with this.


joncampbell123:
Wait... do you want me to apply this or did you change your mind?

heydojo:
It won't merge cleanly for you, I don't think. Some maintainers get really funny about how they take their merge requests and I didn't want to take the risk. That is to say that there will be a patch folder with the patch inside it in the main directory. I forgot also that git hub will track the branch and any future changes in a merge request and not HEAD at the time of proposing the merge.
It's been a while.

Plus I don't think merging this into master until it has been tested is a good idea either.
What do you think about this branch?

PR #623, opened 2018-03-18
PC-98: Add switching EGC/GRCG/16-colors video options to menu.

yksoft1:
I don't know if those features are needed but I sometimes not loving to open the .conf just for those options..


joncampbell123:
I noticed while testing recent code that early 1990s PC-98 games like to read the BIOS data area to determine if EGC/GRCG hardware is available. If EGC is available, they change their graphics drawing code to use the EGC ROP functions.

Later PC-98 games (including Touhou Project) assume EGC is available, so turning it off in dosbox.conf shows you how the game would mis-draw the screen without EGC functions.

Without these options, you would never know.

PR #625, opened 2018-03-19
Updates to ramdrive and imgmount

Shane32:
New:
- More code cleanup in imgmount
- Fixed unchecked bounds while parsing size parameter. and makes sure size parameter values are >0.
- Added -chs parameter, which does the same as -size, but matches imgmake and seems to make more sense than inputting the values in reverse order.  Also, when using -chs, sector size is optional (defaults to 512).
- Added capability to imgmount ramdrives to A: or B:, and fixed some related bugs.

Problems still existing:
- Cannot mount ramdrive to B: and use BOOT to boot to A:, as BIOS does not simulate the fact that two floppy drives are connected.  Tried calling incrementFDD() from BOOT::Run but that made it worse.  Using imgmount to mount the floppy to A: and then boot -l A works as expected, however.

Also:
- Cannot boot to hard drive image (problem existed before this set of commits)

### DIFF
--- a/src/dos/dos_programs.cpp
+++ b/src/dos/dos_programs.cpp
@@ -2280,7 +2280,6 @@ public:
 		//initialize variables
 		DOS_Drive * newdrive = NULL;
 		imageDisk * newImage = NULL;
-		Bit32u imagesize;
 		char drive;
 		std::vector<std::string> paths;
 		//show help if no arguments or -?
@@ -2327,7 +2326,7 @@ public:
 		cmd->FindString("-el-torito",el_torito,true);
 		if (el_torito != "") {
 			//get el-torito floppy from cdrom mounted at drive letter el_torito_cd_drive
-			el_torito_cd_drive = el_torito[0];
+			el_torito_cd_drive = toupper(el_torito[0]);
 			//validate the el_torito loading (look for boot image on the cdrom, etc), and
 			//  find the el_torito_floppy_base and el_torito_floppy_type values
 			if (!PrepElTorito(type, el_torito_cd_drive, el_torito_floppy_base, el_torito_floppy_type)) return;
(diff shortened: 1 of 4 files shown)


yksoft1:
PC-98 could use A: or B: as hard drive, and C: or D: could be used as floppy drives.

Shane32:
Is PC-98 still based on floppy and IDE controllers with a PC type BIOS?  If that's the case, then I would expect that floppies would still be connected to the FDC (positions 0 and 1 with imgmount), and hard drives would be connected to the HDD (positions 2, 3, 4, 5 with imgmount).  Would this be correct?

Assuming that's the case, imgmount with a drive letter will always fail because they are not a 512-bytes/sector drive, and imgmount with a number (and -fs none) will work correctly: only floppies at 0 or 1, and only hard drives at 2/3/4/5.  Once booted into the PC-98 OS, the OS can mount the drives to any drive letter it wants.

Sorry, but I've never used PC-98 so I am unfamiliar with the hardware or software of the system, and just guessing on how it works.

joncampbell123:
I also want to add that the interrupt layout and BIOS are *completely different* from the IBM PC platform.

joncampbell123:
@Shane32 The simplest change I can recommend is to have it check floppy vs hard drive and drive A vs B only if not running in PC-98 mode.

On PC-98 MS-DOS, drive A is the boot drive (even if the hard drive), and B is the next device. For example on one system I have, drive A is the hard disk and B is the floppy drive. Except when I use a bootdisk, then A is the floppy drive and B is the hard drive.

Shane32:
Updated in #631 per your suggestion...although I don't have PC-98 images to test with

joncampbell123:
Here's a whole bunch of test material for you.

https://www.myabandonware.com/browse/platform/pc98/page/3/

Some are FDD, FDI, and some are HDI.

Issue #658, opened 2018-04-09
Cannot run .bat files correctly from console command if they are not in current directory


yksoft1:
In Dosbox SVN, it would be this situation running:
`dosbox e:\demos\autoexec.bat
`
１. Dosbox mounts e:\demos as drive C and change (emulated) current directory to C:\\.
２. Dosbox runs autoexec.bat .
However in current Dosbox-X (In Windows, no matter built by what),
１. Dosbox-X mounts the (physical) current directory as drive C and change (emulated) current directory to C:\\.
２. Dosbox-X tries to run e:\demos\autoexec.bat, but fails as drive C is not mounted as e:\demos.
３. Dosbox-X unmounts drive C and change (emulated) current directory to Z:\.

I think @joncampbell123 had a mistake in #369 by mounting current directory instead of where BAT files are in. However it's more complex to fix than I thought as there is some problems:

- joncampbell123's code seemed to support multiple .BATs in Dosbox-X command line and we should remount C: for each bat file.
- If we met a directory with non-ASCII double-byte filename, there might be problems finding the path separator.

yksoft1:
Seemed to be fixed after #659 merged. Closed.

joncampbell123:
Will do

PR #659, opened 2018-04-09
Fix parsing .BAT filenames, which not in current directory, in command line.

yksoft1:
This is for fixing bug #658.


joncampbell123:
Appreciate it, thanks!

PR #671, opened 2018-04-21
Write 'Dynamic VHD unsupported' error to stdout

Shane32:


yksoft1:
Would you port dynamic VHD support from QEMU or other emulators in the future?

Shane32:
I'm sure it would be very easy to add from scratch, which I'd be happy to do, when I have time.  But I've been pretty busy recently.

joncampbell123:
It depends whether code can be imported from QEMU without too much trouble.

PR #672, opened 2018-04-22
Dynamic VHD support

Shane32:
First beta of dynamic VHD drive support.
- It works on Windows x86 and x64, but I don't know about Linux.
- Does not look for 511-byte footers (pre-Microsoft Virtual PC 2004 file format)
- Although not necessary, it utilizes old code to mount fixed size VHD drives, so any patches that had been implemented will still be effective.
- Does not prevent abnormally large VHD drives from loading (which would be incompatible with int 13h code).  (That probably should be added.)
- It should support differential VHD drives; however, it requires a single function to be implemented so that it can be tested and used.  The function needs to converts filenames from UTF-16 to ASCII, and convert relative pathnames based on the location of the parent VHD filename.  I'm not sure exactly how to do that.
- An additional file containing the imageDiskVHD class was added and included in the VS2015 project.  It will need to be added to the Linux build file.
- I have only tested it with a single dynamic VHD file.  DosBox-X successfully mounts, makes changes to, expands, and boots from the file I was testing.  It mounts by number with '-fs none' also.  More testing would be useful.
- The code is new (not copied from any other project), and should be decently commented for the most part

### DIFF
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -175,6 +175,112 @@ private:
 	diskGeo floppyInfo;
 };
 
+class imageDiskVHD : public imageDisk {
+public:
+	enum ErrorCodes : int
+	{
+		OPEN_SUCCESS = 0,
+		ERROR_OPENING = 1,
+		INVALID_DATA = 2,
+		UNSUPPORTED_TYPE = 3,
+		INVALID_MATCH = 4,
+		INVALID_DATE = 5,
+		PARENT_ERROR = 0x10,
+		ERROR_OPENING_PARENT = 0x11,
+		PARENT_INVALID_DATA = 0x12,
+		PARENT_UNSUPPORTED_TYPE = 0x13,
+		PARENT_INVALID_MATCH = 0x14,
+		PARENT_INVALID_DATE = 0x15
+	};
+	enum VHDTypes : Bit32u
+	{
+		VHD_TYPE_NONE = 0,
+		VHD_TYPE_FIXED = 2,
+		VHD_TYPE_DYNAMIC = 3,
+		VHD_TYPE_DIFFERENCING = 4
+	};
+	VHDTypes vhdType;
+	virtual Bit8u Read_AbsoluteSector(Bit32u sectnum, void * data);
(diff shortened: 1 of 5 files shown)


joncampbell123:
Will you be doing further development with C++11 type code?

I've been planning on eventually making C++11 required in this codebase instead of supporting both GCC 4.6 C++0x and GCC 4.8 and C++11.

Const initialization in the C++ class on the same line as the variable is not supported in C++0x

joncampbell123:
If you're OK with C++0x support still, I will need to modify the VHD class to initialize those variables in the constructor instead.

Shane32:
Go ahead and make any other changes you want also. I don’t generally have much free time to spend on fun stuff like this. Sometime I was hoping to unify all the shared code between different areas of imgmount and boot — but that will have to wait. For example, the code could probably use a more extendable method of supporting file types. And with proper stream support, it could possibly support mounting a image file from within a mounted image file. But eventually it gets to be a bit pointless for the intended purpose

joncampbell123:
While we're on the topic of IMGMOUNT... I noticed that IMGMOUNT will accept multiple disk images and enable swapping between them IF you mount them to a drive letter.

If you mount to 0 or 1 (with -fs none) then swapping does not work and only the last disk is attached.

Do you want to work on that or should I take care of it?

joncampbell123:
I'll make changes on my end, but to continue merging without issues you'll need to git pull before working on the code again.

Shane32:
Either way. I will make sure there are no conflicts when / if I push a change request. 

Let me take a look at that stuff real quick though.

Shane32:
Ok I'm a little bit confused -- are you saying that initializing class variables needs to be done in the class constructor versus with the definition?  (I thought you meant that the const keyword wasn't supported)

Shane32:
i'll work on this some more tonight after dinner.  i found a reference count problem that probably relates to the new code also.

Shane32:
if you push any changes in the meantime, i'll pull them

joncampbell123:
Or of course you can also:

someconstructor() {
something = 12345;
}

int something;

PR #674, opened 2018-04-23
imgmount multiple file loading update

Shane32:
I cleaned up imgmount a bit more, and fixed the imgmount <letter> code so that it properly can swap between auto-geometry-detected or VHD hard drives (using the swap-cdrom hotkey).  (Before it would only work with -size specified for hard drives.)

However, mounting by number (with '-fs none') doesn't use DriveManager by design (as it doesn't mount the image's file system), so there's no simple way that I see to allow for it to allow for swapping between multiple files.  I did quickly add a simple error message if you were to try that.  (The error message could be rewritten, probably, as could a lot of these error messages.)

### DIFF
--- a/src/dos/dos_programs.cpp
+++ b/src/dos/dos_programs.cpp
@@ -2434,8 +2434,9 @@ public:
 		}
 			
 		// find all file parameters, assuming that all option parameters have been removed
-		ParseFiles(paths);
+		ParseFiles(temp_line, paths);
 
+		// some generic checks
 		if (el_torito != "") {
 			if (paths.size() != 0) {
 				WriteOut("Do not specify files when mounting virtual floppy disk images from El Torito bootable CDs\n");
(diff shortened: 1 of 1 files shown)


joncampbell123:
If you can't get IMGMOUNT with -fs none to disk swap that's OK, I can probably figure it out when I have time. Thanks!

Shane32:
If you wanted to add this feature, imgmount should be cleaned up quite a bit now to the point where it should be easy to add.

I really dislike how the patches for mounting different file types are scattered around various classes.  For instance, the .hdi support code is split between a patch in the imageDisk constructor and the imgmount code.  But it would take a bunch of careful thought and redesign to fix without breaking anything.

I'd also suggest breaking imgmount, boot, and mount into separate files from the rest in dos_programs.cpp.  The file is over 4,000 lines of code now.  I also don't know why imgmount and the other programs are under the MOUNT class.  It would seem that all the programs could be under a PROGRAMS class (perhaps), but not under the MOUNT class.

Shane32:
Ideally, perhaps, each file type would have it's own code, with a function to test for compatibility.  The classes could be based on a more generic stream class, and could be chainable, so that code that tests for misaligned MBR structures or similar would work no matter what underlying file type was used.

joncampbell123:
Agreed! Also, the FAT driver should be separated from partition table handling.

Shane32:
Exactly

joncampbell123:
BOOT seems to be able to handle multiple disk images and the swaplist.

I'll see if I can fix IMGMOUNT to work the same with -fs none.

joncampbell123:
Done.

PR #680, opened 2018-04-25
Add an experimental implementation for PC-9821 31KHz video mode.

yksoft1:
I used some display parameters based on http://island.geocities.jp/cklouch/column/pc98bas/pc98dispout2_en.htm, http://tinyvga.com/vga-timing/640x400@70Hz and speculation. 
Experimental and might not be correct at all.

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -1185,6 +1185,10 @@ void DOSBOX_SetupConfigSections(void) {
                     "in 200-line graphics modes upconverted to 400-line raster display. When enabled, odd\n"
                     "numbered scanlines are blanked instead of doubled");
 
+	Pbool = secprop->Add_bool("pc-98 31-khz video mode",Property::Changeable::WhenIdle,false);
+	Pbool->Set_help("Emulate PC-9821 31-KHz VGA compatible video mode when enabled. \n"
+					"Experimental and may not be accurate at all.");
+
 	Pint = secprop->Add_int("pc-98 timer master frequency", Property::Changeable::WhenIdle,0);
 	Pint->SetMinMax(0,2457600);
 	Pint->Set_help("8254 timer clock frequency (NEC PC-98). Depending on the CPU frequency the clock frequency is one of two common values.\n"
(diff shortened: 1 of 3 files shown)


yksoft1:
Read it carefully. I did set the pixel clock to 25175000 in vga_draw.cpp if pc98_31khz_mode is not false.

joncampbell123:
Implementing the 31khz mode without yet setting the BIOS data area bytes to represent it is perfectly fine at this point.

joncampbell123:
I don't see it in the patch. I see additional if() statements in preparation for it, but I don't actually see any assignment of the pixel clock to 25175000.

yksoft1:
// TODO: Found a monitor document that lists two different scan rates for PC-98:
        //
        //       640x400  25.175MHz dot clock  70.15Hz refresh  31.5KHz horizontal refresh (basically, VGA)
        //       640x400  21.05MHz dot clock   56.42Hz refresh  24.83Hz horizontal refresh (original spec?)
		
        if (false/*future 15KHz hsync*/) {
            oscclock = 14318180;
        }
        else if (!pc98_31khz_mode/*24KHz hsync*/) {
            oscclock = 21052600;
        }
        else {/*31KHz VGA-like hsync*/
            oscclock = 25175000;
        }

joncampbell123:
That will do. I'll merge and add that change.

joncampbell123:
Ah! I see my mistake. Your patch didn't need to change anything!

yksoft1:
By the way, 15-khz mode is only there for PC-88 monitor compatibility. Even the first PC-9801 released in 1982 use a 24-khz digital monitor and defaults to 640x400 mode.

joncampbell123:
Do you mind if I change the dosbox.conf variable to a string that specifies 15khz / 24khz / 31khz instead of just a boolean to enable?

yksoft1:
Sure and I'm eager to see how DOS and apps/games was running in 640x200 mode (15-khz mode only support 200-lines if my memory is correct).

joncampbell123:
@yksoft1 That's what I'd like to know too! 15-khz mode if I understand it is probably similar to the IBM CGA where the video timings are similar or the same as NTSC 60Hz video and the center 200 lines are active display out of 262 total lines per field.

yksoft1:
I didn't find 8x8 fonts in anex86.bmp... then how does it show 80x25 characters in 640x200 mode?

joncampbell123:
At the moment, it doesn't. FONT.ROM has bitmaps for 8x8, though DOSBox-X ignores it at this time.

PR #687, opened 2018-04-29
Enable differencing VHD disks

Shane32:
Windows support: absolute and relative parent paths
Linux support: relative parent paths only

- Untested on Linux
- Assumes that C:\dir\dir\ ..\ ..\mydisk.vhd is accepted by fopen
- Fails if the parent disk filename contains unicode characters

### DIFF
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -265,6 +265,7 @@ private:
 	static ErrorCodes TryOpenParent(const char* childFileName, const ParentLocatorEntry &entry, Bit8u* data, const Bit32u dataLength, imageDisk** disk, const Bit8u* uniqueId);
 	static ErrorCodes Open(const char* fileName, const bool readOnly, imageDisk** imageDisk, const Bit8u* matchUniqueId);
 	virtual bool loadBlock(const Bit32u blockNumber);
+	static bool ConvertUTF16toASCII(std::string &string, const void* data, const Bit32u dataLength);
 
 	imageDisk* parentDisk;// = 0;
 	Bit64u footerPosition;
--- a/src/ints/bios_vhd.cpp
+++ b/src/ints/bios_vhd.cpp
@@ -42,6 +42,8 @@
 * - code does not prevent loading if the VHD is marked as being part of a saved state
 * - code prevents loading if parent does not match correct guid
 * - code does not prevent loading if parent does not match correct datestamp
+* - for differencing disks, parent paths are converted to ASCII; unicode characters will cancel loading
+* - differencing disks only support absolute paths on Windows platforms
 *
 */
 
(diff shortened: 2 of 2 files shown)


Shane32:
The MSDN documentation says that fopen uses the "ANSI codepage (CP_ACP)", which means the current default code page.  Typically in the US this is Windows-1252 encoding, which although supporting a slightly extended range of characters, is not UTF-8, and does not support the majority of unicode characters.  If fopen on Linux supports UTF-8, then yes, that would be an easy fix.  On Windows, ideally, one would call _wfopen, which is fine for absolute path names, but for relative path names, one would need to convert the child pathname from Windows-1252 into Unicode before appending the unicode relative parent path name on the end.

So in the end, I just figured that if it only supported ASCII, it would work no matter which compiler or operating system was in use (providing that the string was ASCII, of course).

- What code page does Linux use for fopen?
- Can we be sure what code page fopen uses for other compilers / OSes?
- What do you suggest?

https://msdn.microsoft.com/en-us/library/yeby3zcb.aspx
https://msdn.microsoft.com/en-us/library/windows/desktop/dd374130(v=vs.85).aspx
https://en.wikipedia.org/wiki/Windows-1252

joncampbell123:
In Linux there is no code page enforced at the system call level. On an ext4 filesystem filenames with any byte sequence are possible as long as they don't include NUL or forward slash.

Typically the only part of Linux to enforce encoding would be the filesystem driver when code page conversion is needed.

Windows on the other hand, subjects ANSI versions of the functions to whatever the current code page is (MS-DOS as well).

I was referring to the fact that your code has UTF16 - ANSI conversion functions at a glance and I naturally assumed this was for parsing text fields within VHD files since Microsoft formats are known to use UTF-16 for text strings.

Shane32:
The code is only used for converting the path names for fopen.

So if filenames are just byte sequences in Linux, are filenames with unicode characters typically stored as UTF-8 strings?

joncampbell123:
Usually, yes.

There's a locale environment variable that specifies the language and encoding. Most Linux systems today have that set to use UTF-8.

Shane32:
Here's what I'm thinking to fully support unicode:
- on Linux, use utf8_encode, and then convert \ to /
- on Windows, convert the child pathname from Windows-1252 to unicode, append the unicode path, and use _wfopen

But then I would need to know how to convert a string from Windows-1252 to unicode.  Which I'm not sure how to do.  I could leave it as ASCII-only on Windows for now.

Shane32:
I assume that you would you make the assumption that Linux PCs are using UTF-8 ?

joncampbell123:
I do. The guest to host filename conversion code assumes it, because modern Linux and Mac OS X systems use UTF-8 filenames.

joncampbell123:
At the same time, that code also assumes Windows 98 or later and the availability of the wide char (UTF-16) file API functions to accomplish the same.

joncampbell123:
The idea is that guest MS-DOS applications can use whatever code page they were written for and the filename translation will make that filename valid on the host. So if an MS-DOS program makes a filename with extended CP437 ASCII using graphical box characters or solid/gray blocks, then the filename will be converted to unicode to appear the same on the host.

Shane32:
Right.  In this particular case, the UTF-16 string came off of the host, and is being used on the host, and really doesn't have anything to do with the code page in use by the virtualized OS.  You can't even provide the string to imgmount as a parameter.  Which should make it easier to handle, as presumably the UTF-16 string represents a valid filename on the host.

Shane32:
Well, office emergency just came up.  I'll have to get back to this later.  Thanks for the help!!

Shane32:
See #690 

You can change the ConvertUTF16toASCII as necessary for Linux.  Since I don't have Linux, I can't be sure it works as intended.

PR #690, opened 2018-04-30
VHD differential drives support UTF-8 parent paths on Linux (untested)

Shane32:
… and ISO-8859-1 parent paths on Windows


Shane32:
See #687

PR #693, opened 2018-04-30
Miscellaneous cleanup to imageDisk classes

Shane32:
- reduce scope of variables not accessed outside of imageDisk
- delete unused variables
- protect virtualized software from overwriting past end of a mounted image (for example, a VHD file has a footer, which could be easily overwritten if the chs values were ignored or changed)
- enforce classes that inherit from imageDisk to define and use a unique class_id (what's the point of this variable anyway? let's delete it!)
- similar misc stuff

### DIFF
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -51,12 +51,14 @@ extern const Bit8u freedos_mbr[];
 
 class imageDisk {
 public:
-	enum {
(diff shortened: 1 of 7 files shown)


Shane32:
Some of the variables were changed (e.g. from 32-bit to 64-bit or otherwise), depending on what seemed appropriate.

PR #697, opened 2018-05-01
PC-98: A quick and dirty hack to force an IBM US-English like keyboard layout in BIOS

yksoft1:
Changed some key mappings that different between IBM and PC-98 layouts only.
Note: I think this code will only work for those games/apps that use BIOS for keyboard.

It's a dirty hack and you may ignore this if you think it's unacceptable.


joncampbell123:
It's OK. I'll merge it.

Issue #770, opened 2018-06-19
MOUNT command doesn't accept single quotes around path (original DosBox does)


Tenebrous:
MOUNT C 'C:\Temp' doesn't work and gives a "Can't create drive" error, although this works on baseline DosBox.

aybe:
Fixed

yksoft1:
Single quote is a legal character in Windows file names.
Assume there is a directory 

> D:\a\b\c\'d
If you try mounting it as

> mount c 'D:\a\b\c\'d'
Then the mounted C: would be D:\a\b\c instead of D:\a\b\c\'d as of current master head.

Tenebrous:
While I understand the difficulty, unfortunately this means that in some cases dosbox-x can't be used as a drop-in replacement to original dosbox (e.g. with LaunchBox for example).

May I suggest removing single quotes that appear at the start and end in a pair?

joncampbell123:
Another possibility might be to simply require double quotes and pass single quotes through without translation.

Tenebrous:
That doesn't get around the backwards compatibility issue with (for example) LaunchBox. However I'll raise an issue with them too to allow customisation of the MOUNT command they use.

aybe:
I asked for a review because I was reluctant of this change. 

I think we should just revert the change.

As far as I know, there is DOS in DOSBox and this is the commonly seen syntax for command line arguments.

It just makes sense to me to use delimiters that are forbidden in file names, it keeps things simple and at the same time avoids introducing subtle bugs unless significant rework of the parser is done, e.g. https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/.

joncampbell123:
I thought it was OK, but it doesn't seem to sit well with others here.

Tenebrous:
I'd hold off on this too tbh. I don't think my initial assessment of the problem was correct.

I was seeing the single quotes being passed as a -c arg to dosbox, e.g. -c "MOUNT C 'folder'"

Will get screenies and test some more when I'm at my desktop.

sikthehedgehog:
Quoting myself from the PR:

> And on Linux (and other POSIX systems), everything that isn't the nul character can show up in a path (the only two reserved characters are nul and /, with the latter being the path separator so it can show up in a full path). Given that DOSBox-X supports Linux, this needs to be taken into account.

[...]

PR #772, opened 2018-06-19
Added single quote support

aybe:
**MOUNT command doesn't accept single quotes around path (original DosBox does) #770**

While this easy to fix, I am against that change because ```'``` (single quote) is a **valid** character for a file name:

[8.3 filename](https://en.wikipedia.org/wiki/8.3_filename)

[Naming Files, Paths, and Namespaces](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx#naming_conventions)

### DIFF
--- a/src/misc/setup.cpp
+++ b/src/misc/setup.cpp
@@ -1335,7 +1335,7 @@ CommandLine::CommandLine(char const * const name,char const * const cmdline,enum
     opt_style = opt;
     while ((c=*c_cmdline)!=0) {
         if (inquote) {
-            if (c!='"') str+=c;
+            if (c!='"'&&c!='\'') str+=c;
             else {
                 inquote=false;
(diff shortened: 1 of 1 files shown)


sikthehedgehog:
> While this easy to fix, I am against that change because `'` (single quote) is a valid character for a file name:

And on Linux (and other POSIX systems), everything that isn't the nul character can show up in a path (the only two reserved characters are nul and `/`, with the latter being the path separator so it can show up in a full path). Given that DOSBox-X supports Linux, this needs to be taken into account.

If that's an issue then I think it'd be time to start thinking on ways to deal with it (repeating the quote is a popular solution when you don't need to escape any other character)

aybe:
This PR might have been merged too quickly, the discussion is ongoing at #770.

PR #780, opened 2018-06-22
SHELL: added word navigation with Ctrl+Left/Right.

aybe:
This was bugging me for a while, I've struggled to the point thinking I needed to bring in int10 cursorpos and prompt length but in fact my code was only missing the last ```str_index--;``` X)

It does just like cmd does, jump to beginning of words, or to edges when close.

Note that original cmd does not interpret quotes so if you have ```abcd "efgh ijkl"```, the separation is reached between ```efgh``` and ```ijkl```, sounds good IMO.

I have tested it for a while but please do it too.

### DIFF
--- a/src/shell/shell_misc.cpp
+++ b/src/shell/shell_misc.cpp
@@ -203,6 +203,51 @@ void DOS_Shell::InputCommand(char * line) {
                 }
                 break;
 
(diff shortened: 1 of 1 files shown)


joncampbell123:
Looks good to me. It only adds scan code handling without modifying anything else.

PR #783, opened 2018-06-22
Caret fix

aybe:
### Partially works

The movement works well,

But writing does not, here on last command I type first line full of Q's, second line full of W's, I step back five chars then type E and you see the result:

dosbox_001

### The default scan code handler seems to do some questionable stuff:

Not sure what to do here ....

https://github.com/joncampbell123/dosbox-x/blob/7bc4de0e85408d408511b42268e30444fe72cccf/src/shell/shell_misc.cpp#L528

And same with ```25/28/50.com```, also the missing cursor on 50 does not help.


joncampbell123:
There may be some explanation for that. It's possible the original programmer had to do that for some hackish reason.

If all SetCaretPos() does is move the cursor up or do nothing, why is it SetCaretPos() and not MoveBackAndPossiblyUpALine() ?

aybe:
Okay, I'll dump it and start from scratch, 

LOL because it was for avoiding such naming scheme, probably better to do:

```
void CaretPosition(&row, ...)
void CaretPositionFixUp()
```

Any idea on why 50 row mode doesn't show cursor, or is another trap in disguise ?

aybe:
This couldn't forget to bring INS/OVR mode, getting closer ...

joncampbell123:
Let me know when you're ready for me to merge.

aybe:
Yes I will !

aybe:
Also, while fixing this, I found out that per-edge word-navigation was possible.

Sounded cool but requires you to smash your keyboard twice as much as regular, so I didn't bother.

joncampbell123:
Looks good to me!

yksoft1:
I had made INT10_GetScreenColumns and INT10_GetCursorPos working in PC-98 mode, however I can't make it recognize CTRL+Left or CTRL+Right....

aybe:
I guess that if you can find BIOS docs for it then it should be relatively easy to do.

(couldn't find any actually ...)

aybe:
Repository for NEC PC-98 stuff #787

Some docs but all Japanese .............................................

aybe:
This becomes 0x20 when INS is on:

capture

But what address is this ?

aybe:
Dunno what I had mind ... this was irrelevant.

So if this combination can't be recognized, maybe this is because the scan code is different, however I can't figure out how to use pc98 emulation, does it need a BIOS ?

aybe:
Ok I think I need some rest I guess : ```#machine=pc98``` for sure does nothing !

aybe:
@yksoft1 

Here,

https://github.com/joncampbell123/dosbox-x/blob/develop/src/shell/shell_misc.cpp#L148

There should be some code that translates ```cr``` when <kbd>Ctrl</kbd> is down.

@joncampbell123 

PC98 docs for this, where are they ?

joncampbell123:
@abye I have test code in DOSLIB that can run on PC-98 and return keyboard scancodes.

I'll pull out one of my PC-98 units and see if ctrl+arrow keys is supported.

To answer one of your questions, @aybe, PC-98 does not have console TTY functions in the BIOS. The PC-98 BIOS provides the base video modesetting functions and MS-DOS provides the TTY functions.

aybe:
So what does that mean ? I don't get it.

joncampbell123:
@aybe The IBM PC BIOS provides modesetting and TTY functions in the BIOS. The PC-98 BIOS only provides the modesetting, MS-DOS provides the console driver for TTY functions. That means MS-DOS is the one that handles printing to the console and manipulating the cursor, not the BIOS.

joncampbell123:
@yksoft1 I incorporated the code changes you posted.

As far as I can tell, given the INT 18h test program on a PC-98 laptop, the BIOS does not return any different scan code for left/right arrow and CTRL+left/right arrow.

Left and CTRL-left both return scan code 0x3B00.

One possible hack to allow detecting CTRL-left/right might be to use INT 18h to read the keyboard "status" byte where bit 4 reflects whether or not CTRL is held down when reading the scan code from the BIOS. Then the left/right arrow code can act on that information.

joncampbell123:
Please try the latest commit.

joncampbell123:
The latest commit should provide full left/right, CTRL+left/right, ESC, and INS key handling in both IBM PC/AT and NEC PC-98 mode.

aybe:
Well done 👍

PR #792, opened 2018-06-26
VFRCRATE: Added help and improved output.

aybe:
Didn't really understand the purpose of [fraction syntax](https://github.com/joncampbell123/dosbox-x/compare/develop...vfrcrate?expand=1#diff-d75b1308381ee4c49dd744fa57b1edadR433) so I didn't add it to the help ...


joncampbell123:
The fraction syntax is there to allow fractional frame rates, such as HD 720p standard 60000/1001 instead of 59.94.

aybe:
Added

PR #793, opened 2018-06-26
XBRZ

aybe:
I am pushing this so it doesn't rot.

Couldn't assert that all is correct but it does work indeed.


joncampbell123:
I've merge the latest code changes into your xBRZ branch.

I have a few issues with this code yet.

First of all, my Linux system does not have "ppl.h" which this branch seems to require.

Second, I'm concerned that xBRZ support code is written without consideration that the developer compiling the code might want to disable or enable it. There's no C_XBRZ #define in config.h for the developer to comment out if they don't want it or in case the xBRZ scaler code is causing problems in the compile stages.

If you add #if C_XBRZ around the code so that a proper compile can be done with or without the xBRZ scaler (without breaking the GUI or output code), I'll do my part in this branch to add the ppl.h detect and configure option to enable/disable xBRZ in configure.ac for Linux systems, and then this code would be appropriate to merge into master.

joncampbell123:
On my Linux system with GCC 4.8, the missing ppl.h is the only compile error, all other code appears to compile fine.

joncampbell123:
If you're wondering why someone would want to disable xBRZ on compile, think about those who may be concerned about EXE bloat, or compile DOSBox-X for older systems (including the official lowend MinGW builds where MMX/SSE has been disabled). There's also the HX DOS builds where any parallel processing in this code is pointless because I doubt the HX DOS extender would have any multi-core support in it's Win32 API emulation.

aybe:
Makes sense, will take a look.

aybe:
I don't understand how to implement C_XBRZ when for instance you have something like that:

https://github.com/joncampbell123/dosbox-x/pull/793/files#diff-3b1f3c818385e56283d07b63d93a2ee7R3240

?

joncampbell123:
Like this:

#if C_XBRZ
if (sdl.xBRZ.whatever) {
}
else
#endif
    non_xbrz_stuff()

aybe:
I think I've got it right, both versions do build and run.

Two things,

It only runs with the following config, @alexat says it runs windowed but I couldn't make it work.

```
[sdl]
fullscreen=true
output=surface

[render]
scaler=xbrz_bilinear
```

And ```aspect``` is forced no matter what.

alexat:
[render]
output=surface
scaler=xbrz_bilinear

image
image
image

Just for clarity: are you building under Windows 10 system? I need to somehow reproduce it not working under some configuration, then I may be able to pinpoint what's amiss with how scaler/window parameters should change at runtime.

aybe:
Yes Windows 10, visual studio 2017, HD 7850.

________________________________
De : alexat<mailto:notifications@github.com>
Envoyé : ‎27/‎06/‎2018 09:18
À : joncampbell123/dosbox-x<mailto:dosbox-x@noreply.github.com>
Cc : Aybe<mailto:aybe.one@hotmail.com>; Author<mailto:author@noreply.github.com>
Objet : Re: [joncampbell123/dosbox-x] XBRZ (#793)


[render]
output=surface
scaler=xbrz_bilinear

[image]<https://user-images.githubusercontent.com/29947/41958545-09b9b080-79f3-11e8-92fc-b76d068f1166.png>
[image]<https://user-images.githubusercontent.com/29947/41958592-2d2764b8-79f3-11e8-9ef5-351858e2ffcb.png>
[image]<https://user-images.githubusercontent.com/29947/41958630-4c555fde-79f3-11e8-9e25-10c3c416822f.png>

Just for clarity: are you building under Windows 10 system? I need to somehow reproduce it not working under some configuration, then I may be able to pinpoint what's amiss with how scaler/window parameters should change at runtime.

alexat:
Totally strange. Windows 10, VS 2017 Community. 

I know it's totally bad to offer, but can you please somehow try to test with my binary (better in an isolated environment)? Maybe something's different with my build environment then, if this runs, I'll try to reproduce in a clean VM with only VS2017 installed. I've built this for x86 target just from GIT sources 
without any special changes I usually do for myself (AVX2/LTCG).

Please note that it requires at least X2 'real' window size in surface mode to start scaling (this can be adjusted in code if seen viable). In direct3d mode it should scale along with window size despite changing aspect, but scaler will again be enabled only at X2 the size.

[dosbox-x-test.zip](https://github.com/joncampbell123/dosbox-x/files/2140268/dosbox-x-test.zip)

aybe:
In your binary it's just as the regular one,

Tonight I will take a look and debug see if I can find the cause.

aybe:
Your code works fine, it's me that didn't understand how to use it:

2018-06-27 21_56_16-dosbox-x 0 82 7 30000 cyc_ms dz

2018-06-27 21_56_52-dosbox-x 0 82 7 30000 cyc_ms dz

It is not obvious to someone that window has to be twice as big (even though you wrote it in last message, I missed it).

What I think you should do:

 - resize window automatically if possible else output a message in the log
 - write a README.xbrz that explains it in details, should be linked to from README

With one or the other, it's less likely to be misused.

aybe:
And by the way, isn't it possible to make it work even when ```aspect=false``` ?

Or does it look so ugly that you've chosen not to ?

Thanks

alexat:
@aybe 
I am thinking about making it scale to 2x at anything over 1x but less than 2x window. It's quite trivial, just need to see if xBRZ+downscaling gives some benefit over just no scaling.

For the other part, you mean occupying only part of window padding it with black boxes? That's how it works in surface mode, in D3D mode D3D part still scales result to full window size, invalidating aspect ratio in the process. I need to re-check code paths of both to see if proper aspect correction can be applied to both, maybe having its own xbrz_aspect= flag).

aybe:
It should be obvious when xbrz is enabled so, you could either double the window size enough so it is visible if the screen can handle it; or write some documentation so one knows everything that is needed for proper operation, i.e. are there cases when it doesn't work ?

I was going to suggest a place like this but unfortunately there's none for scalers:

https://github.com/joncampbell123/dosbox-x/pull/793/files#diff-6145266fb539c7f572a075978cc12cdaR1451

So maybe a proper section in the README.xbrz would be best.

---

For the aspect I meant, why doesn't it work when aspect=false ? 

Like this but twice the size obviously:

2018-06-27 21_56_16-dosbox-x 0 82 7 30000 cyc_ms dz

joncampbell123:
I've maintained the code where aspect=false means to stretch to fill the window (even if distorting the image). I've modified the D3D code however to support aspect=true in the same way that OpenGL does.

aybe:
Great, will test that tomorrow.

alexat:
README.xbrz (probably needs some corrections)

[README.xbrz.txt](https://github.com/joncampbell123/dosbox-x/files/2144087/README.xbrz.txt)

alexat:
@joncampbell123 
Are the modifications pushed to some branch so I can merge it and test along?

I have made my own variant of D3D aspect ratio correction for xBRZ. Currently, I pad the output texture with black bars to maintain aspect ratio. The problem is, it only starts working when xBRZ does, so when scale factor is <=1, it does not do anything, causing 'jumpy' behavior.

@aybe 
I tested enabling 2X xBRZ scale for anything between 1..2 and it is okay, but with nearest neighbor, it really degrades quality compared to native D3D upscaling :) With bilinear it looks good though. Won't do any pull requests yet, need to sort out D3D/OpenGL aspect correction first.

alexat:
@joncampbell123 @aybe 
Okay, I'll be doing a next pull request in few minutes to this branch.

joncampbell123:
@alexat The code is written so that the DOS screen completely fills the texture, then OpenGL and D3D draw the texture according to aspect ratio. The padding is NOT done within the texture.

joncampbell123:
There's a merge conflict at the moment with src/gui/sdlmain.cpp between xbrz and master, would it be possible to resolve it in your xbrz branch?

alexat:
Took a bit of time to rework aspect rate correction code for Direct3D/OpenGL so it now works properly assuming square pixels, and added a bit of the same to 'surface' with xBRZ enabled. 

#805

alexat:
@joncampbell123 I've resolved conflicts in #805, but I don't know if I should have done it like that, merging #805 branch with master.

PR #795, opened 2018-06-26
Mouse Integration DOS ? (too easy to be true)

aybe:
After fiddling failed with ```INT33_Handler``` since DOSZIP only triggers mouse release (0x06), I decided to try another approach and to much of my surprise it works !

What I've done is that I've added ```synced``` in conf to ensure good results, in fact ```bool emulate``` parameter upstream in the call stack could have been changed but I didn't bother since it could break something else.

With the following I get VM-like integration:

```
[sdl]
autolock=false
synced=true
```

I didn't really test beside DOSZIP but it works well,

PS this still suffer of poor handling where when you move away the mouse too fast, cursor is not exactly at the edge you exited.

What do you think ?

PPS basically logic was already in but it was ```bool emulate``` that was resulting in poor UX (try with ```synced=false``` to see what I'm talking about)

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -545,6 +545,7 @@ struct SDL_Block {
     struct {
         bool autolock;
         bool autoenable;
+        bool synced;
         bool requestlock;
         bool locked;
         Bitu sensitivity;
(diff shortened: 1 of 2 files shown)


aybe:
Also, you can see that event pumping is struggling to keep up with mouse position when you burst your hand a bit, didn't really know what to do about it.

aybe:
And maybe we could do the following:

When

`autolock` = true -> `synced`= false

`autolock` = false -> `synced`= true

And just hide `synced `to the user.

aybe:
It is slightly faster in Release but still noticeable.

aybe:
Okay last thing, basically if you're happy about it, that whole thing could be summarized as:

``` 
if(!autolock)
    sync cursor
```

aybe:
1. showing the cursor was just for debug and will be removed because if you look well, it fails with it after pausing your hand for a bit (gets moved to center and that delta is added).

2. too I think it's the right behavior but the problem currently is that when 

```
[sdl]
autolock=true
synced=true
```
this syncing doesn't work.

To fix this it needs significant rework of that block, movement too should happen in last ```else``` or did I miss something ?

PR #796, opened 2018-06-26
Mouse integration fix

aybe:
On the last commit,

- I refactored by moving ```Mouse_CursorMoved``` params above, 
- this allowed me to entirely remove ```if(sdl.mouse.locked || !sdl.mouse.autoenable)``` condition
   which prevented syncing to work as we defined
- the whole block is simpler
- added a helper for showing cursor only when hooked by app
- removed ```[sdl] synced```, was very confusing because would work only when ```autolock=false```

i.e. behavior is the same, synced works as expected, simpler code and config

Let me know how that works !

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,12 +18,12 @@ TODO add version
   - Improve mouse support:
     - Added pointer integration behavior for DOS applications.
     - Auto-lock is now disabled by default, this results in
-	  pointer integration becoming the default behavior.
+      pointer integration becoming the default behavior.
     - Added feedback for auto-lock state, can be visual or auditive.
     - When the mouse is not locked, its position is exactly where
-	  the mouse is reported by the system. This results in a more
-	  consistent experience between DOSBox-X and host OS,
-	  e.g. High DPI mouse don't become suddenly slow in DOSBox-X.
+      the mouse is reported by the system. This results in a more
+      consistent experience between DOSBox-X and host OS,
+      e.g. High DPI mouse don't become suddenly slow in DOSBox-X.
     
 0.82.7
   - Mac OS X builds now honor showmenu=false by leaving the
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,15 @@ TODO add version
     - Axes can be remapped for devices with questionable layout.
     - User-settable deadzones for joystick bindings in mapper,
       mappings like WSAD keys to axes is less frustrating.
+  - Improve mouse support:
+    - Added pointer integration behavior for DOS applications.
+    - Auto-lock is now disabled by default, this results in
+	  pointer integration becoming the default behavior.
+    - Added feedback for auto-lock state, can be visual or auditive.
+    - When the mouse is not locked, its position is exactly where
+	  the mouse is reported by the system. This results in a more
+	  consistent experience between DOSBox-X and host OS,
+	  e.g. High DPI mouse don't become suddenly slow in DOSBox-X.
     
 0.82.7
   - Mac OS X builds now honor showmenu=false by leaving the
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -6207,7 +6207,7 @@ void SDL_SetupConfigSection() {
     Pbool->Set_help("Mouse will automatically lock, if you click on the screen. (Press CTRL-F10 to unlock)");
 
     const char* feeds[] = { "none", "beep", "flash", nullptr};
-    Pstring = sdl_sec->Add_string("autolock_feedback", Property::Changeable::Always, feeds[0]);
+    Pstring = sdl_sec->Add_string("autolock_feedback", Property::Changeable::Always, feeds[1]);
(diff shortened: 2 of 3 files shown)


aybe:
Also, there are two things that should be addressed in the future:

- when a mouse event occurs in SDL pump, something like SDL_GetGlobalMouseState should be called after and fix up the event data accordingly, this will fix the problem when you move away too quickly the cursor still reaches edge in the window

- not a big deal, when you switch mouse on/off (i.e. start/exit a mouse app) when not locked, first movement can look weird because mouse is not where app expects it, but we have no control on that since some app centers the cursor, some do not (really cosmetic)

aybe:
Lastly, when unlocking mouse and therefore should be visible, this does not happen until user moved a bit.

aybe:
_when a mouse event occurs in SDL pump, something like SDL_GetGlobalMouseState should be called after and fix up the event data accordingly, this will fix the problem when you move away too quickly the cursor still reaches edge in the window_

I haven't found a way to get cursor desktop position in SDL event polling loop, of course, it gets triggered only when mouse is over, leaving it for now.

_Lastly, when unlocking mouse and therefore should be visible, this does not happen until user moved a bit._

Same, detecting mouse state with 'hidden' is not reliable.

aybe:
@joncampbell123 

I have fixed the outside movement not being registered and left space for implementing non-Windows platforms.

Works great so far.

aybe:
### Big changes !

Now that integration works well now, had to try address bugs and improve the rest.

Double pointer issue is gone but I ensured that either is always visible because when app did not use the mouse, none were seen and it was very confusing if user wants to go to the menu for instance.

The big change is that now ```autolock=false``` by default, I believe this is definitely better and I'll explain.

When you use DOSBOX for the first time, it's baffling as on why it swallows your mouse. Most of the time I find myself pressing <kbd>Win</kbd> or <kbd>Alt</kbd>+<kbd>Tab</kbd> just because I want to regain my mouse DOSBOX just stole !!  For changing something in the menu or even just maximizing the window before gaming, very boring to be honest.

Now it is like in VMware, you can move freely by default which is _definitely better_ UX IMO.

Before the change, it was frustrating as mentioned above and basically hides that friendly behavior this PR is all about.

Hope you'll keep it as the default behavior ! I'd suggest you to do A/B builds comparison to see the differences.

The other big thing I've added is notification of autolock status, it can be visual, auditive or disabled. It is triggered whenever appropriate, lost focus, mouse click, <kbd>Ctrl</kbd>+<kbd>F10</kbd>.


I guess that's all for now !

aybe:
PS

To me you can merge **after testing** because I think it's good enough,

For the two other bugs I won't bother,

- it's impossible to know where an app puts the mouse after it has started, some center, some do not
- my way of detecting whether the mouse is in use seems to work well after all

aybe:
PPS

You might wonder why I came up with that notification stuff ?

It's because since I've fixed the dual pointer issue, it was not obvious anymore that mouse was un-locked when an app uses the mouse (the SDL cursor was not visible anymore).

Then I tried to come up with something cool and configurable at the same time.

joncampbell123:
@aybe I'll patch up the remaining issues. When you write these patches please keep in mind that Windows is not the only platform DOSBox-X compiles to.

aybe:
Glad you like it !

aybe:
I will look at adding relevant stuff for SDL2,.

And try address the Linux part if I can get it to compile 😃

aybe:
I'll see what I can do for adding an option to disable this entirely.

aybe:
@yksoft1 

Here's your switch, for old-school mode use:

```
[sdl]
mouse_emulation=always
```

aybe:
[This is my definitive commit to this feature:](https://github.com/joncampbell123/dosbox-x/commit/2ea5c7a6bb4007a8c895232345774eec600c9539)

Now, mouse emulation is entirely configurable, as on whether you want it or not and when:

```
+        "When is mouse emulated ?\n"
+        "integration: when not locked\n"
+        "locked:      when locked\n"
+        "always:      every time\n"
+        "never:       at no time\n"
+        "If disabled, the mouse position in DOSBox-X is exactly where the host OS reports it.\n"
+        "When using a high DPI mouse, the emulation of mouse movement can noticeably reduce the\n"
+        "sensitiveness of your device, i.e. the mouse is slower but more precise.");
```

@yksoft1 

For old-school mode:

```
[sdl]
mouse_emulation=locked
```

@all

By default it behaves as follows:

- integration: no emulation, consistent with OS
- locked: emulated, better precision (old behavior kept)

So as usual except that when not locked it is better that what it used to be.

yksoft1:
Why you don't disable synced movement if you set MOUSE_EMULATION_LOCKED or MOUSE_EMULATION_NEVER?

aybe:
When you use locked mode it's exactly as how it used to be, emulation when locked, no emulation when not locked.

I tested every case before pushing.

aybe:
If you have a high dpi mouse, crank it to the max and you will clearly spot when it is emulating vs when it is not.

PR #798, opened 2018-06-27
Ins key cursor update occurs only during shell

aybe:
Is this the right way ?

No more pass-through when a program runs, does not get restored when program exits, though.

I checked with 6.22 on a VM, it doesn't get restored too.

Funnily, it's the opposite in 6.22, small cursor on OVR, big cursor on INS.

### DIFF
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -424,6 +424,7 @@ static Bitu DOS_21Handler(void) {
         case 0x00:      /* Terminate Program */
             DOS_Terminate(mem_readw(SegPhys(ss)+reg_sp+2),false,0);
             if (DOS_BreakINT23InProgress) throw int(0); /* HACK: Ick */
+            dos_program_running = false;
             break;
         case 0x01:      /* Read character from STDIN, with echo */
             {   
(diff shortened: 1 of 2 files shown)


aybe:
Not sure whether all these should set that boolean ?

http://stanislavs.org/helppc/int_21-0.html
http://stanislavs.org/helppc/int_21-4b.html
http://stanislavs.org/helppc/int_21-4c.html
http://stanislavs.org/helppc/int_21-26.html
http://stanislavs.org/helppc/int_21-31.html

joncampbell123:
I don't really know to be honest.

If you have a working MS-DOS bootdisk, you could boot it up and run DOSKEY to see what DOSKEY does with the flag after each call.

joncampbell123:
A lot of the editing controls at the shell so far resemble MS-DOS command interpreter behavior when DOSKEY is resident.

DOSKEY editing behavior seems consistent across IBM PC and PC-98 MS-DOS.

aybe:
Okay I'll try.

aybe:
That's not what I was looking for in fact, my question was rather:

Do programs only raise 4C or can they raise 00 or 31 when exiting ?

If so then I'd set this boolean too in these places.

joncampbell123:
All three exit paths exist on MS-DOS.

DOS programs can exit with INT 20h, INT 21h AH=0, and INT 21h AH=4Ch, they all work.

aybe:
Alright

PR #805, opened 2018-06-28
Xbrz

alexat:
- Quite extensive aspect ratio correction fix. It assumes 'real' pixels in both source/output are square (should be the case). Not really xBRZ related, but well, it was due to xBRZ + aspect behavior.
- xBRZ now honors aspect ratio correction as well. 
- For modes allowing for native aspect ratio correction (D3D/OGL/xBRZ), render-based one is not used anymore.

Please test. I've checked all the possible combinations of output + aspect parameters, plus xBRZ scaler in each of combination, + some other scalers (normal2x/3x/advmame2x/hq2x forced) in non-xBRZ modes to make sure I did not break anything. Checks pattern included different resolutions: text modes 80x25 and 80x50, 320x200, non-standard (Lost Vikings), resizing back and forth, going to fullscreen and back from default and non-default sizes.

Related options:
[sdl] output=surface/direct3d/opengl(nb/hq)
[render] aspect=true/false
[render] scaler=xbrz/xbrz_bilinear

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,8 +1,11 @@
 TODO add version
-  - Improved shell:
+  - INT 10h AH=10h now ignores AL=3 in PCjr mode.
+  - Fixed keyboard handler bug in PCjr mode that caused some CPU
+    register corruption and general crashiness in games.
+  - Improved shell: (Aybe, Joncampbell123)
     - Ctrl+Left and Ctrl+Right permits word-navigation.
     - Added emulation of 'Ins' key behavior.
-  - Num Lock stays on at startup and is synchronized with host
+  - Num Lock stays on at startup and is synchronized with host (Aybe)
     when DOSBox-X window gains focus (Windows). (Aybe)
   - Added visual feedback to Hat/D-pad buttons in mapper. (Aybe)
   - Added documentation for 'dir' command sorting switches. (Aybe)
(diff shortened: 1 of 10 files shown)


alexat:
Also merged master branch resolving trivial conflicts.

joncampbell123:
I just compiled DOSBox-X on VS2017. Selecting the xBRZ scaler, even if I force the scaler, doesn't appear to do anything.

alexat:
Please share a config file, maybe we have to add something more to internal requirements. I use a very clean config though myself:

[sdl]
output=surface
;output=direct3d
;output=opengl
;aspect=true

[render]
scaler=xbrz_bilinear

On extending window/going fullscreen, it does the job )

joncampbell123:
I started DOSBox-X with no configuration file whatsoever, then selected the Surface output and the xBRZ scaler while still at the DOS prompt. Nothing.

alexat:
Changing output / scaler in the menu does not cause full GUI reinitialization I presume?
I indeed do not even see output mode change from direct3d, could it be changed at all?
I've not digged to DOSBox-X menu option yet, but will try to see what it does.

alexat:
Okay, this is related to how configuration is read / applied because xBRZ is more 'sdlmain' code having options in 'render' section. I'll take some time to fix the realtime configuration adjustment, it will require xBRZ structure to be moved to render.h and some early initialization routine added.

alexat:
@joncampbell123 
Adjusted online reconfiguration from menus for xBRZ scaler and output mode changes.

PR #809, opened 2018-06-29
Fix a single forgotten conditional

### LINKED ISSUE #808: Current code stretches text vertically
This was built under Windows 10, 29 June 2018, 6:44 AM EST:
capture

alexat:
Caused automatically selected window size to be always aspect corrected under D3D output mode.
Fixes #808


alexat:
All other modes do have this conditional, this is the single path that was lacking it.

PR #814, opened 2018-06-29
Support online reconfiguration for the new features (xBRZ and aspect offload) from menu

alexat:
As mentioned in #805 , online configuration possibiity is indeed required so it's done.

Please see commit doc for the change details. Most of the small changes are sdl.xBRZ to render.xBRZ due to structure re-placement.


joncampbell123:
That did it, thanks!

PR #816, opened 2018-06-30
[for evaluation only] Attempt on factoring out output code from sdlmain.cpp.

alexat:
**This is *NOT* for immediate merge, just for evaluation.**

sdlmain.cpp is already a mess. Lately, with all the xBRZ and aspect correction changes, I contributed more to it becoming unreadable in output code paths, with almost identical (scaling) parts present. And so here I'm trying to partially refactor output code, to make it easier to grasp.

If it's okay to do it like it's done with surface output, I will continue, eventually splitting all significant output code from sdlmain.cpp into separate files per surface/direct3d/opengl with identical set of exported output related functions, increasing code reuse for xBRZ and maybe other stuff. The only thing, the result would probably have to have minor adjustments by someone more knowledgeable in regards to portability.

Please see commit description and diff for change details. No functional changes done except reverting commits that broke xBRZ fully filling up surface output.


joncampbell123:
Aargh, your changes undid the code needed to prevent xBRZ from overdrawing the SDL drawn menus. Other than that, works for me.

I merged your changes into a sdlmain_refactor branch in DOSBox-X along with the changes needed to compile under Linux on my system.

https://github.com/joncampbell123/dosbox-x/tree/sdlmain_refactor

joncampbell123:
I'll see if I can fix that on my end.

Please merge from https://github.com/joncampbell123/dosbox-x/tree/sdlmain_refactor into your branch to keep thing synced.

joncampbell123:
I like the cleanup, because it untangles a lot in sdlmain.cpp however there's a reason I modified the xBRZ output code to use the sdl.clip rectangle.

I've been maintaining the code to determine how the frame fits at GFX_SetSize(), and then store that in sdl.clip (sdl.clip.x, sdl.clip.y, sdl.clip.w, sdl.clip.h). All output code is supposed to draw and scale the image to fit that rectangle. That way the scaling and positioning is consistent.

The consistent rule I'm trying to keep in DOSBox-X is:

- if aspect=true, the image is scaled and fitted to the window to keep aspect ratio. Either width or height is reduced so that nothing is cut off, but the image fills one of the two dimensions (x or y) of the window.
- if aspect=false, the image is supposed to scale to fill the window entirely even if it distorts the image.

The other purpose of using sdl.clip is to make it possible for the SDL drawn menus to work. In Linux where neither Windows style nor Mac OS X style menus are around, DOSBox-X has it's own custom drawn menus that it draws at the top of the SDL surface. When that is done, sdl.y is adjusted downward to make room for those menus.

Your xBRZ code as it stands now draws over or covers up those menus because it ignores sdl.clip.

That's why I replaced clipWidth, etc. assignments now in src/output/output_surface.cpp with assignments that directly pull from sdl.clip.

Again, if aspect=false, the sdl.clip rectangle should represent a region (especially in Windows) that completely covers the client area of the window, and in Linux, should cover the client area except the menu drawn at the top.

joncampbell123:
One more change that I made, was to have output=surface use the same aspect ratio fitting code as output=opengl if xBRZ was selected in order to fit to the screen instead of sitting in the center as surface normally does.

alexat:
I see. Well, clip size precalculation can be modified for xBRZ then so xBRZ scaler code uses it to fit in the full size, without double-correcting at all. Sorry for not following exactly, did you do the change? If not, I'll do just that with the next pull request.

joncampbell123:
I've made some changes in this repo, in branch sdlmain_refactor.

joncampbell123:
I'm working on a cleaner re-implementation of the surface fit to window logic I had before if xBRZ is in use.

alexat:
Basically it's as follows for xBRZ: we take full size minus menu, put it into clip w/h. Then if aspect correct is enabled, we correct w/h and x/y to provide aspect corrected 4:3 window for xBRZ to fit in.

joncampbell123:
That's exactly what I have in mind. Unless changed, the OpenGL and Direct3D code do the same already.

alexat:
Yeah, they do just that by design for the texture size and fit. The only difference is that for surface, it's xBRZ code that is scaling the output 'texture'.

joncampbell123:
Pull the latest code, I've re-added fitting to the window for output=surface and xBRZ. I think it should be clean enough.

Again, I like the separation of outputs into separate CPP files, that's good cleanup!

alexat:
Ok )

I'm now splitting OpenGL and D3D stuff out, will push then. This won't be the end of it, I see i.e. aspect correction code is duplicated, will probably move some generic routines to output_tools.h/cpp later on.

aybe:
One of the ideas I've had about refactoring was something like the following:

pseudo-code:

```
class scaler (abstract)
class xbrzscaler : scaler
etc ... when a scaler is not available to current platform, the fallback would be last valid one

scaler current = ...; // assigned where appropriate

then in these huge methods:

current.blit(some state struct);
```

I think you get the idea, we'd end with much shorter and easier to understand methods.

(we'll see if your attempt is better than mine when you're done)

alexat:
My current code split run indeed has future conversion to classes in mind. It's just it would be too hard to undertake in one go :) For post-scalers, xBRZ integration indeed can be generalized as well.

joncampbell123:
I like the idea of a scaler C++ class, especially if it helps clean up the way scalers are defined currently in the code.

The current render scaler code is a tangled mess of header files that include one another with #defines to modify behavior each time.

That code would be ripe for cleanup if the scalers were written with templating in mind. Templating by pixel format, for example. That can be done if you declare a packed structure that matches a pixel format type with bitfields to match the R/G/B fields and const methods to return information on R/G/B range. I've written such code in a professional product before where YUV->RGB conversion or bilinear interpolation is involved.

joncampbell123:
Of course similar code could be grouped together if other template parameters are provided.

joncampbell123:
Another benefit of declaring structs and templating per pixel format of course is that adding pixel formats becomes less difficult.

alexat:
xBRZ is totally not a good example of that, it's color space affixed.

Another thing with it... From the DOSBox code perspective currently, xBRZ is in no way a scaler though. It's a kind of output extension, 'post-scaler'.

joncampbell123:
@alexat That's true. Then said revision should define "scaler" and "postScaler" C++ classes :)

joncampbell123:
I'm impressed so far with the xBRZ scaler as implemented now. Mind if I tweet that DOSBox-X has xBRZ? I have a fair number of followers who like to hear about DOSBox-X and old-school DOS stuff in general.

alexat:
For me, I totally don't mind, xBRZ makes DOSBox-X even more cool :)
I wonder why long standing xBRZ code was not taken into DOSBox except for (now obsolete) Daum builds.

PR #817, opened 2018-06-30
Continuing on sdlmain.cpp cleanup: OpenGL part

ghost:
This is OpenGL part and some preparation/generalization for Direct3D part.

Many supplementary OpenGL related code still remains in sdlmain.cpp, this needs to be assessed and moved out as well but later, after all primary output code paths will be out.

Some D3D related files moved, correction for non-VS builds makefiles would be needed.

### DIFF
--- a/vs2015/dosbox-x.vcxproj.filters
+++ b/vs2015/dosbox-x.vcxproj.filters
@@ -279,12 +279,6 @@
     <ClCompile Include="..\src\fpu\fpu.cpp">
       <Filter>Sources\fpu</Filter>
     </ClCompile>
-    <ClCompile Include="..\src\gui\direct3d.cpp">
-      <Filter>Sources\gui</Filter>
-    </ClCompile>
-    <ClCompile Include="..\src\gui\hq2x_d3d.cpp">
-      <Filter>Sources\gui</Filter>
-    </ClCompile>
     <ClCompile Include="..\src\gui\menu.cpp">
       <Filter>Sources\gui</Filter>
     </ClCompile>
@@ -295,9 +289,6 @@
       <Filter>Sources\gui</Filter>
     </ClCompile>
     <ClCompile Include="..\src\gui\render_scalers.cpp">
-      <Filter>Sources\gui</Filter>
-    </ClCompile>
-    <ClCompile Include="..\src\gui\ScalingEffect.cpp">
       <Filter>Sources\gui</Filter>
     </ClCompile>
     <ClCompile Include="..\src\gui\sdl_gui.cpp">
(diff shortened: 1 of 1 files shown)


aybe:
Currently code does not compile but I believe this is expected ?

joncampbell123:
It compiles here on Linux.

I fixed the aspect ratio code to take the aspect ratio from the VGA emulation rather than assume 4:3.

What compiler warnings/errors are you getting?

alexat:
Compiles at least on VS2017 @ mine. Please share error you get indeed.

aybe:
```
3>sdlmain.obj : error LNK2019: unresolved external symbol "unsigned int __cdecl OUTPUT_SURFACE_GetBestMode(unsigned int)" (?OUTPUT_SURFACE_GetBestMode@@YAII@Z) referenced in function __catch$_SDL_main$1
3>sdlmain.obj : error LNK2019: unresolved external symbol "unsigned int __cdecl OUTPUT_SURFACE_SetSize(void)" (?OUTPUT_SURFACE_SetSize@@YAIXZ) referenced in function __catch$_SDL_main$1
3>C:\Dev\dosbox-x\vs2015\..\bin\Win32\Debug\/dosbox-x.exe : fatal error LNK1120: 2 unresolved externals
```
(develop branch)

alexat:
Hm, are output_surface_sdl.cpp / output_surface_sdl2.cpp appear in the source list / properly compiling, these functions are in these two.

aybe:
I only have the declaration functions of both methods.
https://github.com/joncampbell123/dosbox-x/blob/develop/src/output/output_surface.h#L7

alexat:
Um, and
https://github.com/joncampbell123/dosbox-x/blob/develop/src/output/output_surface_sdl.cpp
https://github.com/joncampbell123/dosbox-x/blob/develop/src/output/output_surface_sdl2.cpp

aybe:
Yes in fact they're on HDD,

but:

2018-06-30 20_15_30-window

Seems like you're not using VS since they're not on vcxproj:

https://github.com/joncampbell123/dosbox-x/commits/master/vs2015/dosbox-x.vcxproj

Is this correct ?

aybe:
And to be precise, vs2015.

alexat:
Oh. Seems like VS2017 uses only filters and does not modify vcxproj then... I don't have them in .vcxproj indeed but they do display and compile )

image

alexat:
@aybe 
VS2017 Community, using vs2015 project files

aybe:
Alright, at least I'll know where to look at in case it fails in the future 😃 

Lettings you guys finish the refactoring, tbh I'm left in the dust !

alexat:
@aybe 
To be honest, I don't think it would really be 'finished' anytime soon ) 
I'll try to logically split it and reduce code duplication where I can, but converting it to classes is a whole other story.

PR #825, opened 2018-07-02
VMware like lockable keys

aybe:
# VMware like lockable keys

**## Do not merge, it's not finished !!!**

 - at start, fetching user settings for num/caps/scroll and enabling them
 - at stop, restore locked keys state in host

Then (not working well)
- when got focus, restore internal state
- when lost focus, restore host state
(just like VMware)

@joncampbell123 
The problem is that there is noise from key events in the SDL pump loop as well as changes made by IRQ1 handler, I couldn't figure out on how to address these

Removed old, useless sync to host behavior, now DOSBox has its own state and can be set in config.

**Additional information**

The KEYCHECK program below is useful:
http://www.ers.kestar.com.au/dos.html

Will probably fail to build in Linux since it's not finished.


joncampbell123:
Please merge in my recent changes concerning INT 10h and VESA BIOS extensions, in case your changes might overlap.

aybe:
I have your changes but honestly this is driving me crazy, there are so many factors to check out and now it's so messy that I'm considering scratching this up, I shouldn't have committed things in this order because I'm screwed.

IRQ1 is messing with BIOS initialization, obvious since both are called in the normal loop ... except when I debug but wait a few seconds ... these non-lit leds are now lit as I expect them.

I'll prolly start again from scratch but not until I know the way on how to do that, so, 
- DOSBox has its internal state of these keys
- appropriate keys are smashed whenever focus changes to reflect the active environment
- how to deal with these interactions with handlers ?

Sorry if what I'm saying makes little sense 😃

aybe:
I hate these moments when I spent quite some time and something still doesn't work 🎯

joncampbell123:
One thing I've learned to do when working on a messy codebase like this is that when adding new features or making a major change, do it *slowly*. Commit by commit, incremental changes. Try that.

Doing a whole lot at once in a commit is a bad idea.

aybe:
You're definitely right, tried again on #826

PR #826, opened 2018-07-03
VMware like lockable keys Take 2

aybe:

- options in config to set wanted state for num, caps and scroll
- set user scheme at start
- restore host scheme at exit
- **restore dynamically when got/lost focus (not yet)**

Removed the old synchronization where DOSBOX would follow the host.

**Additional information**

The KEYCHECK program below is useful:
http://www.ers.kestar.com.au/dos.html

Should build in Linux now, (hopefully)


aybe:
@joncampbell123 

I fixed the startup problem but now I want to address the synchronization when window is active or not.

Now what I'd need you to explain me, is how can I play nicely with IRQ1 when

- on got/lost window focus, I press according keys to restore host/internal state
- to prevent these keyboard handlers to register these changes
- I've tried a boolean then raising IRQ1 but it failed and it's a poor man job
- any ideas on this ?

Alternatively, give a quick picture of all that thing works because honestly it's so messy that I can't figure anything !

Thanks

aybe:
Basically I diffed what I did then staged thing by thing, unfortunately the 'files changed' hurts your eyes.

To get a better view of what changed, see the commits instead, it's much easier to understand.

aybe:
@joncampbell123 

I guess it's working now, test and merge if okay.

aybe:
@joncampbell123 

I have to say that I'm stumped by this, it works VERY erratically,

While it seems to works fine when you run it, in fact it is just pure luck, the behavior is not consistent when you either debug/wait for some time.

So when you have a minute ... 👍

aybe:
i.e. the code is all in place, one just has to spot where this corruption comes from .... the more I look at it, the less I have clues !

joncampbell123:
The problem with Windows and the num/scroll/caps LEDs is that Windows provides a way to directly read their state, but the ability to toggle their state only by sending a fake VK scan code to yourself.

The problem with simply sending yourself the scan code is that you can't tell fake from real input in WM_KEYDOWN/WM_KEYUP, so you run a real risk of an infinite loop sending yourself fake codes if the user presses any of those keys.

This is why the code I wrote includes a low level keyboard hook and remapping of those keys to different messages in the way I wrote it.

If Microsoft had provided a way to directly set LEDs and keep the keyboard driver from controlling them if desired, this code would be a lot easier.

aybe:
I see, very subtle, I've been studying VMware and they are able to do that but I don't know how ...

The other things that this brought in were AttachThreadInput, BlockInput but this in turns brings first the need to understand how dosbox really works, then sdl too. (remember reading somewhere that some stuff on it 'can' happen on a different thread).

What was bugging me with my code was this:
https://github.com/joncampbell123/dosbox-x/pull/826/commits/ac63b8035e07f785cb230377353fb9d21c5f8fc1#diff-3b1f3c818385e56283d07b63d93a2ee7R5524
But in fact it did work !! However when you step through at different pace, results differ ... until I concluded there's definitely something wrong happen outside. 

Anyway what turned out to be easy isn't so, let's just assume it's paused for now unless one attempts stronger measures.

aybe:
- using the value of an external key state to manage an internal key, counter productive but ironically, it worked !
- stepping: depending where you put breakpoints in the different parts involved in all of dosbox, would produce different results (invert result)
- outside: input is received from outside, something wrong's happening, sometimes
- measures: it can't be as simple as that apparently, maybe  AttachThreadInput and BlockInput are the solution, but to me it starts to like being a little too much for a feature supposedly simple
- paused: spent too long on it and failed, letting this thing sleep, meanwhile, someone might have an idea on how to fix it

hope that makes sense !

Allofich:
Thank you! Yes, I think that clarifies the situation (just seemed a little hard to understand to me).

joncampbell123:
Is this pull request ready to merge or do you need to revise it further?

aybe:
No it is not in fact as I mentioned before, closing it for now :(

Issue #822, opened 2018-07-01
Minor warnings from VS 2017 in 1 July code


emendelson:
Probably you know this, but these are the warnings generated by current code from VS2017 running under Windows 10 64:

9>..\src\dos\dos_programs.cpp(756): warning C4018: '>': signed/unsigned mismatch

9>..\src\gui\sdlmain.cpp(5223): warning C4244: '=': conversion from 'const long' to 'Sint16', possible loss of data
9>..\src\gui\sdlmain.cpp(5224): warning C4244: '=': conversion from 'const long' to 'Sint16', possible loss of data

Only three, and they clearly don't really matter, and if this is a needless post, my apologies for wasting bandwidth!

joncampbell123:
Eh, they're not all that important, but worth taking care of at some point.

Allofich:
All warnings for DOSBox-X code are currently taken care of in MSVC builds (VS 2019, not VS 2017, but they raise the same kind of warnings).

Various warnings are still raised for library code used in the compilation. But these and any new warnings are visible at any time, and the warnings above were probably taken care of long ago. This issue can be closed IMO.

PR #831, opened 2018-07-05
Generalize aspect correction (fit and extend) functions, removing duplicated code from outputs, minor warnings fix in sdlmain.cpp

alexat:
Two changes:
- Generalize aspect correction (fit and extend) functions, removing duplicated code from outputs
- Seems like updated config.h did not get to the latest patchset, effectively disabling Direct3D output, sorry
- Also minor fix to the mouse movement event code to convert long to int16 with saturation (eliminating compiler warnings) (part of #822)

Also updated up to master.


alexat:
Also should be updated to master. 
I am also preparing the next small patchset that will eliminate a bit of aspect correction code duplication, will do it against master. Thanks.

alexat:
Done moving aspect correction functions out, shortening output code.

alexat:
Done moving aspect correction functions out, shortening output code.

I think output code is now readable enough to finalize this part of refactoring sdlmain outputs.
Everything else remaining can be done later on separately.

Issue #860, opened 2018-07-23
Printing


the1fitz:
Installed 32bit dos box on win10 32bit pc it runs my old dos program well, but when I click on the DOSBox-x DOS menu and click "printers" the box is empty???? What do I do to attach a printer... There are printers attached to my win10 pc. (they are network printers however)  I tried to use my program to print  to lpt1: and I get a printer error.

Also there is no readme in the windows installer, and there is no description of how to configure or make autoexec bat entries , I was hoping to launch my program in a DOSBox-x window when I open it. 

Are there any how to documents or tutorials for just running old dos programs with DOSBox-x?

Thanks in advance 
the1fitz

alexat:
For 'autoexec', see [autoexec] section at the end of default .conf. 
My own example (I'm just a fellow DOSBox-X user, hope you can get something out of this example):

[autoexec]
# Lines in this section will be run at startup.
# You can put your MOUNT lines here.
MOUNT -nocachedir C: D:\EMULATION\DOS
C:
SET PATH=Z:\;C:\VC;C:\ULTRASND;C:\TOOLS;C:\FBC
CD \
C:\ULTRASND\ULTRAMID.EXE -16
VC

alexat:
@joncampbell123 @aybe 
I've imported back printer emulation code from DOSBox-Daum along with FreeType library into  https://github.com/alexat/dosbox-x/commits/printer branch of mine, but it's not yet ready at all to be a pull request and is quite large, given entire FreeType library is added. 

Alas, I don't seem to have much spare time recently to finish it soon enough, so mentioning it just in case you may want to take it over :)

alexat:
Well, with the latest pull request in, it is like that (identical to Daum build):

```
[printer]
printer=true
dpi=360
width=85
height=110
printoutput=png
multipage=false
docpath=./print
timeout=0

[parallel]
parallel1=printer
parallel2=disabled
parallel3=disabled
dongle=false
```

Which is quite logical I think. You direct one (and only one) parallel port to the printer emulation in [parallel] section, and dot matrix printer emulator settings are in [printer] section.

ThePillenwerfer:
Thank you for the clarification.  I'd downloaded version 0.82.8, the .exe and .conf files are dated 15 July so obviously work has been done since.

DOSBox-X certainly seems far better than plain DOSBox and it'll be utterly fantastic if, and when, it can print.

landloafer:
What model of printer that is emulated?

[...]

PR #861, opened 2018-07-31
Pull in printer support back from Daum branch sources

alexat:
- Works, code is quite readable although I did not dig in deep
- Compiles in all Debug/Release/SDL 2 modes under windows
- Includes FreeType 2.9.1, static build, as a Windows build dependency
- C_PRINTER option in config.h
- Linux build needs to be adjusted to make it compile optionally and add FreeType 2
- Please absolutely test compilation before including, I'm not sure if my VC project changes are suitable


alexat:
Fixes #860

joncampbell123:
Doesn't break Linux compilation, so far, which is good.

Issue #858, opened 2018-07-20
Raspberry Pi Target


yksoft1:
**Is your feature request related to a problem? Please describe.**
Currently, Raspberry Pi series don't have acceptable X acceleration. Dosbox-X's Linux target is designed for X, currently it was too slow for Raspbian under Raspberry Pi 1 and 2, especially under PC-98 mode.

**Describe the solution you'd like**
Raspberry Pi series do have OpenGL ES2 hardware acceleration and SDL2 seems already includes a backend for that. I wonder if we could port it to our SDL1 fork, or make SDL2 code compatible with it.

**Describe alternatives you've considered**
Dosbox SVN recently had an armv7l dynrec core merged. If we could merge it into Dosbox-X, maybe we can make Dosbox-X perform better on Raspberry Pi.

brunocastello:
What actually are the steps and requirements to compile DOSBox-X on my raspberryPi3/raspbian? Just running ./build-debug threw me into an error after a few minutes.

joncampbell123:
Try ./build-debug-no-avcodec, you probably don't have FFMPEG development headers on the Pi.

brunocastello:
> Try ./build-debug-no-avcodec, you probably don't have FFMPEG development headers on the Pi.

Oh, about that...

<img width="675" alt="Screen Shot 2019-07-26 at 12 26 46 AM" src="https://user-images.githubusercontent.com/23453202/61923931-1c58f000-af3c-11e9-80be-a653674aced0.png">

brunocastello:
Fixed that, installed automake and lib tool. But another error now, this time about the doslib:

<img width="683" alt="Screen Shot 2019-07-26 at 1 18 36 AM" src="https://user-images.githubusercontent.com/23453202/61925594-6396af00-af43-11e9-800a-ce8f21014be3.png">

joncampbell123:
apt-get install ncurses-devel (I think), you need ncurses headers.

joncampbell123:
Most distros install libraries but not headers.

brunocastello:
I installed, its libncurses5-dev btw. I tried again and a new error...

<img width="1034" alt="Screen Shot 2019-07-27 at 12 31 15 AM" src="https://user-images.githubusercontent.com/23453202/61989454-fb5cd180-b005-11e9-9525-c5759c98a0a3.png">

joncampbell123:
[...]

Issue #600, opened 2018-03-10
Overall terrible performance on RPi 3


HelloOO7:
So I compiled this recently for my Raspberry Pi 3 to try running Half-Life and even with Voodoo fully set up it gets a maximum of about 0,7 FPS in the beginning. Windows 95 itself is pretty choppy, but usable, except for the disk speed (HL1 installation - 420 megabytes took 2 hours) and the screen refreshing - when i open up a window or boot up Windows, the picture shows up slowly "rendering" from the top. Oddly enough, the pre-compiled DOSBox SVN build for the Pi runs just fine, except for no Voodoo support > no games. My config file should be in the attachment. Am I doing something wrong or is the Pi 3 just uncapable of running advanced versions of DOSBox?
[dosbox.zip](https://github.com/joncampbell123/dosbox-x/files/1799766/dosbox.zip)
Thanks

joncampbell123:
The PI is too slow for DOSBox-X to run those games.

My guess is that DOSBox SVN has a dynamic core that does JIT emulation to compile segments of x86 code into ARM assembly. DOSBox-X has no such core.

HelloOO7:
Alright, thanks, looks like I'm off to compiling Dosbox Daum for the 5th time now. 

Just a quick question, do you think JIT could be implemented in Dosbox-X somewhen later on? I really like it, not only because of the wide range of features but also because it was the only Dosbox with Windows games potentially playable that didn't fail to compile on my Pi.

joncampbell123:
I suppose I could see if it's possible to port the ARM dynamic core to DOSBox-X. I have a Raspberry Pi 3 I could try it on, at least.

HelloOO7:
If you'd ever take your time to do this, I'll gladly opt in to test on my 3 too. Just gimme a buzz if you want! And again, thanks very much.

Daniel-Trevitz:
> If you'd ever take your time to do this, I'll gladly opt in to test on my 3 too. Just gimme a buzz if you want! And again, thanks very much.

buzzzz

vivian-ng:
Are there any instructions or scripts to cross compile dosbox-x for a Raspberry Pi or similar ARM target? When I try to compile natively on the RPi3... it took, let's just say, quite a long bit of time.

joncampbell123:
I wouldn't expect anything to compile quickly on the Pi to be honest.

I once compiled a Linux kernel on a VIA C3 processor and it took a whole day.

[...]

PR #948, opened 2018-11-29
Dynamic Recompilation for ARM

Daniel-Trevitz:
# Description

Add the dynamic recompilation core from dosbox svn trunk
Do some basic work to add the GUI menu cpu selection.

#600, #858 

Add the dynamic recompilation core from dosbox svn trunk

**Additional information**

Only minimally tested. The game I wanted to work on the RPi3 now does, and it seems to be running too a little too quickly. Tested using Wacky Wheels on RPi3 with SDL2 without X (from ssh onto fb)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -425,6 +425,24 @@ case "$host_cpu" in
     c_targetcpu="m68k"
     c_unalignedmemory=yes
     ;;
+   armv7l)
+    AC_DEFINE(C_TARGETCPU,ARMV7LE)
+    AC_MSG_RESULT(ARMv7 Little Endian)
+    c_targetcpu="arm"
+    c_unalignedmemory=yes
+    ;;
+   armv6l)
+    AC_DEFINE(C_TARGETCPU,ARMV4LE)
+    AC_MSG_RESULT(ARMv6 Little Endian)
+    c_targetcpu="arm"
+    dnl c_unalignedmemory=yes
+    ;;
+   aarch64)
+    AC_DEFINE(C_TARGETCPU,ARMV8LE)
+    AC_MSG_RESULT(ARMv8 Little Endian 64-bit)
+    c_targetcpu="arm"
+    c_unalignedmemory=yes
+    ;;
    *)
     AC_DEFINE(C_TARGETCPU,UNKNOWN)
     AC_MSG_RESULT(unknown)
@@ -640,6 +658,25 @@ AC_ARG_ENABLE(core-inline,AC_HELP_STRING([--enable-core-inline],[Enable inlined 
   fi
 ],)
 
+dnl automake 1.14 and upwards rewrite the host to have always 64 bit unless i386 as host is passed
+dnl this can make building a 32 bit executable a bit tricky, as dosbox relies on the host to select the
+dnl dynamic/dynrec core
+AC_MSG_CHECKING([whether Apple user wants to override the build process to produce a 32 bit binary])
+case "$host" in 
+  *-*-darwin*)
+   if test x$c_targetcpu = xx86_64 -a x$c_sizep = x4 ; then
+     AC_MSG_RESULT(yes)
+     AC_DEFINE(C_TARGETCPU,X86)
+     c_targetcpu="x86"
+   else
+     AC_MSG_RESULT(no)
+   fi
+   ;;
+  *)
+   AC_MSG_RESULT([no, not on Apple])
+   ;;
+esac
+
 dnl FEATURE: Whether to enable dynamic core
 AC_ARG_ENABLE(dynamic-core,AC_HELP_STRING([--disable-dynamic-core],[Disable all dynamic cores]),,enable_dynamic_core=yes)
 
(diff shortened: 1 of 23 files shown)


joncampbell123:
Nice! Before I accept the pull request, does the core run anything else in a satisfactory manner on your end?

When I have time I'll pull out my Raspberry Pi 3 and see if the ARM version of this core works.

Daniel-Trevitz:
Just ran Commander Keen 1 for a minute and it seemed fine. Ran "Sky roads" as well without any issues.

joncampbell123:
What about 32-bit protected mode games, like DOOM?

Daniel-Trevitz:
Legality aside, I jsut now grabbed doom95.zip from here: https://www.dosgamesarchive.com/download/doom/
doom95.exe complains that it is a Windows NT executable.

joncampbell123:
No, I mean the original DOOM, for MS-DOS.

I don't expect Windows 3.1 and Windows 95 to run against this core just yet :)

sikthehedgehog:
Because, um, that's a Windows port… You need the DOS version (even the shareware version will do).

That said, Wacky Wheels isn't a 32-bit game? Huh (but then again I never looked into it really)

joncampbell123:
I'm only asking to confirm that basic 16-bit real mode and 32-bit flat protected mode stuff works.

Daniel-Trevitz:
Just ran doom19s.exe (Misunderstood what the 95 meant...) works great.

joncampbell123:
Sound good to me then.

joncampbell123:
Unfortunately for me, the dynrec core just causes DOSBox-X to hang on x86_64 Linux.

joncampbell123:
What configuration did you test this on?

Daniel-Trevitz:
I know extremely little about autoconf - but it looks like C_DYNAMIC_X86 is not the default like it should be for x86.

joncampbell123:
Correct. My primary Linux dev system is x86_64 and C_DYNAMIC_X86 was written for 32-bit targets. It still works well when compiled on 32-bit Windows though.

joncampbell123:
Just to check you're compiling to 32-bit ARM on your Pi 3, right?

It looks like the problem is with the way dynrec handles CPU_Cycles. It writes to the memory address from the generated code. Problem is, it assumes the variable is 32 bits wide. On my x86_64 Linux system, it's 64-bit wide. So when CPU_Cycles goes negative it appears to DOSBox-X instead as a very large integer and the tick never completes.

joncampbell123:
I just made some commits that add "dynamic" to the menu for dynrec, and a hack to the dynrec core that works around the incomplete write to CPU_Cycles, which allows dynrec to (so far) work on my system.

joncampbell123:
Awesome, I can run DOSBox-X with cycles=120000 and Blood runs full framerate.

joncampbell123:
I just got the Windows builds with VS2017 to use the dynrec core when targeting 64-bit Windows, and it works.

joncampbell123:
Anyway, thanks again for porting over the dynrec core, it works very well so far!

yksoft1:
Tried and worked on my MSYS2 mingw-w64 toolchain.

@joncampbell123  If the dynrec from DOSBox SVN is enough fast, would we replace our dynamic_x86 for that even if we are on x86-32?

fr500:
I thought the dynarecs were removed for accuracy.
dynrec x86 is not as fast as dynamic_x86

joncampbell123:
It was removed because it has problems with the way page faults are handled, which can cause problems with 32-bit protected mode games and OSes like Windows 95.

But then enough people complained that I added it back, though with the restriction that by default it defer to the normal core whenever 386 protected mode paging is enabled.

fr500:
Ah! good to know!
The new dynarec seems promising, they even added an AARCH64 emitter recently so it seems quite active, and two weeks ago the x86_64 fixes came.

joncampbell123:
The original way DOSBox SVN handled page faults is to push a page fault frame onto the stack and recurse into another execution loop to handle the page fault. When the page fault returns, it exits the sub-loop and continues where it left off.

That's perfectly fine for DOS games and Windows 3.1

The problem is Windows 95, NT, and Linux, where a page fault is handled in user-space and can be interrupted by the kernel when it decides it's time to switch tasks (to resume handling the page fault when the task is revived). Windows 95 doesn't always return to the task that faulted in the same order it left it, and if another task page faults, then DOSBox will recurse again to handle it.

For example, say Windows 95 is running task A, which page faults (to pull in swap) but the kernel decides instead to switch to task B. Task B during it's run page faults (to pull in swap) but Windows 95 decides to switch to task C. Task C is given it's time, then Windows 95 decides to switch back to task A, which then completes it's page fault handler and returns, causing DOSBox to exit the loop and resume whatever it was doing. Problem is, DOSBox is resuming whatever it was doing for task A in task B. That isn't enough to cause problems usually, though eventually it does cause something to get missed and Windows 95 starts to run a little funny and eventually crash. However this can happen in a way that DOSBox never returns from the sub-loop it spawns, and continues running a sub-loop on another page fault. Left to run long enough this can fill up the stack and crash DOSBox.

In the original DOSBox SVN project the CPU cores were written in such a way that if an instruction was interrupted by a page fault, and Windows 95 task switches during a page fault such that they are completed out of order, the CPU state becomes inconsistent and things in Windows 95 start crashing. One of the first things I did in DOSBox-X around 2013 was fix up the normal core so that CPU state is consistent: it either completes and continues, or a page fault happens and the CPU state is rolled back to what it was when the instruction started. Doing that made Windows 95 a lot more stable in DOSBox-X using the normal core!

joncampbell123:
I made substantial modifications to the normal core to throw a guest page fault as a C++ exception so that the CPU core is interrupted immediately and given to the Normal loop as a page fault to happen. This is the "non recursive page fault" handling I implemented.

It even happens to work with the dynamic core... sort of, at least on Windows. On Linux unfortunately the dynamic core doesn't mesh well with the GCC exception handling frame data, so if a C++ exception is thrown while within dynamic core the C++ runtime acts as if nobody were there to catch it (because it can't trace the stack) and DOSBox-X crashes with an uncaught C++ exception. This is why the code was brought back with the restriction to defer to normal core if 386 paging is enabled. Without 386 paging, no C++ exceptions are thrown and it's safe to use.

fr500:
Yeah I have noticed windows works a lot better on X.
On upstream I get "application stopped working" all the time!

Regarding DYNREC's ARM performance, it's decent, not groundbreaking, games like GTA still struggle on Tegra X1 for instance.

That said it's a lot better than with the normal core. Example (not trying to plug my channel, just excited to see discussion, I have been working on dosbox libretro a lot):
https://www.youtube.com/watch?v=VM8WzYw0tAA

I may try to tackle DOSBox-X libretroization again at some point.

My dosbox fork is not too deep, I try to keep the upstream code untouched as much as I can, but I stumbled in some interfaces that had changed and didn't have enough time to ask last time I tried I don't know much about emulators, I just know about libretro :)

PR #955, opened 2018-12-08
Make MT-32 header mmath.h compatible with non-SSE2 math.

yksoft1:
I found that mt32emu from upstream does not require SSE2 to build at all. After some investigation I found that ykhwong (author of DOSBox Daum) modified mmath.h of mt32emu to use a "fast math library for float" which requires SSE2.
I just changed that to use generic C floating point math when SSE2 is unavailable, and removed SSE2 requirement from configure.ac.
Don't know if that would work in non-x86/x64 platforms.


joncampbell123:
Can you first confirm that this does not break MT32 emulation on your end?

yksoft1:
> Can you first confirm that this does not break MT32 emulation on your end?

On my end I could confirm that MT-32 emulation is still OK after this.

PR #1016, opened 2019-02-20
Windows Desktop ARM64 port

driver1998:
# Description
Windows Desktop ARM64 port

No need for the spanking new Windows ARM laptops to run the x86 builds via emulation.

Both SDL and SDL2 version compiles and tested with VS2017 15.9 and Win10 SDK v17134
ARMv8 dynarec core working

Disabled OpenGL output for ARM64, since opengl32.lib is not provided in the SDK
Might work with third party OpenGL implementation though

**Additional information**
My test device: Raspberry Pi 3B running Windows 10 ARM64 build 17134
Would be nice if somebody can test this port on officially supported platform (Like Snapdragon 835 and 850 based laptops).


joncampbell123:
Nice!

Issue #1024, opened 2019-03-08
Enhancements for CGA Emulation


frank-deng:
* Is it appropriate to make even scanlines dimmer for CGA modes like cga, cga_rgb, cga_mono, so as to resemble the behaviour of the old CGA/EGA monochrome or color monitors?
* Color switching for cga_mono seems to have missing under SDL1.
* For cga_mono or some other monochrome displays like mda, hercules, is it possible to specify the color in the configuration file (green/amber/white) apart from switching it using keyboard?
* Is it appropriate to add grayscale effect for other displays like EGA, VGA, or even newer cards like svga_s3? Maybe VGA-supported monochrome/grayscale monitors could be used for budget-saving purpose in some cases.

joncampbell123:
There's already a scaler that renders to every other scanline, if you'd like that.

However a scaler that acts similar but with a dimmer second line would be more appropriate, I agree.

I'm not sure if there's a dosbox.conf option for it, but there is monochrome display color selection tied to a mapper shortcut.

You want a 'machine=vga_mono?' I don't really see the point for S3 emulation to do it. I don't think monochrome VGA matter beyond the original IBM PS/2 line and it's VGA implementation.

This would affect VGA rendering and the INT 10h emulation (concerning the BIOS call to read back DCC monitor/card combination).

joncampbell123:
If I recall, the monochrome VGA monitor uses only one of the R, G, or B pins and relies on the BIOS to do the weighted RGB -> grayscale conversion.

I'm guessing its the G (green) pin, can anyone with a mono VGA monitor confirm?

joncampbell123:
I do know from actual hardware experience the Tseng ET3000 is a bit overzealous about monitor detection to the point that it will mistake most LCD monitors as a monochrome VGA monitor.

frank-deng:
I've tried scan2x and tv2x scalars, and they only work for 320x200 graphics mode. CGA's 640x200 graphics mode and text mode not work.

joncampbell123:
What happens if you force the scaler?

frank-deng:
I've searched the code of `src/hardware/vga_draw.cpp`, and I haven't found how CGA text mode and 640x200 graphics mode rendered yet.
What should I do to render even lines dimmer or skip rendering even scan lines?

frank-deng:
[...]

PR #1041, opened 2019-04-03
Try to apply scan2x and tv2x filter to 640x200 mode

frank-deng:
# Try to apply scan2x and tv2x filter to 640x200 mode

```#1024```


joncampbell123:
I'm fixing up the VS2015 project files right now...

Before I merge this can you describe exactly what this changes and how?

PR #1045, opened 2019-04-05
Add scaler for grayscale effect

frank-deng:
Add new scaler called `gray2x` to create a grayscale effect.

This scaler simply does the weighted RGB->Grayscale conversion, with the VGA hardware level unaltered, may be useful to emulate the look of a monochrome VGA monitor.

Files modified:
`src/dosbox.cpp`
`src/gui/render.cpp`
`src/gui/render_scalers.cpp`
`src/gui/render_scalers.h`
`src/gui/render_templates.h`


joncampbell123:
I'm going to rename it from gray2x to gray. It doesn't upscale by 2x like normal2x does.

Issue #975, opened 2018-12-28
Back-porting DOSBox-X features/fixes to DOSBox SVN


joncampbell123:
**Describe the solution you'd like**
Small incremental patches to DOSBox SVN to incorporate DOSBox-X bugfixes and enhancements (but only those related to DOS gaming).

joncampbell123:
Ideas:
- Support for PC speaker frequency rapid changes/modulation (Apogee/Duke Nukem)
- Goldplay mode
- Option to render scanline by scanline (instead of requiring machine=vgaonly)
- Addition of machine=mda as cut-down emulation of machine=hercules
- MIXER_FillUp() calls to changes in state to OPL2/OPL3 and Tandy 3-voice emulation. There are a few games that play "digitized speech" on Adlib cards by rapid changes to an FM channel (some old golf game), and there are a few games that play a digitized PCM channel on 3-voice (Skate or Die).

fr500:
first order of business would be asking them to use git :P
DOSBox could benefit hugely from moving to git, there is so much fragmentation.

I have a github mirror I keep up-to-date here, git svn is a pain but it works at least.
https://github.com/fr500/dosbox-svn

joncampbell123:
Okay...

Most of these items, excluding the machine=mda and scanline by scanline options, are impossible with the mixer as-is in SVN at the moment.

The mixer in SVN will need to be updated to support rendering up to the current emulator "time" at any moment, so the PC speaker, goldplay mode, OPL3 and Tandy code can call channel->FillUp() prior to changing state to support changes to state that occur faster than the 1ms interval the mixer is called to render audio.

joncampbell123:
When I submit a patch to do so, I'm going to go for minimal changes to the mixer, without all the lowpass filter and slew rate logic I put into DOSBox-X since those parts are probably not interesting to the DOSBox devs. Those features listed won't be possible until calls to FillUp() can trigger an accurate rendering of audio up to exactly that emulation time within the current tick interval.

joncampbell123:
> first order of business would be asking them to use git :P
> DOSBox could benefit hugely from moving to git, there is so much fragmentation.
> 
> I have a github mirror I keep up-to-date here, git svn is a pain but it works at least.
> https://github.com/fr500/dosbox-svn

That would be nice, however there is the risk of losing the last 18 years worth of SVN commit comments.

joncampbell123:
Requested by others on the Vogons forums:
[...]

PR #1054, opened 2019-04-27
Start of integration of DOSBox SVN where DOSBox-X left off

Allofich:
# Description

_Summary of changes brought by this PR._

Integrates the next 3 DOSBox SVN commits starting from where they seem to be missing from DOSBox-X.

May be slightly helpful for https://github.com/joncampbell123/dosbox-x/issues/975, as a more common base would probably make any feature integration go more smoothly.

**Additional information**

I was looking at the DOSBox-X code, and it looks like it started on GitHub based on an older SVN base:
https://github.com/joncampbell123/dosbox-x/commit/f3318240311646e6aa14663a9e3e2c41d0da2a68

Then you integrated the daum branch, which included newer commits from mainline:
https://github.com/joncampbell123/dosbox-x/commit/21614b8c3f2e945fb46b0f1bbd3d9f12f7847355

I think this gave you as far as commit r3873 from mainline.
https://sourceforge.net/p/dosbox/code-0/3873/

You have later commits that selectively add in other mainline commits, like this one:
https://github.com/joncampbell123/dosbox-x/commit/85de2c9372bd9201964446e10d93266bc54672c9

I would like to try to help fill in the gaps so that the improvements from mainline DOSBox are also in DOSBox-X.

This PR includes the next 3 commits after r3873, authored by qbix79.

https://sourceforge.net/p/dosbox/code-0/3834/
https://sourceforge.net/p/dosbox/code-0/3835/
https://sourceforge.net/p/dosbox/code-0/3836/

There was no conflict with DOSBox-X work, and DOSBox-X successfully compiled and ran. I'm doing the compilation through AppVeyor compiling with Visual Studio 2017 and running the result on Windows 10, so I'm only able to check those environments.

If you're okay with this, I'm thinking of maybe doing these in sets of 3 commits like this, skipping over those that are already in DOSBox-X, and prioritizing DOSBox-X if there are conflicts.

### DIFF
--- a/NOTES/Skipped SVN commits.txt
+++ b/NOTES/Skipped SVN commits.txt
@@ -1,0 +1,4 @@
+Commit#:	Reason for skipping:
+
+3835		Preparatory commit for 3836, which was skipped.
+3836		Assumes the Windows API. midiOutGetNumDevs() and midiOutGetDevCaps() are specific to Windows and do not exist anywhere else.
--- a/src/debug/debug.cpp
+++ b/src/debug/debug.cpp
@@ -513,7 +513,7 @@ void CBreakpoint::Activate(bool _active)
 // Statics
 std::list<CBreakpoint*> CBreakpoint::BPoints;
 CBreakpoint*			CBreakpoint::ignoreOnce = 0;
-Bitu					ignoreAddressOnce = 0;
+PhysPt					ignoreAddressOnce = 0;
 
 CBreakpoint* CBreakpoint::AddBreakpoint(Bit16u seg, Bit32u off, bool once)
 {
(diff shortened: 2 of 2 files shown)


joncampbell123:
I appreciate the work!

However I cannot accept commit 83631d7661e88887b854b5133afec78f409ab6eb because it assumes the Windows API. I compile DOSBox-X on Linux and Mac OS X as well, and there are others that compile it for AmigaOS, BeOS, and HaikuOS.

joncampbell123:
To clarify, midiOutGetNumDevs() and midiOutGetDevCaps() are specific to Windows and do not exist anywhere else.

Allofich:
OK, actually going in and modifying the contents is a bit out of my depth. Shall I drop the two midi commits, or would you like to modify them yourself as needed? If we drop, I suggest keeping a list of skipped-over commits from SVN, which I could add to this PR.

joncampbell123:
@Allofich Go ahead and drop the MIDI commits and note that they were dropped. I'm OK with the first commit that corrects typos in the source code.

Allofich:
OK, I did so. I dropped 3835 because it was a preparatory commit for 3836, but I may add it again later if I find another SVN commit that wants to use `midi.h`, which was added by 3835 added.

PR #1055, opened 2019-04-27
Integrate next 4 SVN commits

Allofich:
Successfully built and ran.


Allofich:
I replaced the tabs of SVN r3845 to be spaces, to match the surrounding DOSBox-X code.

Allofich:
By the way, I can confirm that, as the r3845 commit says, Lemmings 3D Sound Blaster sound is working now that this has been merged. It was silent before.

PR #1060, opened 2019-04-28
Next 6 SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/3854/ - In this PR
https://sourceforge.net/p/dosbox/code-0/3855/ - Skipped. Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/3856/ - Skipped. Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/3857/ - Skipped. Conflicts with work done in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/3858/ - Skipped. Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/3859/ - Skipped. The function modified in the commit doesn't exist anymore in DOSBox-X, but the behavior exists in what I assume is the equivalent function in DOSBox-X. You might want to check to be sure.
https://sourceforge.net/p/dosbox/code-0/3860/ - The part of the commit that was missing from DOSBox-X is in this PR. The other part conflicts with DOSBox-X work and might be unneeded.
https://sourceforge.net/p/dosbox/code-0/3861/ - The part of the commit that was missing from DOSBox-X is in this PR, but I think you may not want it, as you commented out the part of this commit that did exist in DOSBox-X as it was apparently causing problems.

Can you take a look at 3859, 3860, and 3861 when you have time? I got your message about PRs not being accepted for a little while until release is over, that's fine.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.82.19 (next)
+  - Integrated commits from mainline: 3854, 3860, 3861, 3862,
+    3864, 3866 (Allofich)
   - Remove old dynamic x86 core. Dynamic core is exclusively
     dynrec now.
 0.82.18
--- a/NOTES/Skipped SVN commits.txt
+++ b/NOTES/Skipped SVN commits.txt
@@ -6,3 +6,7 @@ Commit#:	Reason for skipping:
 3838		Related to skipped commit 3836.
 3846		Large commit
 3847		Conflicts with https://github.com/joncampbell123/dosbox-x/commit/ac70b5599fb2fb824be7ed627ec4e70aa8b6bc79
+3857		Conflicts with DOSBox-X
+3859		Conflicts with DOSBox-X. Masking/wrapping is implemented in DOSBox-X by masking the array access per pixel.
+3863		Conflicts with DOSBox-X
+3865		Conflicts with DOSBox-X
--- a/include/fpu.h
+++ b/include/fpu.h
@@ -168,7 +168,7 @@ typedef struct {
 	FPU_Tag		tags[9];
 	Bit16u		cw,cw_mask_all;
 	Bit16u		sw;
-	Bitu		top;
+	Bit32u		top;
 	FPU_Round	round;
 } FPU_rec;
 
--- a/src/fpu/fpu_instructions.h
+++ b/src/fpu/fpu_instructions.h
@@ -44,6 +44,13 @@ static void FPU_FNOP(void){
 	return;
 }
 
+static void FPU_PREP_PUSH(void){
+	TOP = (TOP - 1) &7;
+//	if (GCC_UNLIKELY(fpu.tags[TOP] != TAG_Empty)) E_Exit("FPU stack overflow");
+	fpu.tags[TOP] = TAG_Valid;
+	fpu.use80[TOP] = false; // the value given is already 64-bit precision, it's useless to emulate 80-bit precision
+}
+
 static void FPU_PUSH(double in){
 	FPU_PREP_PUSH();
 	fpu.regs[TOP].d = in;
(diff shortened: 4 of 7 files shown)


joncampbell123:
https://sourceforge.net/p/dosbox/code-0/3859/

That function does not exist anymore because DOSBox-X changed VGA emulation to render 32bpp RGBX instead of 16bpp RGB.

Masking/wrapping is already implemented by masking the array access per pixel. It may be slightly slower than DOSBox SVN's implementation but it is also simple and straightforward and easier to maintain.

joncampbell123:
https://sourceforge.net/p/dosbox/code-0/3860/

This could be manually implemented, I see what it's trying to do and it's fairly straightforward.

joncampbell123:
https://sourceforge.net/p/dosbox/code-0/3861/

The part that changes direct TOP manipulation to FPU_PREP_PUSH() could be applied, the other changes that E_Exit() on FPU overflow/underflow will not be applied.

I explicitly removed those FPU stack checks because bombing out of an emulator like that is completely inappropriate for an FPU overflow or underflow. Real hardware doesn't seem to mind overflow and underflow. Those checks will also trigger for many perfectly fine DOS games and demoscene prods, and they DOSBox SVN from running Windows 95 with MSIE 4 or with the "Active Desktop" mode because those seem to rely a lot on the FPU.

Allofich:
I commented out those lines in fpu_instructions.h the same manner that they are commented out in fpu_instructions_x86.h.

If you would prefer that I remove them entirely I can do so, but it seems like then the others should also be removed.

joncampbell123:
Commenting them out is fine. As long as they don't do anything in DOSBox-X.

Those checks are the reason DOSBox-X is able to run things DOSBox SVN cannot. Notice in the testing matrix I've maintained how many demos are marked FAIL with "FPU stack underflow" errors.

http://dosbox-x.com/demoscene-compat.html

Allofich:
I'm totally fine with your decision! I've looked at that page before, I like seeing how much accuracy DOSBox-X has accomplished, and it was the main reason for me to start using it (also the frequent updates). I looked at the page now, and you're right, there are a lot of FPU stack overflows/underflows with SVN.

Allofich:
Hold, on a subsequent commit is necessary for this to build correctly.

Allofich:
https://sourceforge.net/p/dosbox/code-0/3862/ - In this PR
https://sourceforge.net/p/dosbox/code-0/3863/ - Skipped. Conflicts with DOSBox-X (sdlmain.cpp seems to be completely different and I didn't see anywhere to apply SVN's changes)
https://sourceforge.net/p/dosbox/code-0/3864/ - In this PR. (This fixes the build problem by reordering code)
https://sourceforge.net/p/dosbox/code-0/3865/ - Skipped. Conflicts with DOSBox-X. (vga_draw.cpp also seems to be quite different in DOSBox-X, so nowhere to apply the changes)
https://sourceforge.net/p/dosbox/code-0/3866/ - In this PR

joncampbell123:
Good catch on the Hercules mode set, I didn't catch that either.

joncampbell123:
@Allofich It might be more informative for developers and end users to list in the CHANGELOG the SVN commit from mainline as well as what the SVN commit is meant to fix (from the SVN commit message).

Allofich:
> @Allofich  It might be more informative for developers and end users to list in the CHANGELOG the SVN commit from mainline as well as what the SVN commit is meant to fix (from the SVN commit message).

OK, the next time I send a PR, I'll redo the change list items. I'll think about a good way to concisely write what was changed. I'll also update the changelist entry that was listed for the previous DOSBox-X version, if that's fine with you.

joncampbell123:
Works for me.

PR #1065, opened 2019-04-29
Next 3 mainline commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/3867/ - Skipped. Conflicts with DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/3868/ - Skipped. Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/3869/ - Skipped. Could be added, but I don't know enough to judge if this is needed for DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/3870/ - Skipped. Conflicts with DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/3871/ - Skipped. Conflicts with DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/3872/ - Skipped. Already in DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/3873/ - Skipped. Conflicts with DOSBox-X.
3874 through 3877 - Skipped. Already in DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/3878/ - Skipped. Out-of-date copyright year change.
3879 through 3888 - Skipped. Already in DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/3889/ - In this PR
https://sourceforge.net/p/dosbox/code-0/3890/ - Skipped. Already in DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/3891/ - Skipped. Conflicts with DOSBox-X.
3892 through 3894 - Skipped. Already in DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/3895/ - In this PR (partial)
https://sourceforge.net/p/dosbox/code-0/3896/ - In this PR (partial)

I ran into an area where a lot of commits were already integrated over to DOSBox-X, so lots were skipped. Also skipped some SDL-related commits that conflict with DOSBox-X because the files are very different now.

About https://sourceforge.net/p/dosbox/code-0/3869/:
This one could be added, but I don't know enough about it to judge. DOSBox-X has a later part in the file not present in the SVN code that also sets `-mno-ms-bitfields` so I skipped this one. Do you know if it would be relevant for DOSBox-X?

About https://sourceforge.net/p/dosbox/code-0/3895/:
I only integrated the code cleanup part. It looks like DOSBox-X does not do the same check for valid characters in the path? Also, I did not do the DOS_CompressMemory() changes because DOSBox-X has an additional

```
	if (*blocks > total)
		DOS_CompressMemory(segment-1);
```

here and I didn't know if the SVN commit would be compatible with that. Do you know?

About https://sourceforge.net/p/dosbox/code-0/3896/:
It's a bit risky because I tried to mix in the SVN commit with the existing DOSBox-X code. It compiles and runs, so there is no obvious problem to me, but this needs your experienced eye. Also, we have a test case game for this one. This change is supposed to fix the sound in Tempest 2000. If you move the cursor at the title screen, a voice will say "Excellent". In master branch DOSBox-X (and, strangely, the DOSBox 0.74-2 copy I have), the voice plays high-pitched. After this change (whether it is this hybrid of code, or exactly the SVN change, I tried both) the voice plays low-pitched. It also plays in the low pitch in the DOSBox ECE copy I have. 

I searched YouTube videos to find what it should sound like. I found one of the MS-DOS version from years ago that was high-pitched, probably played on DOSBox (https://www.youtube.com/watch?v=dEB4SOyOC0k&t=31s). The game is a port of an Atari Jaguar title, and there are videos of the Jaguar version, both on emulator and (what seems to be) the real machine (https://www.youtube.com/watch?v=ijjQpOuognI). They play the "Excellent" voice in a pitch that is lower than master branch, but higher than what this change makes it. To my ear, this change produces a closer sound. The question is, though, what did it sound like in the MS-DOS version on real Sound Blaster 16?

Because I don't know what it is supposed to sound like, I've left off the "fixes Tempest 2000" part of the fix in the changelog and commit message for now.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,14 @@
 0.82.19 (next)
-  - Integrated commits from mainline: 3854, 3860, 3861, 3862,
-    3864, 3866 (Allofich)
+  - Integrated commits from mainline (Allofich)
+    3854 - "top" is now used as 32 bit in dynrec core
+    3860 - Don't scroll at unspecified video page
+    3862 - Lower the influence of the aspect table correction trick when using high scale factors (320x200 => 2000x1200)
+    3864 - Code reordering to fix build
+    3866 - Correction to Hercules video height parameter
+    3867 - The mapper now uses the wrapper as well
+    3889 - Fix the possible/suggested values for integer properties.
+    3895 - Minor cleanup
+    3896 - Do less to update the frequency of an active SB DMA transfer
   - PEGC emulation will now print a warning if the guest
     application or OS attempts to use 256-color planar mode.
   - PC-98 PEGC 256-color linear framebuffer is not mapped by
@@ -23,8 +31,15 @@
     of VGA emulation.
   - Added debugger commands "VGA DRAW" and "VGA AC" to view
     drawing and attribute controller state in the VGA emulation.
-  - Integrated commits from mainline: 3834, 3839, 3840, 3843,
-    3845, 3849, 3850, 3853 (Allofich)
+  - Integrated commits from mainline (Allofich)
+    3834 - Fix typos
+    3839 - CD audio status now returns zero start and end times when no track is playing. Fixes "The Manhole".
+    3840 - Add "ADDLOG" debug command to manually add a message to the log.
+    3843 - "Strip off leading zeroes from the IP", a fix for the serial modem.
+    3845 - Add a small delay when raising the Sound Blaster 8-bit IRQ to emulate the slowness of the DSP, fixes Llamatron 2012 and Lemmings 3D.
+    3849 - Add ability to set debugger breakpoint on AL values.
+    3850 - The SB DMA callback now ignores previously selected, but not currently selected, DMA channels. Fixes Visual Player 2 with SB16.
+    3853 - Minor cleanup to joystick code.
   - PSG noise channel emulation (PC-98 FM board) apparently
     broke on Mac OS X due to type promotion by Clang/LLVM.
     Modified the code to behave as originally intended, to
--- a/NOTES/Skipped SVN commits.txt
+++ b/NOTES/Skipped SVN commits.txt
@@ -10,3 +10,9 @@ Commit#:	Reason for skipping:
 3859		Conflicts with DOSBox-X. Masking/wrapping is implemented in DOSBox-X by masking the array access per pixel.
 3863		Conflicts with DOSBox-X
 3865		Conflicts with DOSBox-X
+3867		Conflicts with DOSBox-X
+3869		Not sure if this is needed for DOSBox-X
+3870		Conflicts with DOSBox-X
+3871		Conflicts with DOSBox-X
+3873		Conflicts with DOSBox-X
+3891		Conflicts with DOSBox-X
--- a/src/dos/dos_files.cpp
+++ b/src/dos/dos_files.cpp
@@ -69,6 +69,7 @@ bool DOS_MakeName(char const * const name,char * const fullname,Bit8u * drive) {
 	char tempdir[DOS_PATHLENGTH];
 	char upname[DOS_PATHLENGTH];
 	Bitu r,w;
+	Bit8u c;
 	*drive = DOS_GetDefaultDrive();
 	/* First get the drive */
 	if (name_int[1]==':') {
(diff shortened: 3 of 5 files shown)


Allofich:
> It looks like DOSBox-X does not do the same check for valid characters in the path?

Going through the git history on `dos_files.cpp`, it looks like this change dates back to the commit where you integrated the daum branch https://github.com/joncampbell123/dosbox-x/commit/21614b8c3f2e945fb46b0f1bbd3d9f12f7847355.

Allofich:
[Comparison.zip](https://github.com/joncampbell123/dosbox-x/files/3128267/Comparison.zip)

Captures of the high (master branch) and low (this PR) "Excellent" voice.

joncampbell123:
Recent commits seem to fix screen issues on KQ1 in PCjr mode, yes.

joncampbell123:
https://sourceforge.net/p/dosbox/code-0/3869/

DOSBox-X already sets -mno-ms-bitfields if the host is MinGW or Cygwin, no need to merge this one. See configure.ac line 354.

joncampbell123:
https://sourceforge.net/p/dosbox/code-0/3895/

This will have to be manually added, carefully. The compress memory thing might be a good change.

I know of one demoscene production "So Be It" that does not run all the way through in DOSBox-X because a demopart eventually corrupts the MCB chain then attempts to EXEC the next part's EXE file.

joncampbell123:
I seem to remember that on SB16 there is one register for the sample rate, therefore 0x41 and 0x42 (input and output sample rate commands) both change the one value even though they're documented to change input vs output.

joncampbell123:
I just had a thought: Regarding the game with the sample rate issue, did you try sbtype=sbpro2 instead of sbtype=sb16?

I know from real hardware that Sound Blaster and Sound Blaster Pro cards take a time constant, but there is a limit (23KHz) unless you use the "high speed" DSP commands to play audio, which then lifts the limit to 45KHz.

Sound Blaster 16 cards do not impose the same constraint. Time constant values up to 45KHz are allowed whether or not you use the normal or "high speed" DSP playback commands.

It's possible the game expects to send a time constant that's too high and get 23KHz playback anyway.

DOSBox-X Sound Blaster emulation is written to emulate these constraints (unless the recent commits changed that).

Allofich:
> I just had a thought: Regarding the game with the sample rate issue, did you try sbtype=sbpro2 instead of sbtype=sb16?

I just tried.
With this PR applied (and the game's setup re-configured to use Sound Blaster Pro), sbpro2 sounds the same as sb16 does with this PR applied.
Interestingly, using the build from just before this PR was applied, sbpro2 plays as extremely slow and low-pitched to where it is unintelligible.

[Unintelligible.zip](https://github.com/joncampbell123/dosbox-x/files/3129984/Unintelligible.zip)

The fact that the two cards sound the same now gives me some confidence that it is now correct, or at least more correct.

Allofich:
> https://sourceforge.net/p/dosbox/code-0/3869/
DOSBox-X already sets -mno-ms-bitfields if the host is MinGW or Cygwin, no need to merge this one. See configure.ac line 354.

Right, but it looks like that mainline commit only set the flags if a certain test succeeded. Does DOSBox-X also apply the test somehow?

If no one has complained about problems with building DOSBox-X with MinGW, it may not be an issue anyway.

Allofich:
> https://sourceforge.net/p/dosbox/code-0/3895/

> This will have to be manually added, carefully. The compress memory thing might be a good change.

> I know of one demoscene production "So Be It" that does not run all the way through in DOSBox-X because a demopart eventually corrupts the MCB chain then attempts to EXEC the next part's EXE file.

I'll experiment with it a bit and see if "So Be It" gets fixed or not.

Allofich:
Tempest 2000 also has Gravis Ultrasound as a sound option. I tried it in the build from before this PR, and the pitch sounds exactly the same as the Sound Blaster options do after this PR. I think I'm confident enough now to add to the changelog that the sound for this game is fixed on Sound Blaster cards.

Allofich:
> I'll experiment with it a bit and see if "So Be It" gets fixed or not.

It doesn't. It crashes in the same spot with and without the change.

PR #1082, opened 2019-05-06
Handle INT 2F enhanced-mode Windows check

Allofich:
# Description
Handles INF 2F, 0x1600 call to check for enhanced-mode Windows.
Behaviorally it is the same as master, but this call will no longer generate an error in the log.

No

**Additional information**

http://www.oldlinux.org/Linux.old/docs/interrupts/int-html/rb-4495.htm

Confirmed in MS-DOS 6.22 that this call when made from MS-DOS, rather than from Windows, it will simply return with AX as 1600, the same as when it entered. This same thing happens in master branch since it just errors out as unimplemented but we don't need to treat it as an error.

### DIFF
--- a/src/dos/dos_misc.cpp
+++ b/src/dos/dos_misc.cpp
@@ -188,6 +188,9 @@ static bool DOS_MultiplexFunctions(void) {
 
 		}
 		return true;
+	case 0x1600:	/* Windows enhanced mode installation check */
+		// Leave AX as 0x1600, indicating that neither Windows 3.x enhanced mode nor Windows/386 2.x is running, nor is XMS version 1 driver installed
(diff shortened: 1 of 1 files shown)


joncampbell123:
@yksoft1 I believe you asked about the Windows installation check, this commit silences that warning.

joncampbell123:
@Allofich Will you be adding entries to the CHANGELOG to log these changes?

Allofich:
> Will you be adding entries to the CHANGELOG to log these changes?

Yes, sorry, forgot about that. I'll write that up for this PR and send it in now.

Within the next day or so I'll try and get the SVN changes in as well, before doing any more of those.

PR #1087, opened 2019-05-08
Next SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/3933/ - Skipped. Not relevant.
https://sourceforge.net/p/dosbox/code-0/3934/ - In this PR, although disabled.
https://sourceforge.net/p/dosbox/code-0/3935/ - Skipped. Not relevant.
https://sourceforge.net/p/dosbox/code-0/3936/ - In this PR
https://sourceforge.net/p/dosbox/code-0/3937/ - In this PR

3934 has no effect for DOSBox-X, as the affected code was commented out previously (apparently you didn't care for it). Therefore this just updates the disabled code to the newer version, but doesn't un-disable it.

The supposed fix, that it makes "dosbox -fullscreen A.EXE" work, already works in DOSBox-X in my testing.

For 3936, I can confirm that this commit allows Dune to run with Adlib Gold music, which doesn't work in master branch.
Dune II also has an option for Adlib Gold music, but it doesn't work with this commit. The game will freeze when you try to select the house to play as if you try to use Adlib Gold. This also happens if you try to play with Adlib Gold in DOSBox SVN.

I don't know how to test 3937, but it's a straightforward change with no conflicts.

Compiles and runs.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -48,7 +48,7 @@
     non-VGA machine types. Fixes Tandy and EGA display in
     International Hockey booter.
     - Handle "copy H*.txt file.txt" correctly
-    - Fix detection of always_inline attribute with ming 4.9.2
+    - Fix detection of always_inline attribute with MinGW 4.9.2
     - Introduce mount -pr to mount paths relative to last
     loaded configuration file.
     - Use normal teletype function for non-ANSI output so the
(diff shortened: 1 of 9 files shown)


joncampbell123:
I commented out 3934 because I was cleaning up that part of the code, I looked at that tangled mess of code, and said "nope, maybe later".

joncampbell123:
How is Adlib Gold support supposed to work if there is no code to register ports 38Ah-38Bh? for the Adlib emulation? Did that get missed?

PR #1092, opened 2019-05-09
Next SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/3938/ - In this PR
https://sourceforge.net/p/dosbox/code-0/3939/ - Skipped. A commented-out log message that seems unnecessary.
https://sourceforge.net/p/dosbox/code-0/3940/ - In this PR
https://sourceforge.net/p/dosbox/code-0/3941/ - In this PR
https://sourceforge.net/p/dosbox/code-0/3942/ - In this PR

These all involve FCBs.

For 3940, I merged the commit with your code to stop attempts to make Shift-JIS characters uppercase, and your use of `ascii_toupper` in place of `toupper`. I think it's correct but you may want to double-check it. The other commits were straightforward without conflicts.

Confirmed to compile and run.

### DIFF
--- a/include/dos_inc.h
+++ b/include/dos_inc.h
@@ -588,6 +588,7 @@ public:
 	void SetAttr(Bit8u attr);
 	void SetResultAttr(Bit8u attr);
 	bool Valid(void);
+	void ClearBlockRecsize(void);
 private:
 	bool extended;
 	PhysPt real_pt;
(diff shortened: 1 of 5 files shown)


joncampbell123:
@yksoft1 Would you like to double-check that this doesn't affect Shift-JIS file copies? I think I can fix the issue on Windows with file time/date and Japanese filenames.

PR #1095, opened 2019-05-11
More SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/3948/ - In this PR.
https://sourceforge.net/p/dosbox/code-0/3949/ - In this PR
https://sourceforge.net/p/dosbox/code-0/3950/ - In this PR.
https://sourceforge.net/p/dosbox/code-0/3951/ - Skipped. Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/3952/ - In this PR.
https://sourceforge.net/p/dosbox/code-0/3953/ - In this PR.

Confirmed to compile and run.
3948 is adjusted from the original commit for DOSBox-X. In that commit they made `ansi.nrows` and `ansi.ncols` into local variables `nrows` and `ncols`, which are updated from the BIOS on each use to the latest values.

From your comment about PC-98

```
        //      On real hardware, the BIOS does NOT manage the console at all.
        //      TTY handling is entirely handled by MS-DOS.
```

it seems it shouldn't happen for PC-98. so I conditioned the updates on `IS_PC98_ARCH` being false.

Also, DOSBox-X accesses the values from within various functions and it was easier/cleaner to continue using `ansi.nrows` and `ansi.ncols`, so I did not recreate them as local variables like the original commit did.

3953 is an update to 3952 so I combined them into one commit.

### DIFF
--- a/include/regs.h
+++ b/include/regs.h
@@ -62,7 +62,7 @@ struct Segment {
 enum SegNames { es=0,cs,ss,ds,fs,gs};
 
 struct Segments {
-	Bitu val[8];
+	Bit16u val[8];
 	PhysPt phys[8];
 	PhysPt limit[8];
 	bool expanddown[8];
--- a/include/setup.h
+++ b/include/setup.h
@@ -383,10 +383,10 @@ public:
 class Prop_multival:public Property{
 protected:
 	Section_prop* section;
-	std::string seperator;
+	std::string separator;
 	void make_default_value();
 public:
-	Prop_multival(std::string const& _propname, Changeable::Value when,std::string const& sep):Property(_propname,when), section(new Section_prop("")),seperator(sep) {
+	Prop_multival(std::string const& _propname, Changeable::Value when,std::string const& sep):Property(_propname,when), section(new Section_prop("")),separator(sep) {
 		default_value = value = "";
 	}
 	Section_prop *GetSection() { return section; }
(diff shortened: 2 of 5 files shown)


joncampbell123:
I accept the patches, but the ANSI emulation will need to read from a different memory address in PC-98 mode to update nrows and ncols. Should be easy to do, no worries.

https://github.com/joncampbell123/dosbox-x/issues/1098

Allofich:
I see you noticed (in the issue comments) that I added a check for PC-98 mode.

I just noticed now that I accidentally added some of the PC-98 checks with tabs in front of them. Not a huge deal I guess, but they look out-of-whack in the GitHub view.

In general I've noticed the pattern seems to be that DOSBox SVN code (at least from the commits I've been looking at) has tabs, while DOSBox-X added code has spaces. Might be good to make it all consistent at some point. Right now I just try to make additions match whatever their surroundings are using.

joncampbell123:
@Allofich Not always. Tabs end up in DOSBox-X whenever I modify the code in Visual Studio 2017. Spaces are added because of my VIM configuration on Linux.

joncampbell123:
The intent is to use 4 spaces rather than hard tabs.

Allofich:
> Tabs end up in DOSBox-X whenever I modify the code in Visual Studio 2017.

Just in case you don't know, if you don't want that to happen in Visual Studio, you can change the options:
Tools -> Options -> Text Editor -> C/C++ -> Tabs ->Click "Insert spaces" instead of "Keep tabs"

Anyway I'll continue to just try to match tabs/spaces to the surrounding code for now, unless you ask me to do something else.

PR #1113, opened 2019-05-21
Take removed option out of reference .conf

Allofich:
In https://github.com/joncampbell123/dosbox-x/pull/1111, with regards to dosbox.reference.conf, I accidentally only removed the description of the removed option but left the default value in.

In https://github.com/joncampbell123/dosbox-x/commit/3e7ef9de73e9e6db294a7673d30d9d3387d1d61b,  you restored the description (was this intentional?)

The commit message for https://github.com/joncampbell123/dosbox-x/commit/a5822f0318eced02ad95a0bec21cee694035035c sounds like you are removing the option from the .conf file, but is actually removing something in dosbox.cpp.

Seems like things got a little confused (or I'm missing something?) This PR removes the option's description and default value setting from the reference .conf file.

### DIFF
--- a/dosbox.reference.conf
+++ b/dosbox.reference.conf
@@ -1319,11 +1319,6 @@ dongle=false
 #                                                      direct    Non-standard behavior, encode the CALL FAR directly to the entry point rather than indirectly
 #                                                   Possible values: auto, off, msdos2, msdos5, direct.
 #                                            share: Report SHARE.EXE as resident. Does not actually emulate SHARE functions.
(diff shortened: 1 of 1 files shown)


joncampbell123:
The dosbox.reference.conf file is generated by a script.

I had run it, noticed the return of the option, and then removed it from dosbox.cpp.

I guess I forgot to re-run the script.

Allofich:
Okay, thanks for explaining what happened.

PR #1121, opened 2019-05-29
Minor file reorganization

Allofich:
I noticed the `patch-integration` folder. I moved the skipped SVN commits file there since it seems appropriate, whereas the `NOTES` folder where I put it before seems like it is for notes about particular programs.

Removed the `VERSION` file as it seems to just be a leftover from the SVN-Daum merge.


joncampbell123:
If you have any bugfix only pull requests, please submit by 6PM PST tonight.

Tomorrow's release will begin building at 6PM PST tonight.

joncampbell123:
Thank you for your patience, the release is out. Pull requests are open again, and I have some development I want to do as well.

PR #1125, opened 2019-06-02
Fix value conversion warnings

Allofich:
# Description

Fixes many value conversion warnings.

Warnings that appear when building with Visual Studio/AppVeyor.

None known.

**Additional information**

Current master branch builds with 2103 warnings when built with the Visual Studio 2017 image in AppVeyor for x64. The vast majority of the warnings are for type conversions, such as a `Bit8u` value being assigned a `Bitu` value. With this PR the number of warnings are reduced to 1855.

There are still many more warnings to address, but I wanted to get your input before going further, in case you don't like my approach.

My approach:
1) Find a warning in the compiler output.
2) Go to the warning line in the code and find out which variable in the line is causing the issue.
3) Check all usage of the warning-causing variable and see if it can be narrowed. (Ex. It is `Bitu` but is never assigned anything larger than a `Bit16u`, or it is a counter variable that never goes higher than a number within the range of `Bit8u`) If so, change type to the narrower variable. This sometimes allows removing no-longer necessary casts.
4) If it's unclear/too hard to check all the usage cases of the variable, just stick a cast in front of the variable being assigned to to silence the warning.

To me step 3 seems like the preferable solution, and methods like `writeB(PhysPtr address, Bitu val)` are better as `writeB(PhysPtr address, Bit8u val)`, since they are supposed to only take a byte anyway. Narrowing the scope of variables allows for silencing type-conversion warnings without sticking casts everywhere (and allows removing some existing ones that were probably placed for that purpose). Maybe there is some reason I don't know of, though, that using `Bitu` is preferable.

Confirmed to compile and run.

### DIFF
--- a/include/callback.h
+++ b/include/callback.h
@@ -59,7 +59,7 @@ static inline PhysPt CALLBACK_GetBase(void) {
 	return (PhysPt)(((PhysPt)CB_SEG << (PhysPt)4U) + (PhysPt)CB_SOFFSET);
 }
 
-Bitu CALLBACK_Allocate();
+Bit8u CALLBACK_Allocate();
 
 void CALLBACK_Idle(void);
 
--- a/include/cpu.h
+++ b/include/cpu.h
@@ -147,7 +147,7 @@ void CPU_JMP(bool use32,Bitu selector,Bitu offset,Bitu oldeip);
 void CPU_CALL(bool use32,Bitu selector,Bitu offset,Bitu oldeip);
 void CPU_RET(bool use32,Bitu bytes,Bitu oldeip);
 void CPU_IRET(bool use32,Bitu oldeip);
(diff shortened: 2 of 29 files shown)


joncampbell123:
The patch is fine for the callback system but at some point I'm going to want to support callback IDs beyond 256 (beyond the range of Bit8u). Just FYI. When I do, I'll make that a typedef.

joncampbell123:
Typecasts and type changes should be handled carefully, it's possible for the fix to break emulation.

Some compiler warnings should be ignored if fixing them would easily break the code. On Linux for example, the dynamic core core generates compiler warnings that cannot be fixed without rewriting the code (and possibly breaking it).

joncampbell123:
I corrected one change you made that broke word-sized reads from the 3Dfx Voodoo emulation and might have caused problems.

PR #1136, opened 2019-06-09
Warning fixes

Allofich:
Some size-conversion fixes. Size-conversion warnings in Visual Studio 2017 build fall from 1843 to 1818.

Also 2 fixes to possible reads beyond the range of data arrays. These are warnings in the Visual Studio editor but not included in the above warning numbers.

### DIFF
--- a/include/fpu.h
+++ b/include/fpu.h
@@ -207,7 +207,7 @@ static INLINE void FPU_SetCW(Bitu word){
 }
 
 
-static INLINE Bitu FPU_GET_TOP(void) {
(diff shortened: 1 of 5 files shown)


Allofich:
The fixed range read warnings were:

src/dos/dos_programs.cpp line 1812:
`memcpy` accessed index 32, but the array being accessed only has up to index 31

src/dos/drive_cache.cp line 756:
`dirSearch[id]` is accessed. `id` could be `MAX_OPENDIRS` (which is 2048), but `dirSearch` is 2048 in size and so can only be accessed up to index 2047.

PR #1143, opened 2019-06-18
Revert SVN r4206

Allofich:
SVN r4206
https://github.com/joncampbell123/dosbox-x/commit/15338b598d676d953af53d21e96a77c9528164a2
causes _The Elder Scrolls: Daggerfall_ to crash DOSBox-X when started, with

`DYNREC:Can't run code in this page`

appearing twice in a row in the log.

This doesn't happen in mainline DOSBox, despite having the same commit in it (except that in the case of mainline commit, changes were also applied to `core_dyn_x86`, which was removed from DOSBox-X).

Reverting the commit stops the problem. Any idea what the cause is? Can we salvage the commit, and do you think it's useful? (It _sounds_ useful, as the original SVN commit message states it "fixes some hard to reproduce problems")

I'm putting off updating the skipped SVN commits file and removing the CHANGELOG entry for now.

### DIFF
--- a/include/paging.h
+++ b/include/paging.h
@@ -77,11 +77,9 @@ static inline Bitu MEMMASK_Combine(const Bitu a,const Bitu b) {
 #define PFLAG_READABLE		0x1
 #define PFLAG_WRITEABLE		0x2
 #define PFLAG_HASROM		0x4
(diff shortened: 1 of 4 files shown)


joncampbell123:
I don't fully understand the dynamic core and my focus is elsewhere, so I cannot answer questions about it.

PR #1144, opened 2019-06-18
MAME sound from SVN, except Tandy

Allofich:
The MAME sound from SVN. I was going to add the Tandy sound part as well, but it conflicted some with DOSBox-X so I skipped that part. 

The updated `gameblaster.cpp` in this PR includes the fixes that were made to the original `gameblaster.cpp` in DOSBox-X.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -133,6 +133,7 @@
     drive letter.
     - Move all stack alignment operations into one place and
     some optimalisations to RISC x64 dynamic core.
+    - Merged in MAME sound from mainline, except for Tandy sound. 
 0.82.19
   - Prefetch core fixed up, made more aggressive, and string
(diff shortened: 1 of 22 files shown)


joncampbell123:
The only modifications made to Tandy emulation is to trigger the mixer channel to render output every time registers are written so that the emulation is sample accurate (needed for Tandy music on the title screen of "Skate or Die" so the guitar is audible)

joncampbell123:
If you want to bring in the latest tandy.cpp all you have to do is add code to call MIXER_FillUp() on every write to the registers so that changes are immediately reflected in the output. I'll check on my end with the aforementioned DOS game.

I can ZIP up the game and provide it if you want.

Allofich:
OK, I'll take a look at it later.

> I can ZIP up the game and provide it if you want.

That's OK, I got it, so you don't need to do that, but thanks.

PR #1154, opened 2019-06-27
Updated build instructions for Ubuntu

Megabobster:
I fixed the formatting, changed it to recommend `build` instead of `autogen.sh` etc to properly compile the internal SDL, and added a few packages that I noticed `configure.sh` wanted. I also think the original line using wildcards is too general and will probably install unnecessary packages, but I'm a noob and I feel like that's outside my wheelhouse.


joncampbell123:
It's fine, you can use apt-get list or yum list to see what the patterns match and refine it later.

PR #1155, opened 2019-06-28
Fix typecast warnings

Allofich:
Fixes more typecast warnings.

Number of compile warnings in Visual Studio build down from 1365 in master to 1157.

Compiles and runs.

### DIFF
--- a/include/cpu.h
+++ b/include/cpu.h
@@ -142,10 +142,10 @@ bool CPU_LMSW(Bitu word);
 void CPU_VERR(Bitu selector);
 void CPU_VERW(Bitu selector);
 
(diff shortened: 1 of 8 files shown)


joncampbell123:
This commit breaks Windows 3.1.
I'm looking at possible causes here.

joncampbell123:
Apparently rm_off needs to remain Bitu in src/cpu/core_full/support.h or else Windows 3.1 cannot run properly.

joncampbell123:
I know it sounds odd, but even if you don't explicitly use core=full, DOSBox SVN and DOSBox-X fall back to full core in certain cases (such as recursively handling page faults when the INT 10h emulation callback function is running in vm86 mode).

PR #1177, opened 2019-07-08
Fix VS2019 compiler warnings

Allofich:
A few compiler warnings appear when using Visual Studio 2019 that don't appear with Visual Studio 2017.

```
'return': conversion from 'double' to 'float', possible loss of data include\util_math.h 30 14 dosbox-x
'sprintf' : format string '%d' requires an argument of type 'int', but variadic argument 2 has type 'Bitu' src\hardware\parport\printer.cpp 1591 34 dosbox-x
'initializing': conversion from 'double' to 'float', possible loss of data src\gui\sdl_mapper.cpp 1529 18 dosbox-x
'initializing': conversion from 'double' to 'float', possible loss of data src\gui\sdl_mapper.cpp 1528 18 dosbox-x
```

This PR silences them.

### DIFF
--- a/include/util_math.h
+++ b/include/util_math.h
@@ -27,7 +27,7 @@ struct DOSBox_Vector2
 	}
 
 	inline float magnitude(void) const {
-		return sqrt(sqrMagnitude());
+		return (float)sqrt(sqrMagnitude());
 	}
 
 	inline float sqrMagnitude(void) const {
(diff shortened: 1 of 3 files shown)


joncampbell123:
This is fine, though some C runtimes as far as I know offer sqrt() for double, sqrtf() for float, and on Linux, possibly sqrtl() for long double.

I think the same pattern holds for pow(), etc. as well

PR #1180, opened 2019-07-13
Minor fixes

Allofich:
Updated a few old copyright messages.

Reversed a check for SDL CD initialization. The check was incorrect and had been causing the message 

`WARN GUI:Failed to init CD-ROM support`

to appear in the log on startup.


joncampbell123:
I appreciate the help, I've been busy with my professional work the last week writing a minimal RTMP server to fill a gap needed by my clients for live streaming or I'd work on DOSBox-X more.

PR #1182, opened 2019-07-15
Fix arithmetic overflow and shift warnings

Allofich:
Fixes warnings that appear in the Visual Studio 2019 IDE about potential arithmetic overflows and shifting problems.

Most are for the case where you have an arithmetic operation between two variables, with the result being casted to a larger type. At least one of the variables needs to be casted to the larger type before the operation is done to make sure no overflow happens that wouldn't have happened with the larger type.

There are also a few fixes for shifts, where data could be lost when shifting using the smaller type. Casting to the larger type before the shift fixes the warning.

### DIFF
--- a/include/dma.h
+++ b/include/dma.h
@@ -121,7 +121,7 @@ public:
 	}
 	void SetPage(Bit8u val) {
 		pagenum=val;
-		pagebase=(Bitu)((Bitu)(pagenum >> DMA16_PAGESHIFT) << (Bitu)(16u + DMA16_PAGESHIFT));
+		pagebase=(pagenum >> DMA16_PAGESHIFT) << ((Bitu)16u + DMA16_PAGESHIFT);
 	}
 	void Raise_Request(void) {
 		request=true;
--- a/src/aviwriter/riff_wav_writer.cpp
+++ b/src/aviwriter/riff_wav_writer.cpp
@@ -95,7 +95,7 @@ int riff_wav_writer_set_format_ex(riff_wav_writer *w,windows_WAVEFORMATEX *f,siz
 		return 0;
 
 	w->fmt_len = sizeof(windows_WAVEFORMAT);
-	if (__le_u16(&f->cbSize) != 0u) w->fmt_len += 2u + __le_u16(&f->cbSize);
+	if (__le_u16(&f->cbSize) != 0u) w->fmt_len += (size_t)2u + __le_u16(&f->cbSize);
 	if (w->fmt_len > len)
 		return 0;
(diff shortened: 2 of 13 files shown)


joncampbell123:
Ehh.... I tend to be very cautious of compiler warnings of that type because you can easily break the code fixing them unless you fix it correctly. I'm also wary that what makes VS2019 shut up could trigger compiler warnings on GCC about data type truncation.

By the way if typecasting an int as (Bitu) makes the compiler shut up you might try just appending a "u" suffix to the integer instead to indicate that it is unsigned. "ul" makes it unsigned long, "ull" unsigned long long, etc.

I'll go over these very carefully and check that it doesn't break anything. I suggest compiling the code on your end and checking that it doesn't cause any breakage on your end as well.

joncampbell123:
Testing now.

joncampbell123:
I would like someone at the GCC project explain this:

In file included from dos_programs.cpp:40:0:
../../include/dma.h: In member function ‘void DmaChannel::SetPage(Bit8u)’:
../../include/dma.h:124:41: warning: conversion to ‘Bit32u {aka unsigned int}’ from ‘int’ may change the sign of the result [-Wsign-conversion]
   pagebase=(pagenum >> DMA16_PAGESHIFT) << ((Bitu)16u + DMA16_PAGESHIFT);
                                         ^
All variables in that expression and the destination are unsigned.

So, unsigned = unsigned >> unsigned becomes signed at the shift operator? WTF GCC?

This nonsense is why -Wsign-conversion is commented out in configure.ac, or I'd enable it by default.

Allofich:
Maybe making it

pagebase=(Bit32u)((pagenum >> DMA16_PAGESHIFT) << ((Bitu)16u + DMA16_PAGESHIFT));

would silence that, even though it doesn't seem like it should be any different.

Allofich:
My guess of what's happening.

The original line was

`pagebase=(Bitu)((Bitu)(pagenum >> DMA16_PAGESHIFT) << (Bitu)(16u + DMA16_PAGESHIFT));`

I took out the two (Bitu) casts at the front. `pagenum` and `DMA16_PAGESHIFT` are both `Bit8u`. 

At

https://stackoverflow.com/questions/20236483/strange-gcc-short-int-conversion-warning

someone says "If an int can represent all values of the original type, the value is converted to an int; otherwise, it is converted to an unsigned int."

If that is so, `(pagenum >> DMA16_PAGESHIFT)` is converted to an int. The << is applied to it, then it is converted to unsigned int.

If this is the case, I probably shouldn't have removed that second Bitu cast. Seems like it should be a Bit32u cast, though, not a Bitu cast.

PR #1188, opened 2019-07-27
DOS CON: Don't disable ANSI when resetting the attribute

nmlgc:
This got overlooked when [SVN r4097](https://sourceforge.net/p/dosbox/code-0/4097/) was merged in 65b73d0, which consequently broke `ESC [0m` for all non-PC-98 architectures.


joncampbell123:
Good catch, thanks!

PR #1197, opened 2019-08-10
Minor fixes

Allofich:
Fixes some more arithmetic overflow warnings.
Remove some unnecessary semicolons, parentheses, incorrect casts (casting to int for a Bit8u assignment).

Compiles and runs.

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -3199,7 +3199,7 @@ int utf8_encode(char **ptr,char *fence,uint32_t code) {
             *p++ = (char)(0x80 | ((code >> 6) & 0x3F));
             *p++ = (char)(0x80 | (code & 0x3F));
             break;
-    };
+    }
 
     *ptr = p;
     return 0;
@@ -3268,7 +3268,7 @@ int utf8_decode(const char **ptr,const char *fence) {
             c = (unsigned char)(*p++); if ((c&0xC0) != 0x80) return UTF8ERR_INVALID;
             ret |= c&0x3F;
             break;
-    };
+    }
 
     *ptr = p;
     return ret;
--- a/src/gui/zipfile.cpp
+++ b/src/gui/zipfile.cpp
@@ -63,7 +63,7 @@ int ZIPFileEntry::read(void *buffer,size_t count) {
     if (file == NULL || file_offset == (off_t)0) return -1;
     if (position >= file_length) return 0;
 
-    size_t mread = size_t(file_length - position);
+    size_t mread = (size_t)file_length - position;
     if (mread > count) mread = count;
 
     if (mread > 0) {
--- a/src/ints/bios.cpp
+++ b/src/ints/bios.cpp
@@ -3903,9 +3903,7 @@ void PC98_BIOS_SCSI_CALL(void) {
 
                     for (i=0;i < ssize;i++) PC98_BIOS_FLOPPY_BUFFER[i] = mem_readb(memaddr+i);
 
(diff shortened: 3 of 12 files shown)


joncampbell123:
I have some reservations about accepting commit https://github.com/joncampbell123/dosbox-x/pull/1197/commits/a17ec7e73ed199fe250ba0bf5c29c5c48d63da0c

The typecasting in include/pic.h is there to make sure the result of CPU_CycleMax-CPU_CycleLeft-CPU_Cycles is the correct type.

src/gui/zipfile.cpp is done for the same reason, to make sure the result is size_t. Also size_t(...) is a valid typecast in C++ and recommended going forward as is the use of static_cast<...>() and reinterpret_cast<...>()

Allofich:
As I understand it, for example, if you have:

```
int x;
int y;
long long z;

z = ((long long)x + y);
```
then z will be assigned a value of type `long long`, because casting at least one of the values in an operation to a larger value will make the result be of that larger value. (long long)(x + y) won't account for overflow in (x+y), but after we take care of the overflow, I don't think we need another (long long) cast outside the (x + y) parentheses, like (long long)((long long)x + y).

Testing with
https://www.onlinegdb.com/online_c++_compiler
I tried:

```
    int x = 2147483647; // Maximum value of int
    int y = 1; // A value of 1 to go past the maximum
    long long z = 0;
```

then

```
    z = (x + y); // Overflow
    cout<<z<<std::endl;
    z = (long long)(x + y); // Overflow
    cout<<z<<std::endl;
    z = ((long long)x + y); // OK
    cout<<z<<std::endl;
```
Results are:
-2147483648  (overflow)                                                                                                
-2147483648  (overflow)                                                                                         
2147483648   (OK)

In the final case, which mirrors the casting I'm doing here, z is correctly storing a non-overflowed value, as 2147483648, which is 1 beyond the range of `int`.

From the above, it seems OK to change
(64bit type) = (64bit type)(x + y)
to
(64bit type) = ((64bit type)x + y)

but is it not OK?

> The typecasting in include/pic.h is there to make sure the result of CPU_CycleMax-CPU_CycleLeft-CPU_Cycles is the correct type.

I'm not quite sure why (maybe because I have since updating the Windows SDK I'm using), but while this line showed up earlier in the Visual Studio 2019 IDE as having an arithmetic overflow warning, it doesn't now, so I'll revert this. Looking at the types it seems it should be fine as it was.

joncampbell123:
This is what GCC complains about with -Wsign-conversion, which is why the type conversion was introduced.

The last time I compiled with VS2017 this did not trigger any warnings, nor did it cause any warnings with Clang/LLVM on my Macbook Pro.

    zipfile.cpp: In member function ‘int ZIPFileEntry::read(void*, size_t)’:
    zipfile.cpp:66:42: warning: conversion to ‘long unsigned int’ from ‘off_t {aka long int}’ may change the sign of the result [-Wsign-conversion]
         size_t mread = (size_t)file_length - position;
                                          ^

Allofich:
> The last time I compiled with VS2017 the original form of this code did not trigger any warnings

These "arithmetic overflow" warnings are not compiler warnings, so you won't see them there. They are part of on-the-fly code analysis that happens while as you look at the code in Visual Studio 2019.

Based on https://github.com/joncampbell123/dosboxx/commit/8e0fedd3e542f01a7acf23f48eb8c7ff68b5e295, I guess when there is conversion between unsigned/signed, we need to cast each operand to the result's type to avoid that warning? I'll try to keep that in mind.

PR #1201, opened 2019-08-12
More SVN commits

Allofich:
Continuing to go over skipped SVN commits.

https://sourceforge.net/p/dosbox/code-0/3968/
Causes a beep to play. You can hear it by pressing CTRL+G to type "^G" at the command prompt. (Note: There seems to be a difference in command prompt handling with MS-DOS. When testing with MS-DOS in the 86Box emulator, just pressing CTRL+G won't cause a beep, but running "echo ^G" will)

https://sourceforge.net/p/dosbox/code-0/3980/
The main purpose of this one is already in DOSBox-X, but it included returning a failure when the disk isn't active, which isn't in DOSBox-X. This PR adds it in.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,9 @@
   - Integrated commits from mainline (Allofich)
     - In the mapper, display disabled items or events
     with no binding in grey. 
+    - Implemented BIOS beep sound for ASCII character 7
+    - Return failure for INT 13 format calls if the
+    drive is inactive.
 0.82.20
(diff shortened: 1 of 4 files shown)


Wengier:
@Allofich You have done a good job of integrating SVN commits to DOSBox-X before. If possible, can you please try to integrate the recent SVN commits as well? Thanks! Let us know if you need help to integrate any recent SVN commits.

Allofich:
Sure, I've been intending to do that. I'll try to do it soon.

PR #1202, opened 2019-08-13
Fix uninitialized member variable warnings (VS2019 IDE)

Allofich:
Fixes several uninitialized member variable warnings in the Visual Studio 2019 IDE (code analysis warnings, not compiler warnings). Just initializing everything to 0 or false. In one case it was an enum so the 0 enum is used.

Also fixes an assignment of a boolean to it's opposite that looked like x =! x, to make it look as expected, like x = !x.

### DIFF
--- a/include/8255.h
+++ b/include/8255.h
@@ -115,28 +115,28 @@ public:
     }
 public:
     //! Port A write mask. Controls which bits are writeable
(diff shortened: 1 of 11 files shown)


joncampbell123:
@Allofich So you like it when negation looks like an operator? :)

PR #1205, opened 2019-08-18
SVN GUS commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/4001/
and
https://sourceforge.net/p/dosbox/code-0/4004/

4004 is a fix for 4001.

A few things from the SVN 4001 commit are omitted here:

- Removing auto amplification
DOSBox-X specifically has a setting for it, so I thought you might want to keep it, so it's not removed by this PR.

- Enabling GUS logging
When this is enabled you get a log full of "Read GUS at port ---" when using it, so it's not enabled by this PR.

- Slight rounding offset in the panning tables to match original GUS
tables
This part of the file has been redone in DOSBox-X so the SVN changes don't apply.

GUS seems to work the same before and after testing this.

### DIFF
--- a/patch-integration/Skipped SVN commits.txt
+++ b/patch-integration/Skipped SVN commits.txt
@@ -1,9 +1,7 @@
 Commit#:	Reason for skipping:
 
-4001		Conflict
 4002		Conflict
 4003		Conflict
-4004		Conflict
 4007		Conflict
 4008		Conflict
 4009		Conflict
--- a/src/hardware/gus.cpp
+++ b/src/hardware/gus.cpp
@@ -43,10 +43,9 @@ enum GUSType {
 };
 
 //Extra bits of precision over normal gus
-#define WAVE_BITS 2
-#define WAVE_FRACT (9+WAVE_BITS)
+#define WAVE_FRACT 9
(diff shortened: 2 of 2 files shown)


Allofich:
Noticed a potential problem.

The SVN commit changes `WAVE_FRACT` from being `(9+WAVE_BITS)` to `9`.

I don't know what this change is about, it's not described in the commit message or any comments. The potential problem is that there is a usage of `WAVE_FRACT` in DOSBox-X not in SVN.

Line 260
`WaveAdd = ((Bit32u)(val >> 1)) << ((Bit32u)(WAVE_FRACT-9));`

Also, with `WAVE_FRACT` being 9 now, its usage here become pointless because its just a shift by 0.

Allofich:
Sorry for the force-pushes, reworked the autoamp part. Should be OK now.

joncampbell123:
I don't remember clearly but there was a modification I made to have the GUS render at the actual sample rate the DAC would run rather than upconvert to 44.1KHz as SVN does by default.

To explain what that means, the GUS hardware sample rate varies depending on how many active channels you're using. At 14 channels the DAC renders at 44.1KHz and at 32 it's about ~19KHz.

Programs have to factor that in to program the correct fractional step values to get the sample rate they need.

PR #1218, opened 2019-08-25
Fix -Wmisleading-indentation GCC warnings

Allofich:
Trying to build with GCC, there are several warnings that appear. Many of them (19) are `-Wmisleading-indentation` warnings, and they are all fixed by this PR.

There were two cases causing this warning:
1. There were some cases in `debug.cpp` where there is a line like:
`if(dline_len < 28) for (c = (Bitu)dline_len; c < 28;c++) dline[c] = ' '; dline[28] = 0;`

where `dline[28] = 0;` would trigger the warning. I think this line is supposed to be within the if block (in other similar code in the same file it is within the if block), so in addition to silencing the warning I put it within the if block.

2. Most cases were just caused by a mixture of using tabs and spaces.

```    if (IS_PC98_ARCH)	    if (IS_PC98_ARCH)
        PROGRAMS_MakeFile("PC98UTIL.COM",PC98UTIL_ProgramStart);

	PROGRAMS_MakeFile("CAPMOUSE.COM", CAPMOUSE_ProgramStart);
```

This would cause it because of the tab before `PROGRAMS_MakeFile("CAPMOUSE.COM", CAPMOUSE_ProgramStart);`

I know you are only accepting last-minute bug fixes, and I don't know if you consider silencing warnings as bug fixes, but here it is. If you want to wait until after the release that is fine with me. I did check the debugger with this PR just in case (the only part with slightly modified behavior) and didn't see any difference.

### DIFF
--- a/src/debug/debug.cpp
+++ b/src/debug/debug.cpp
@@ -1232,7 +1232,10 @@ static void DrawCode(void) {
 		// Spacepad it up to 28 characters
         if (no_bytes) dline[0] = 0;
 		size_t dline_len = strlen(dline);
-		if(dline_len < 28) for (c = (Bitu)dline_len; c < 28;c++) dline[c] = ' '; dline[28] = 0;
+        if (dline_len < 28) {
+            for (c = (Bitu)dline_len; c < 28; c++) dline[c] = ' ';
+            dline[28] = 0;
+        }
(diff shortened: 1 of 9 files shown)


joncampbell123:
This pull request falls under the "bug fixes only" category, does not add any new features, so I will merge the request.

PR #1222, opened 2019-08-27
Fix more GCC warnings

Allofich:
Fixes for more GCC warnings.

Fixed the one warning about a possibly uninitialized variable being used.
Fixed 26 warnings about unused return values, mostly for `fread`.


Allofich:
Please hold for a bit, making a modification.

Allofich:
Should be okay now.

Allofich:
I'll need some time, but I'll try to verify them when I can.

Allofich:
Loading codepages and images are unaffected by the changes as far as I see. Only limited testing, but when I tested with some codepage files and image files the tests for the return values from `fread()` are passing.

Still, seems sensible to wait until after the release since we're so close and this PR is just to silence warnings, not to fix any known bugs.

joncampbell123:
Will hold until after 0.80.21 release.

PR #1234, opened 2019-09-01
Fix warnings

Allofich:
1. There were some value conversion warnings that appeared in the Visual Studio 2017 builds recently, fixed so there are 0 warnings again.

2. Fixed various GCC warnings. Building with GCC 9 I see no GCC code warnings now with the code itself. There is still this:

```
dosbox.cpp:968:6: note: variable tracking size limit exceeded with ‘-fvar-tracking-assignments’, retrying without
  968 | void DOSBOX_SetupConfigSections(void) {
```

and various warnings about unrecognized warning flags, issues in the configure file, etc., but no more warnings for the code itself.

3. Fixes one non-code warning, about renaming `configure.in` to `configure.ac`.

Compiles and runs.


Allofich:
There were 3 code warnings when running with Clang 8. Fixed 2 of them.

The remaining one is already commented in the code, wasn't sure what to do about it.

```
gui_tk.cpp:149:45: warning: result of comparison of constant 91 with expression
      of type 'GUI::Font::SpecialChar' is always true
      [-Wtautological-constant-out-of-range-compare]
  ...if (font->toSpecial(text[seqstart++]) != '[') break; /* FIXME: Clang/LLV...
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~
```

PR #1247, opened 2019-09-13
Minor fixes

Allofich:
1. Fix a few Visual Studio value conversion warnings to bring it back to 0 warnings.
2. Fix a couple GCC "misleading indentation" warnings caused by the mix of tabs and spaces. (Probably the whole codebase should be converted from tabs to spaces, as these warnings will continue to pop up, and the mixture looks uneven when viewing the code. That would touch pretty much every file, though. Do you care about it interfering with the "blame" GitHub function?) 
3. Fix the Clang warning about the comparison that is "always true". The reason it said that seems to be that the comparison was comparing the Font:SpecialChar enum against the "[" character, or 91, for which there is no SpecialChar enum. There doesn't seem to be any reason to cast one side of the compare to SpecialChar as was being done there, so I changed it to just compare the char values.
4. Also fix another Clang FIXME in the code, where for some reason a boolean was being cast to a char, then used in a switch statement with only two cases for 1 and 0. Changed it to an if-else. Also removed a pointless return statement at the end of this function.
5. Removed a couple warning flags (`Wno-unknown-warning` and `-Wno-extended-offsetof`) that cause a bunch of warnings in GCC and Clang compiler output about being unrecognized. I don't see either of these in recent documentation for GCC and Clang, but I'm not all that familiar with building with GCC and Clang.
GCC: https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc/Warning-Options.html
Clang: https://releases.llvm.org/8.0.0/tools/clang/docs/DiagnosticsReference.html#wint-to-void-pointer-cast

Compiles and runs.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -319,8 +319,6 @@ if test x$host = xi386-pc-os2-emx ; then
     LDFLAGS="$LDFLAGS -Zomf -Zmt"
 fi
 
-AC_CHECK_CXXFLAGS([ -Wno-unknown-warning ])
-
 dnl I would like to know of any concerns given by the C++ compiler.
 dnl Clang/LLVM already does this to some degree, let's get GCC to do it too.
 AC_CHECK_CXXFLAGS([ -Wall ])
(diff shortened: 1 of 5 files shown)


joncampbell123:
DOSBox-X inherits the hard tabs from DOSBox SVN, but should always use soft tabs (4 spaces) where possible.

PR #1249, opened 2019-09-14
SVN r4009

Allofich:
Implemented https://sourceforge.net/p/dosbox/code-0/4009/, which I skipped over last time because I wasn't sure it was compatible with DOSBox-X, but it seems to be, and removed other SVN commits from the list of skipped commits after checking them and seeing they are irrelevant to DOSBox-X.

4009's SVN commit message is
"Handle errant IRQs as a real BIOS does. Fixes Tandy DAC in Chuck Yeager's Air Combat. Also remove r3263 workaround, as it's no longer needed."

I'm not sure if the sound is in fact better in "Chuck Yeager's Air Combat". It might be a little better, but I couldn't discern a clear difference. The r3263 workaround is for "Design Your Own Railroad", which appears to still work fine when the workaround is removed.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -57,6 +57,10 @@
   - LABEL reimplemented to imitate MS-DOS behavior
     with regard to how it handles the command line.
   - File I/O checking and cleanup (Allofich)
+  - Integrated commits from mainline (Allofich)
+    - Handle errant IRQs as a real BIOS does. Also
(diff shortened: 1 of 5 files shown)


joncampbell123:
I'm going to do a quick Win32 compile Saturday morning and release it as a beta.

There are some new configuration GUI changes that some here are waiting to try.

PR #1250, opened 2019-09-14
SVN r4256

Allofich:
https://sourceforge.net/p/dosbox/code-0/4256/

"Fix flag behaviour of several shift/rotate instructions, cause
exceptions and fix potential 'pop ss' problems"

I skipped ahead in the SVN commits because there was this interesting one that I wanted to try.

Although they haven't marked it as fixed yet, this fixes this bug in the DOSBox mainline bug tracker:
https://sourceforge.net/p/dosbox/bugs/511/

Before this PR, the program outputs "1". With this PR, it outputs "0". I don't know of any other test cases, but it would be interesting if this fixes some programs.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -60,7 +60,10 @@
   - Integrated commits from mainline (Allofich)
     - Handle errant IRQs as a real BIOS does. Also
     remove r3263 workaround, as it's no longer
-    needed.	
+    needed.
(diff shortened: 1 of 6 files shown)


joncampbell123:
It looks like it's mostly fixes for the full core, but the general shift case code might be helpful. If it breaks anything I'll make commits to fix it.

PR #1251, opened 2019-09-15
Mouse cleanups

Allofich:
Cleanups to `mouse.cpp`.

1. Removed undefined function declaration `WriteMouseIntVector`
2. Fixed arithmetic overflow warnings (Visual Studio 2019 real-time code analysis warnings when viewing the file, not compiler warnings)
3. Fixed misspelling of "horizontal" in log messages.
4. Cleaned up the INT 33 function.
 - Formatting cleanup, remove unnecessary whitespace
 - Re-order case 0x21 to be in order where one would expect to find it. Used a `goto` to recreate the fallthrough from case 0x00.
 - Added stubs for 4 unimplemented cases
5. Changed some log messages for normal operation (ex. setting the mouse's x and y range) to no longer be warnings, but normal log messages.
6. Some log messages were categorized as `LOG_MISC` or `LOG_KEYBOARD`. Changed to `LOG_MOUSE`.

### DIFF
--- a/src/ints/mouse.cpp
+++ b/src/ints/mouse.cpp
@@ -1253,14 +1253,14 @@ static Bitu INT33_Handler(void) {
     case 0x1a:  /* Set mouse sensitivity */
         // ToDo : double mouse speed value
         Mouse_SetSensitivity(reg_bx, reg_cx, reg_dx);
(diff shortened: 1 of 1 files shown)


Allofich:
Hold on a moment, I want to minimize the formatting fixes.

PR #1252, opened 2019-09-16
Clean up to INT 21 handler

Allofich:
- Fix arithmetic overflow warnings (VS 2019 code analysis)
- Correct levels of log messages to be consistent with other areas ( `LOG_NORMAL` for just reporting INT 21 calls, `LOG_ERROR` for reporting unimplemented INT 21 calls)
- Put INT 21 calls in order, with goto to fallthrough to the default case or to go to other calls that do identical behavior.
- Changed INT 21 calls that issue an "unimplemented" warning in the default case to not include case 0x6b, which is a NULL function and already doing what it is supposed to do.
- Added a case for 0x6a, which according to http://www.ctyme.com/intr/rb-3176.htm is "identical to AH=68h in DOS 5.0-6.0"
- Added more descriptive "not implemented" messages for some calls, using http://www.ctyme.com/intr/int-21.htm as reference.
- Fixed typo

### DIFF
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -1040,7 +1040,7 @@ static Bitu DOS_21Handler(void) {
                     reg_al = 0x00;
                     SegSet16(ds,dos.tables.dpb);
                     reg_bx = drive*dos.tables.dpb_size;
-                    LOG(LOG_DOSMISC,LOG_ERROR)("Get drive parameter block.");
+                    LOG(LOG_DOSMISC,LOG_NORMAL)("Get drive parameter block.");
                 } else {
                     reg_al=0xff;
                 }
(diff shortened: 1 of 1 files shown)


joncampbell123:
@Allofich When merging changes from SVN please look out for a known issue where some Windows 3.1 Win32s applications crash with a DOS error when run under SVN but work perfectly fine with DOSBox-X.

Allofich:
OK, sure. If it's a recent issue in DOSBox SVN, maybe I can find which commit caused it.

Allofich:
I haven't been able to reproduce the problem with SVN. I've been trying with the EmuCR SVN build for r4252.
I don't have the sound and video drivers set up, so the game starts without sound and with messed up colors, so that might be preventing the problem somehow, but in both DOSBox-X and SVN I am able to start the demo and load up the first level, then exit the game from the menu, without any errors.

I also tried with the dosbox.conf you provide here:
https://github.com/joncampbell123/msdostest/tree/master/the-incredible-machine-2-demo-classic-pc-games-ia
but the results were the same.

Issue #1099, opened 2019-05-11
Print screen and corresponding interrupt support (following DOSBox SVN patch)


joncampbell123:
**Is your feature request related to a problem? Please describe.**
Add emulation of Print Screen key (and interrupt) to DOSBox-X

See also:

https://sourceforge.net/p/dosbox/code-0/3961

Allofich:
I think we can close this now.

PR #1261, opened 2019-09-27
SVN r3961 (modified)

Allofich:
# Description

https://sourceforge.net/p/dosbox/code-0/3961/

Adapted for DOSBox-X. Please take a close look, because I had to do things differently than the SVN commit for the INT 5 part. I tried to mimic existing code in DOSBox-X but I could have made a mistake.

https://github.com/joncampbell123/dosbox-x/issues/1099

Print Screen key and INT 5 support.

Not that I know of.

**Additional information**

This implements:
1. Print Screen key, which did nothing before. This allows the screenshot feature of Descent to work.

descentr_000

2. Pressing Print Screen stopping the repetition of another key that is being held down when Print Screen is pressed. The Pause key already does this in DOSBox-X, and I saw the Print Screen key will likewise stop another key's repetition in real MS-DOS in 86Box and in the command prompt in Windows 10, so I assume that is correct behavior. In mainline DOSBox Print Screen does not interrupt key repetitions.

3. INT 5 support, as provided by the SVN commit. Unfortunately, I haven't been able to test that it works. The original SVN commit claims that "Horror Hotel (TSR text adventure)" and "The Forgotten Land" benefit from this, but I can't find them to test with.

In `callback.cpp` I noticed that DOSBox-X had some separate behavior in place for PC-98 and Tandy, in `CB_IRQ1`, so this PR's addition there might not be appropriate for them.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -98,6 +98,9 @@
     - Fix flag behavior of several shift/rotate
     instructions, cause exceptions and fix
     potential 'pop ss' problems
+    - Add support for Print Screen key and
+    interrupt. In-game screenshot feature
+    of Descent and Descent 2 now works.
 0.82.21
   - Reduced title bar size of the Configuration GUI.
   - Fixed sizing and positions of some Help menu dialog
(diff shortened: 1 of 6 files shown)


joncampbell123:
PC-98 does not have INT05 as Print Screen, of course.

INT05 also happens to be the exception generated by BOUND on 80186 or higher CPUs.

joncampbell123:
Don't sweat the debugger mistake, it's an easy one to make.

If you imported the change from SVN, it probably broke the debugger interface on their project too.

joncampbell123:
I made a few changes to the code only to keep scan codes in hexadecimal. I can confirm in the debugger that pressing Print Screen invokes INT05.

By the way one good way to test Print Screen is to start up Windows 3.1, press "Print Screen" and you should be able to paste a screenshot into something like Paintbrush (PBRUSH.EXE) after that.

PR #1265, opened 2019-09-28
Note and ignore Logitech CyberMan INT 33 call

Allofich:
I noticed that both The Elder Scrolls: Arena and Descent would log a message about unimplemented mouse function 0x53C1.

Descent's source code is online, so I took a look and found the call, and that it is to detect support for "SWIFT functions" used by the Logitech CyberMan mouse, provided by that mouse's drivers.

Descent code:
https://github.com/videogamepreservation/descent/blob/05216a7e6953aa82902d9107109fa7ffbadd9710/BIOS/MOUSE.C#L346

Logitech CyberMan:
http://www.computinghistory.org.uk/det/12416/Logitech-Cyberman/

Also a DOOM message referencing using the 0x53C1 function to check for CyberMan "SWIFT" support is mentioned here: https://www.doomworld.com/forum/topic/37039-doom-on-lan/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.82.22 (next)
+  - (Allofich) Note and ignore INT 33, AX=53C1 call
+    for the Logitech CyberMan.
   - Fixed odd code addition that disabled (through
     the configuration) XMS, EMS, and UMB when
     booting a guest OS. This fixes a problem where
--- a/src/ints/mouse.cpp
+++ b/src/ints/mouse.cpp
@@ -1340,6 +1340,9 @@ static Bitu INT33_Handler(void) {
         reg_cx = (Bit16u)mouse.max_x;
         reg_dx = (Bit16u)mouse.max_y;
         break;
+    case 0x53C1: /* Logitech CyberMan */
+        LOG(LOG_MOUSE, LOG_NORMAL)("Mouse function 53C1 for Logitech CyberMan called. Ignored by regular mouse driver.");
+        break;
     default:
         LOG(LOG_MOUSE, LOG_ERROR)("Mouse Function %04X not implemented!", reg_ax);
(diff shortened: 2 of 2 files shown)


joncampbell123:
Is there any documentation online what extended functions the Logitech CyberMan adds to INT 33h?

That documentation would be a good addition to hackipedia.org and comments in the source code.

Allofich:
I'm not aware of any. The Descent code additionally calls 0x5301, which apparently gets x and y.

https://github.com/videogamepreservation/descent/blob/05216a7e6953aa82902d9107109fa7ffbadd9710/BIOS/MOUSE.C#L602

Also, The Elder Scrolls: Arena can call 0x53C2 if 0x53C1 returns ax as 1 (if SWIFT functions are supported) and if another condition is also met.

The Elder Scrolls: Arena function (I have not really deciphered any of it, and am not actively trying to) looks like this:

```
;; fn0800_5F42: 0800:5F42
;;   Called from:
;;     0800:9ACA (in fn0800_9AC6)
fn0800_5F42 proc
	mov	word ptr [8FA0],0000
	mov	ax,53C1
	les	dx,[9104]
	int	33
	cmp	ax,0001
	jnz	5F76

l0800_5F56:
	cmp	byte ptr [9108],01
	jnz	5F76

l0800_5F5D:
	or	word ptr [8FA0],01
	mov	ax,53C2
	int	33
	and	ax,0003
	cmp	ax,0001
	jnz	5F74

l0800_5F6F:
	or	word ptr [8FA0],02

l0800_5F74:
	clc
	ret

l0800_5F76:
	stc
	ret
```

joncampbell123:
Apparently not even Ralph Brown knows about those INT 33h calls.

http://www.ctyme.com/intr/int-33.htm

PR #1275, opened 2019-10-05
Minor correction

Allofich:
Pedantic fix. The screen the game was freezing on was a password input screen.


joncampbell123:
Ah, it's a password.

That's what I get for not understanding Chinese writing.

PR #1311, opened 2019-11-03
Windows Explorer context menu entries

aybe:
# Description

Adds context menu entries in Windows Explorer for easy invocation of DOSBox-X

no

yes

"Open with DOSBox-X" context menu entry:

- starts DOSBox-X
- mounts the directory on which this context menu entry was invoked on as drive C
- switch to C drive

i.e. start DOSBox-X and mount current directory as C in two mouse clicks from Windows Explorer

no

**Additional information**

What they do:

- scripts checks for admin permissions and instruct what user has to do if they're not available
- installer searches for DOSBox-X executable and uses first one found, if any
- registers entries in registry: `Directory` but also `Directory\Background` for an easier usage
- uninstaller provided for removing them

Delivery method: 

These could be added to Windows builds, next to the DOSBox-X executable.


joncampbell123:
It might be better (and might not require Admin privileges) if your script were to install this under HKEY_CURRENT_USER\Software\Classes (the current user's Class registration) rather than for all users.

This is something that would probably be better to install per user than systemwide.

aybe:
Absolutely, just learned something new today!

Now they set it up only for current user and at same time do not require admin rights anymore!

joncampbell123:
Does it work on your end?

Have you verified the HKEY_LOCAL_MACHINE keys are not active?

aybe:
Yes, I've tested it before pushing it. 

Funny that you're talking about HKLM, even funnier, I was modifying HKCR directly which seems very odd to me now... I checked HKLM and no entries are to be found in there. By the way, I uninstalled the old version leveraging HKCR first before developing the new version, so there weren't any leftovers.

Now that this little thing works as expected, I am starting to see its shortcomings, namely that it's not possible to, say, pick an ISO to use as a CD-ROM then pick a directory to use as a HDD -> static context menu entries are simply not up to the job. A rough guess is that one should code something using COM but unfortunately I this is beyond my knowledge.

PR #1327, opened 2019-11-07
Brighter properties + line break for clarity.

aybe:
# Description

Better GUI help, brighter property name + line break for clarity.

no

See description.

Probably not, text is still fully rendered to the end

**Additional information**

http://www.lihaoyi.com/post/BuildyourownCommandLinewithANSIescapecodes.html#colors

Now something else appears ! There are line breaks everywhere in the docs, probably someone had in mind that old school idea but now it looks weird. IMO, there shouldn't be any and whoever is editing dosbox.conf should leverage _line wrap_ in its text editor instead ! If that ever has to be changed, many places to update in perspective...

Old version:
dosbox-x_2019-11-07_01-56-40

New version:
dosbox-x_2019-11-07_01-56-44

I picked white since it's really contrasted but maybe you'll prefer something else (dark red/magenta?), for reference:

dosbox-x_2019-11-07_01-49-01

dosbox-x_2019-11-07_01-49-02

dosbox-x_2019-11-07_01-49-03

dosbox-x_2019-11-07_01-49-04

dosbox-x_2019-11-07_01-49-05

dosbox-x_2019-11-07_01-49-06

dosbox-x_2019-11-07_01-49-07

dosbox-x_2019-11-07_01-49-08

dosbox-x_2019-11-07_01-49-09

dosbox-x_2019-11-07_01-49-10

dosbox-x_2019-11-07_01-49-11

dosbox-x_2019-11-07_01-49-12

dosbox-x_2019-11-07_01-49-13

dosbox-x_2019-11-07_01-49-14

dosbox-x_2019-11-07_01-49-15

dosbox-x_2019-11-07_01-49-16

### DIFF
--- a/src/gui/sdl_gui.cpp
+++ b/src/gui/sdl_gui.cpp
@@ -697,7 +697,7 @@ public:
             Property *p;
             int i = 0;
             while ((p = sec->Get_prop(i++))) {
-                msg += std::string("\033[34m")+p->propname+":\033[0m "+p->Get_help()+"\n";
+                msg += std::string("\033[37;1m")+p->propname+":\033[0m\n"+p->Get_help()+"\n\n";
             }
             if (!msg.empty()) msg.replace(msg.end()-1,msg.end(),"");
             setText(msg);


joncampbell123:
I like the blue header text personally, what do you like @aybe?

aybe:
Currently it's doing bright white as I was looking for clarity but it feels a little too contrasted.

What I don't like with dark blue is that it's very difficult to differentiate from help content.

If we zoom very closely, it turns out that green and dark magenta are not too bad:





chrome_2019-11-07_02-39-47

Colors and tastes ... maybe we should just toss a coin at this point!

joncampbell123:
Good point. Red seems like a reasonable choice to me.

Issue #1326, opened 2019-11-07
Building solution now also requires Windows SDK 10.0.17134.0?


aybe:
**Describe the bug**

Not only Windows 8.1 SDK is required but also 10.0.17134.0 is.

Shouldn't it require one or the other, but not both ? Someone please clarify.

joncampbell123:
It should be building all with the same SDK.

If it is possible to build with SDK 10.0.17134.0 but retain the ability for the binary to run on Windows 7 through 10, then all should use the 10.0.17134.0 SDK.

aybe:
Okay, I'll see what I can do.

aybe:
If you do this:

https://github.com/joncampbell123/dosbox-x/commit/2469ecb87db44eabdc503a8de89cb2f0084a9537#diff-ac3b1ae90cb3797dfc4f060f3be32a1d

Then you don't need to upgrade platform toolset anymore (and edit all projects), but then you get:

_error MSB8020: The build tools for Visual Studio 2010 (Platform Toolset = 'v100') cannot be found._

This one is not available in the installer, maybe this one project (freetype) should be upgraded to 2017 like others?

aybe:
#1331

aybe:
Version 10 can target Windows 7:
https://docs.microsoft.com/en-us/cpp/windows/how-to-use-the-windows-10-sdk-in-a-windows-desktop-application?view=vs-2019

In an ideal world, this solution should use this version and see all its projects upgraded accordingly.

#1339

aybe:
closing, refer to #1339 instead

PR #1331, opened 2019-11-07
Reverting to Win 8.1 SDK

aybe:
- going back to Win 8.1 SDK, since some project get [C2118](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2118?f1url=https%3A%2F%2Fmsdn.microsoft.com%2Fquery%2Fdev16.query%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(C2118)%26rd%3Dtrue%26f%3D255%26MSPPError%3D-2147217396&view=vs-2019) with Win 10 SDK
- toolset is 142, VS2017 users are likely to upgrade to 2019, if that's a problem we can always downgrade

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -391,3 +391,4 @@ vs2015/libpng/linux-build/*
 /vs2015/libpdcurses/wincon/ozdemo.exe
 /vs2015/libpdcurses/wincon/firework.exe
 .vscode/settings.json
+vs2015/freetype/objs/freetype.lib
--- a/README.Windows
+++ b/README.Windows
@@ -5,10 +5,6 @@ Additional notes
 
 11/7/2019
 
-Fixups for Visual Studio 2019:
-
-Make sure you install MSVC v141 - VS 2017 C++ x64/x86 build tools (v14.16).
-
 Windows 8.1 SDK option is gone from the installer but can be found here:
 https://developer.microsoft.com/en-us/windows/downloads/sdk-archive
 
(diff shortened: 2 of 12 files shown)


Allofich:
> going back to Win 8.1 SDK, since some project get C2118 with Win 10 SDK

Did you have this problem? DOSBox-X was building without issue using AppVeyor's Visual Studio 2017 image, but it fails to build after this PR. I also can build DOSBox-X locally (or on AppVeyor) using Visual Studio 2019 with 10.0.18362.0, the latest version of the Windows 10 SDK in the Visual Studio Installer.

aybe:
I did,

Now after getting a little more involved with code base I found #1335 to occurs way more times than acceptable. I did not remember the solution behaving like this back then, I've been trying to find the cause but failed.

There have been so many changes that it's hard to tell what's causing this, currently I am thinking redoing this from scratch but it's a bit extreme you'd agree.

For the candidate there's one gotcha, since there are non +VS2017 projects, updating projects properties' is quite subtle (different defaults in versions resulting in different changes in files, tricky if you don't know the subtelties between versions (I do not)).

aybe:
Reopening, labelled -> help wanted.

aybe:
Also, as @joncampbell123 mentioned in #1326:
- there really is not point to need two SDKs
- Win7 target shall still work

I don't really mind which SDK is used unless these are honored and the solution builds properly.

joncampbell123:
The vs2013 project files were contributed by someone else who doesn't want to use VS2015 or higher. I don't maintain those.

If nobody is using those anymore I would like to clear them from the source code.

aybe:
I agree,

I would say the same for vcxproj projects, versions should be consolidated for clarity.

Issue #1322, opened 2019-11-06
Setting [dosbox] memsize from Configuration GUI does not get saved


aybe:
**To Reproduce**
1. delete dosbox.conf
2. start dosbox
3. open Configuration GUI
4. set memsize=32 in [dosbox]
5. save config
6. restart dosbox
7. memsize is still 16

**Environment:**
Microsoft Windows [Version 10.0.18362.388]
0.82.23
dosbox-x-windows-20191031-191254-windows.zip

PR #1332, opened 2019-11-07
Prop_int didn't set modification flag.

aybe:
# Description

memsize is now saved when changed from configuration GUI

#1322

no

probably no

**Additional information**


aybe:
I guess that was the error ...

Now go figure why there are 43 spaces being put in the config ... X)

notepad++_2019-11-07_07-34-05

joncampbell123:
The spaces inserted are based on ALL properties for that section, even if they're not printed.

aybe:
I think that'll do for now 😃 .

Issue #1319, opened 2019-11-06
dosbox.conf won't be loaded from taskbar unless it's pinned


aybe:
**Describe the bug**

If you try to spawn a new instance and program is not pinned, dosbox.conf won't be loaded. Checking with Process Explorer and the current directory is SYSTEM32 which is wrong.

**To Reproduce**

(make sure dosbox is not pinned in taskbar)

1. open dosbox from Explorer
2. dosbox.conf is loaded properly
3. flick dosbox up in taskbar and spawn new instance
4. dosbox.conf won't be loaded (check current directory, it's SYSTEM32)

**Expected behavior**

DOSBox should infer its current directory and not rely on system so dosbox.conf can load.

**Screenshots**

![notepad++_2019-11-06_23-10-12](https://user-images.githubusercontent.com/1537591/68342517-13ef8400-00eb-11ea-8d5d-91449eee5fe3.png)

![PROCEXP64_2019-11-06_23-15-39](https://user-images.githubusercontent.com/1537591/68342653-5fa22d80-00eb-11ea-9a93-fe98874b2581.png)

![PROCEXP64_2019-11-06_23-15-57](https://user-images.githubusercontent.com/1537591/68342661-6630a500-00eb-11ea-8658-f39a2c34e138.png)

**Environment (please complete the following information):**
 Microsoft Windows [Version 10.0.18362.388]
0.82.23
dosbox.conf

aybe:
@joncampbell123 

How is dosbox supposed to behave when its working directory is not where it resides (and possibly a non-writeable place such as SYSTEM32) ? Set its current directory as where it resides ?

JaegerFox:
that is a windows error, try to open epsxe or some other program using that method and you will see the same thing happen

PR #1345, opened 2019-11-12
Ensure local dosbox.conf is used when current directory is unexpected.

aybe:
## Merge #1351 before

# Description

Local dosbox.conf can be ignored when started with a weird current directory such as SYSTEM32.

#1319

**Additional information**

Works but not sure if it was the best place to put it on.


yksoft1:
Consider another way of using GetFullPathName() as #pragma comment(lib) won't work in MinGW.
By the way, what would this cause when dosbox-x.exe is running in Windows system32 directory in Windows systems with no VirtualStore redirect feature such as Windows XP/2003 or ReactOS?

aybe:
OK, isn't the minimum version Win7?

Le 12 nov. 2019 10:03, yksoft1 <notifications@github.com> a écrit :

Consider another way of using GetFullPathName() as #pragma comment(lib) won't work in MinGW.
By the way, what would this cause when dosbox-x.exe is running in Windows system32 directory in Windows systems with no VirtualStore redirect feature such as Windows XP/2003 or ReactOS?

yksoft1:
If you use my nmakefile (/ref-nmake/) in VS2017 command prompt to build, the executable should work  in XP. Don't know whether VS2019's CRT still support XP.

aybe:
Currently I have moved all projects to VS19 (so I'm getting is not a valid Win32 app in XP) but I think it should be possible to downgrade to the old toolset targetting XP, yet keep a good dev experience on Windows.

When this is done, I'll be able to test, I think with some extra checks it should be possible to accomodate the different scenarios.

aybe:
XP support is back -> #1351

aybe:
@yksoft1
Not sure I understand what you meant ... please explain. If dosbox EXE was effectively in SYSTEM32 then this is virtually a no-op, tested in XP and it just works as usual.

yksoft1:
DOSBox-X, before this commit, always wanted to load dosbox.conf not from EXE directory but from the current directory. I don't know if this was intended or a mistake.

joncampbell123:
I prefer that it load from the current directory, or at least load from the current directory first before checking alongside the EXE

aybe:
Okay, my approach is wrong,

I close, if problem surfaces again someone else will probably solve it with a better fix.

PR #1348, opened 2019-11-14
Completed fixes for solution.

aybe:
I did it while still hot, it would probably never have been otherwise.

- reset optimizations to defaults
- removed values that were defaults but marked as dirty
- preprocessor and directories re-ordered so differences are easier to spot
- ensure only Win32 gets SSE, the only platform it is supported on
- using Program Database everywhere, so no edit and continue, doesn't work in dosbox-x anyway
- warnings all to L3 except when higher (freetype), as a consequence there might be some new ones
- using default build output options, simpler output -> to enable only for diagnosis
- kept things that made sense, removed (resetted to default) things that didn't make sense (IMO)
- multi-threaded compilation definitely speeds things though more could done (PCH, etc)

There are probably places where some include directories could be removed but I didn't bother.

As a consequence of defaults, release builds might not produce the fastest code, to be adjusted.

Older settings were an absolute mess, in many places they didn't match sibling builds.

Many removals, little additions, some useless re-additions because of how it's done by VS.

`========== Build: 0 succeeded, 0 failed, 120 up-to-date, 0 skipped ==========`

for me you can merge, chapter closed!

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -36,7 +36,7 @@
   - Fixed all hard disk images created by IMGMAKE having
     VHD footers. (Allofich)
   - Windows Visual Studio solution: (aybe)
-    - cleaned and clarified project properties for dosbox-x
+    - cleaned and clarified project properties for projects
     - removed configurations that didn't make sense
     - fixed broken debugging
     - consolidated to use single versions of SDK and IDE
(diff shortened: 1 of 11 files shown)


joncampbell123:
x64 uses SSE too, in fact Win64 is supposedly made to assume SSE is present.

aybe:
done: #1349

PR #1349, opened 2019-11-14
64-bit SSE

aybe:
Visual Studio says:

> cl : command line warning D9002: ignoring unknown option '/arch:SSE'

...


Allofich:
What was the goal of this PR? It made the "cl : command line warning D9002: ignoring unknown option '/arch:SSE'" warning start appearing, right? And I found other projects where people are instead removing /arch:SSE or /arch:SSE2 to address this warning, as these options are apparently not available for x64 builds.

Listed at the following URL as x86 options:
https://docs.microsoft.com/en-us/cpp/build/reference/arch-x86?view=vs-2019

Not listed at the following URL as x64 options:
https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=vs-2019

Allofich:
I suggest this be reverted, as I don't think it does anything except create warnings in the x64 compiler output. If combined with https://github.com/joncampbell123/dosbox-x/pull/1357, Visual Studio 2019 will have clean compiler output again.

Issue #1350, opened 2019-11-14
MOUNT C . syntax does not work in XP


aybe:
![VirtualBoxVM_2019-11-14_05-56-04](https://user-images.githubusercontent.com/1537591/68827905-ec2c8d00-06a3-11ea-8673-d423255472d5.png)

Works in Win10:

![dosbox-x_2019-11-14_06-00-21](https://user-images.githubusercontent.com/1537591/68827942-11b99680-06a4-11ea-9584-58331bbb4a11.png)

aybe:
I have no idea on how to debug that ... remote VS2019 tools needs at least Win7.

joncampbell123:
I see Microsoft has still not yet fixed the bug with stat() and Windows XP.

For whatever reason, stat() does not work on Windows XP if compiled with VS2015 or later even if you target Windows XP.

This is why I didn't bother to target XP with the VS2015 project file.

If you want XP support, install MinGW and compile with the build-mingw scripts.

yksoft1:
Seems the VS2019 CRT officially dropped XP support once and for all.
VS2015 SP2 and VS2017 had a CRT with stat() working in XP.

aybe:
Now I remember there was such bug ...

joncampbell123:
@aybe I think it's better if VS2017 builds just ignore Windows XP and require Windows 7 or higher. Microsoft doesn't seem to care much about Windows XP anymore. Windows XP users can run the MinGW-built binaries instead.

PR #1351, opened 2019-11-14
Target XP ... from Windows

aybe:
Works as you can see but creates another issue -> #1350 

VirtualBoxVM_2019-11-14_05-56-04

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -42,6 +42,7 @@
     - consolidated to use single versions of SDK and IDE
     - enabled faster builds (multi-processor compilation)
     - fixed broken builds, e.g. ARM, ARM64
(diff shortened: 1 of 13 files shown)


joncampbell123:
dxguid.lib will be needed for Direct3D9 support, which is popular with the current userbase as far as I know, removing it to appease Windows XP when stat() still isn't fixed seems foolish.

Enabling DPI for a very specific version of the toolset is foolish, use a >= sign instead.

https://github.com/joncampbell123/dosbox-x/pull/1351/commits/d8329c2b1b3758e05a8654689c7b2fd42cf7bad9

aybe:
I understand,

I think I will just follow your idea of using MinGW for XP:
https://github.com/joncampbell123/dosbox-x/issues/1350#issuecomment-553790725

Issue #1358, opened 2019-11-28
Switch development to VS2019 instead of VS2017.


joncampbell123:
As stated before, MinGW is explicitly supported as a means to compile binaries compatible with Windows XP.
Microsoft has marked the Windows XP compatible toolset as deprecated. It's foolish to waste time supporting something they will drop soon.

Allofich:
The VS 2017 solution is fully switched over to VS 2019 now, isn't it?

Also, the VS 2013 solution looks like it was last touched 2 years ago and maybe should be removed.

PR #1368, opened 2019-12-22
Update VS2019 solution

Allofich:
# Description

1. Update the Visual Studio 2019 solution to automatically use the latest Windows 10 SDK available.

Reason: I have  Windows 10 SDK 10.0.18362.0 installed, which was the default when installing the C++ pack. I don't have 10.0.17763.0, which is what is specified in master branch currently. Setting the version to "10.0" lets it work with either, as it will automatically use the latest Windows 10 SDK that is installed.

2. Remove /arch:SSE option for x64 builds.

Reason: This option has no effect for x64 Visual Studio builds. All it does is create warnings in the build output that it is being ignored.

Visual Studio's help text says:
"Currently /arch:SSE and /arch:SSE2 are only available when building for the x86 architecture."

It also says "If no option is specified, the compiler will use instructions found on processors that support SSE2." With this PR, no option is specified, so this should be the behavior.

3. Update appveyor.yml

Reason: It needs to be updated to use the Visual Studio 2019 image, as currently it still uses the Visual Studio 2017 image and fails. I also removed some restrictions on when to build (building only for master branch, and only for commits that affect the /src or /vs2015 directories), as these just got in the way for me when using AppVeyor.

I also wanted to rename the "vs2015" folder to "vs2019", but I didn't because doing so will cause the git history to no longer be shown in GitHub for everything within the folder.

https://github.com/joncampbell123/dosbox-x/issues/1358

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,21 +1,8 @@
 #---------------------------------#
-#      general configuration      #
-#---------------------------------#
-
-branches:
-  only:
-    - master
-
-only_commits:
-  files:
-    - src\
-    - vs2015\
-
-#---------------------------------#
 #    environment configuration    #
 #---------------------------------#
 
-image: Visual Studio 2017
+image: Visual Studio 2019
 
 init:
   - git config --global core.autocrlf true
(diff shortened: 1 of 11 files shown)


joncampbell123:
Thank you for not renaming vs2015, a lot of Mac OS X and Linux build scripts assume that directory name too.

Issue #1019, opened 2019-02-23
Dynamic Core crashes in Windows RT


thchi12:
When I set dynamic core on my surface rt with RT 8.1 dosbox stops working.

joncampbell123:
@driver1998 Can you assist with this? I do not have a Windows RT system to test or develop on.

joncampbell123:
The dynamic core for arm seems best suited to the Raspberry Pi as far as I can tell. I don't believe it was adapted for the Windows RT ARM environment at all.

driver1998:
I don't, I only have a Raspberry Pi with Windows 10.
It seems like it is either an OS issue (FlushInstructionCache not working in RT) or a processor issue...

thchi12:
This works on a Lumia 950 XL with Win10 ARM64 Build 17134 installed. So maybe sth is not compatible with RT...

thchi12:
Tested on RPi with Win10 ARM32 PE Build 16299 and dynamic core works.
So it should be an operating system issue...

driver1998:
It might be related to this:
https://github.com/hrydgard/ppsspp/pull/9479#issuecomment-288364851
It is fixed in later version of Windows (at least 17134+, from my and thchi's testing) But for earlier version of Windows (RT is 8.0/8.1), it might be a no-go.

driver1998:
From some recent tests with tichi12, FlushInstructionCache is working on Windows RT.
But RT is ARMv7 Thumb only, which means it may need a specific dynarec target.
I tested the current `risc_armv4le-thumb.h`, `risc_armv4le-thumb-iw.h` and `risc_armv4le-thumb-niw.h` targets, and none of them work.
I don't know how to debug what exact instruction failed though...

PR #1419, opened 2020-02-25
Fix hanging on Windows RT, block dynarec on RT 8.x

### LINKED ISSUE #1418: Hang on startup on Windows RT/Windows 10 ARM32
**Describe the bug**
It hangs on startup, leaving a blank screen with a cursor. 
Sometimes the welcome screen does show up, but hangs shortly afterwards.
The following log show repeatedly in the console:
```
CPU: Illegal/Unhandled Interrupt Called 6
```

**To Reproduce**
Just start dosbox-x.

**Expected behavior**
It should run as normal.

**Screenshots**

 - Operating system
Windows RT, or Windows 10 ARM, running the 32-bit ARM build
 - DOSBox-X release version or commit SHA1 this issue occurred on
v0.82.19+ (It does not hang on Debug build, or when optimization is off in Release build)
 - Used configuration, e.g. ```dosbox.conf```
all defaults

**Additional context**

driver1998:
# Description

Fixes #1418 

Fixes hanging on startup on Windows ARM32, introduced from https://github.com/joncampbell123/dosbox-x/commit/2ff361fed30368b085a0797dd9aaa7e666d1a760.
Blocks access to dynarec on WinRT 8.x, until it is working properly.

No.

**Additional information**
Dynarec on Windows ARM32 works, but only on Windows 10.
Windows RT 8.x only runs ARMv7 Thumb-2 code, which we don't have a backend for.
Attempting to use dynarec on RT will result in an instant crash.

Related #1019

### DIFF
--- a/include/regs.h
+++ b/include/regs.h
@@ -62,7 +62,7 @@ struct Segment {
 enum SegNames { es=0,cs,ss,ds,fs,gs};
 
 struct Segments {
-	Bit16u val[8];
+	Bitu val[8];
 	PhysPt phys[8];
 	PhysPt limit[8];
 	bool expanddown[8];
--- a/src/cpu/core_dynrec.cpp
+++ b/src/cpu/core_dynrec.cpp
@@ -169,6 +169,33 @@ static struct {
 
 #include "core_dynrec/decoder.h"
 
+/* Dynarec on Windows ARM32 works, but only on Windows 10.
+ * Windows RT 8.x only runs ARMv7 Thumb-2 code, which we don't have a backend for
+ * Attempting to use dynarec on RT will result in an instant crash. 
+ * So we need to detect the Windows version before enabling dynarec. */
+#if defined(WIN32) && defined(_M_ARM)
+bool IsWin10OrGreater() {
+    HMODULE handle = GetModuleHandleW(L"ntdll.dll");
+    if (handle) {
+        typedef LONG(WINAPI* RtlGetVersionPtr)(RTL_OSVERSIONINFOW*);
+        RtlGetVersionPtr getVersionPtr = (RtlGetVersionPtr)GetProcAddress(handle, "RtlGetVersion");
+        if (getVersionPtr != NULL) {
+            RTL_OSVERSIONINFOW info = { 0 };
+            info.dwOSVersionInfoSize = sizeof(info);
+            if (getVersionPtr(&info) == 0) { /* STATUS_SUCCESS == 0 */
+                if (info.dwMajorVersion >= 10)
+                    return true;
+            }
+        }
+    }
+    return false;
+}
+
+static bool is_win10 = false;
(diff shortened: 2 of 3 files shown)


joncampbell123:
Wouldn't it be better to detect which ARM CPU it's running and the mode on rather than guess from the Windows version?

Though I'm not fully familiar with ARM, my understanding is that Thumb mode is enabled when the LSB of the instruction pointer is set at least after a branch instruction. I'm not sure if that's reflected in the instruction pointer but it might be worth it to check. Or, since the compiler is generating Thumb instructions, perhaps there's an #ifdef to disable dynarec if compiling to Thumb mode.

joncampbell123:
Second question is why changing val[8] in the segment structure to Bitu helps?

driver1998:
For the first question: The compiler always compiles in Thumb mode. Before Windows 10, running non-Thumb mode code is not supported at all.

> "ARMNT" is the PE image type of ARMv7-Thumb2. Visual C++ uses the #define _M_ARM_NT, though _M_ARM is also #defined by the compiler. Windows RT never leaves Thumb2 mode. In fact, the context switch routines in the kernel set the Thumb bit of CPSR when saving the CPU context of a frozen task, meaning that any interrupt will cause an ARM program to crash when control returns.
> 
> Myriachan, from https://forum.xda-developers.com/showthread.php?t=2503498

The ban is lifted on Windows 10, probably a byproduct from supporting running Android apps on Windows 10 Mobile. But the compiler is still restricted in Thumb mode.

Overall running non-Thumb code in Windows is sort of a hack all along, maybe disabling it all together is a better way to go...

For the second question... I actually have no idea. I checked through the commits between v0.82.18 and v0.82.19 and found that this commit breaks it.

joncampbell123:
Works for me...
If I were to purchase a Microsoft Surface to develop DOSBox-X and enable developer mode, would it allow me to develop Win32 non-UWP applications? Or is it UWP only and jailbreaking required?

driver1998:
If you are talking about Surfaces with Windows RT (1 and 2), then yes, jailbreaking is required.
If you are talking about the new ARM64 tablets, like Surface Pro X, then Win32 is open regardless of developer mode. Developer mode is required for UWP development only.

Also if you have existing ARM SBC and boards, you might be able to spin up a KVM VM and run Windows 10 there.

joncampbell123:
I just checked out a local store, and apparently MOST of the Surface X tablets are Intel x64 based. Only one of the display models had a ARM64 aka "Microsoft SQ1" CPU.

joncampbell123:
Picked up an ARM64-based Surface. ARM and ARM64 builds run fine. However the window is freaking tiny on the high resolution display, so perhaps if Direct3D9/OpenGL output is enabled it might be wise to read screen DPI and pick a more appropriate default size.

PR #1432, opened 2020-03-04
Don't write all MSCDEX log messages as errors

Allofich:
# Description

Clean up of DOSBox-X's MSCDEX log messages to separate between errors and non-errors rather than logging them all as errors.

Avoids confusion. I saw a MSCDEX INT 2F call being logged with ERROR and thought that meant the call was unimplemented, but that wasn't the case. 

It was because this line

`MSCDEX_LOG("MSCDEX: INT 2F AX=%04X BX=%04X CX=%04X",reg_ax,reg_bx,reg_cx);`

used `MSCDEX_LOG`, which was defined to log at level `ERROR`. I made it instead log at `NORMAL` and substituted in a new macro `MSCDEX_LOG_ERROR` for the places where log level `ERROR` is appropriate.

No.

### DIFF
--- a/src/dos/dos_mscdex.cpp
+++ b/src/dos/dos_mscdex.cpp
@@ -30,8 +30,8 @@
 
 #include "cdrom.h"
 
-#define MSCDEX_LOG LOG(LOG_MISC,LOG_ERROR)
-//#define MSCDEX_LOG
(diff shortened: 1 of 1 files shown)


Allofich:
Changed the log level of `MSCDEX_LOG` to `DEBUG`, as I think that may match better with usage in other parts of the program.

Issue #1227, opened 2019-08-29
Microsoft FORMAT.COM crashes after Y/N prompt


joncampbell123:
**Describe the bug**
When running FORMAT.COM in the native DOSBox-X environment with a mounted disk image, FORMAT.COM will crash and jump into random data instead of formatting the disk image.

Wengier:
I have fixed the crash in pull request #1491. Although it still won't actually format the disk (because the required low-level function is not yet implemented), it is now able to properly exit to DOS with an error message.

Wengier:
@joncampbell123 This one can be closed too since FORMAT.COM already works properly now.

PR #1491, opened 2020-04-26
Fixing total size report for mounted CD-ROM drives

Wengier:
When MOUNTing CD-ROM drives in DOSBox-X, by default the free disk size is always zero, but the total disk size reported may be incorrect. It is now fixed in this commit. The total disk size for mounted floppy or hard drives generally does not have this problem.

Also update the help message of the MOUNT command to mention its "-t cdrom" usage for mounting as CD-ROM drives (although you can also mount a CD-ROM drive on the host system as a standard local drive, which is the default).

P.S. This is my first time to create a pull request. Seems to be simpler than making diff files as done previously.


Wengier:
I also fixed the crash with FORMAT.COM as mentioned in Issue #1227 earlier. Although it still won't actually format the disk (because the required low-level function is not yet implemented), it is now be able to properly exit to DOS with an error message.

PR #1493, opened 2020-04-27
Fix PasteClipboard limitation in SDL2 build and COPY command crash

Wengier:
I have recently implemented the PasteClipboard feature for the SDL2 build, but as I mentioned previously it required KEYBUF.COM in order to paste long contents from the Windows clipboard. In a private email @emendelson sent me earlier he told me he still hoped this limitation to be removed someday just like the SDL1 build so that he can move his project to SDL2, so here it is. The fix works by combining the original code I wrote and some existing code written for SDL1 build, but completely bypassing the SDL1 keyboard handling code. It only fixes the feature for the SDL2 build, whereas everything in SDL1 build stills works the same as before.

By the way, due to this fix the clip_paste_speed config option that was originally written for SDL1 build now works in SDL2 build too. So "clip_paste_speed" now applies to both SDL1 and SDL2 builds in Windows.

I have also fixed the COPY command crash that happened in certain cases. This occurred in both 0.83.0 and 0.83.1-git. When it happens DOSBox-X quits completely with no warnings. It is now fixed in this pull request.

In addition, I fixed MS-DOS 7+ DEL command with wildcards not working in non-Windows platforms. Previously the code handling this part was only for Windows (within #if defined (WIN32)), but it is now tested in Linux too.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -98,12 +98,10 @@
     change the keyboard modifier such as none, alt, ctrl,
     shift, or disable this feature (default). (Wengier)
   - Updated the PasteClipboard feature in SDL1 build
-    to support Unicode text translations. (Wengier)
-  - Fixed the PasteClipboard feature in SDL2 build;
-    a keyboard buffer expansion program like keybuf.com
-    (e.g. "KEYBUF 1024") is recommended to be loaded
-    (just once) before pasting long contents from the
-    Windows clipboard in SDL2 build. (Wengier)
+    to support Unicode text translations; also fixed this
+    feature not working in SDL2 build. The config option
+    "clip_paste_speed" is added to speed up or slow down
+    the pasting speed for different DOS programs (Wengier)
   - Support for DOSLIB2's w95sysrg/w95sysrs utils to
     get/set system registry location in DOS 7+. (Wengier)
   - Long filename support added, adapted from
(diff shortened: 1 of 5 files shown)


joncampbell123:
I'll accept this for now but long term I think it would be better for DOSBox-X itself to handle LFN wildcard matching so that it's more consistent emulation and not dependent on any quirks of the host system or it's APIs.

What does the 0x1000 value mean for fFlags and SHFileOperation()?

Wengier:
@joncampbell123 I fully agree with your point. I was actually thinking about this too. The Windows-only code in drive_local.cpp added previously was directly copied from the vDosPlus code, and since vDosPlus is Windows *only* and it *only* supports local drives, this code worked very well in the past for that project. But since DOSBox-X is cross platform and supports more than local drives, I agree it is better to place the relevant code in dos_files.cpp instead in the long term.

As for the 0x1000 value, I added it a few years ago to the vDosPlus code to solve a certain problem, but right now I do not think I can still remember its exact meaning. Perhaps I should put a comment when I added it at that time.

Wengier:
@joncampbell123 Now I remember it, 0x1000 means FOF_NORECURSION. I changed it to 0x1000 back then because FOF_NORECURSION did not compile in the earlier VS version I was using to build vDosPlus at that time, although it should build fine in the current version of VS.

joncampbell123:
Ah. Usually when I deal with compilers that lack newer #defines, I just copy the define into the source file with an #ifdef around it to not define it if already defined.

    #ifndef FOOBAR
    #define FOOBAR 1234
    #endif

Wengier:
@joncampbell123 That is a good way to solve such problems I think. I believe I can use this in the future too.

PR #1494, opened 2020-04-28
Fix display for RE-DOS command and add help infos

Wengier:
I found that when using the RE-DOS command to restart the kernel of the emulated DOS, the display of the welcome message box may have some color issues, so I fixed them in this pull request. Tested with the default svga_s3 as well as CGA and PC98 modes to make sure it displays correctly.

Also, I added help messages for the RE-DOS and COMMAND commands when you run them with /?.

In addition, for the Tab completion I made CHDIR, MD/MKDIR and RD/RMDIR commands directory-only just like the CD command.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,8 +1,4 @@
 0.83.1
-  - PC speaker: Mode 3 (square wave) and a counter value
-    of 1 produces a low frequency square wave. Ulrasonic
-    frequencies do not begin until count == 2. This fixes
-    helicopter noises in Paratrooper.
   - dosbox-x.conf is now recognized as the default config
     file name in addition to dosbox.conf (Wengier)
   - Tandy DAC output fixed to slowly ramp last sample to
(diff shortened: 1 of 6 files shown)


Wengier:
I also fixed the COPY error when there is no enough space left in the destination by adding a simple free disk space check before copying files. Without this check, COPY command can either result in a copy failure or it will appear succeed, but the target file is either completely empty or only partially copied.

joncampbell123:
Appreciate the work! Ready for me to merge?

Wengier:
@joncampbell123 I just finished cleanup the error messages for the SUBST command. So it is ready to be merged now.

Wengier:
It is ready to be merged now.

Wengier:
@joncampbell123 By the way, perhaps the sample dosbox.reference.conf file should be renamed to dosbox-x.reference.conf too to reflect that it is a DOSBox-X sample config file (instead of a DOSBox one).

joncampbell123:
Agreed. Merging now.

PR #1507, opened 2020-05-03
Code cleanup

Allofich:
Some code cleanup, using static analysis tools.

1. Using "const" for unmodified values
2. Using class initialization or constructor list initialization instead of assignment within constructor bodies
3. Removing pointless access specifiers
4. Other minor cleanup

### DIFF
--- a/include/shell.h
+++ b/include/shell.h
@@ -54,7 +54,7 @@ public:
 	BatchFile(DOS_Shell * host,char const* const resolved_name,char const* const entered_name, char const * const cmd_line);
 	virtual ~BatchFile();
 	virtual bool ReadLine(char * line);
-	bool Goto(char * where);
+	bool Goto(const char * where);
 	void Shift(void);
 	Bit16u file_handle;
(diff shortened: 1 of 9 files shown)


joncampbell123:
You could also modify src/dos/drive_fat.cpp to just use 0xFF & ~A & ~B and then you do not need the (Bit8u). A comment could be put there to say that at the time of the modification, DOS_DTA::SetupSearch's attribute parameter is type Bit8u, 8 bits wide.

joncampbell123:
Also code cannot find the utf8_encode/utf8_decode functions anymore because a function parameter was changed to const while the definition in dosbox.h was not updated to match (fixing now).

Allofich:
> Also code cannot find the utf8_encode/utf8_decode functions anymore because a function parameter was changed to const while the definition in dosbox.h was not updated to match (fixing now).

Ah, sorry, it compiled so I thought it was okay. For the future I'll be sure to check that declaration and definition match.

PR #1511, opened 2020-05-04
Fix Pause + Reset system, FluidSynth logging, and support mouse copy/paste in PC-98 mode

Wengier:
I noticed that in most builds of DOSBox-X there is a bug when you reset the guest system while it is currently paused. The internal DOS will be reset, but in the menu "Pause" remains checked even though it is no longer paused, and the "Pause" option is no longer usable in DOSBox-X. So I made a pull request to fix this. It also fixes the logging issue (and the resulting delay) related to FluidSynth support. In addition, the Windows clipboard copy & paste via the right mouse click feature now supports PC-98 mode too just like the regular mode.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 0.83.2
-  - The Windows clipboard copy & paste via the right mouse
-    button feature now supports PC-98 mode too. (Wengier)
+  - The copy & paste Windows clipboard text via the right
+    mouse button feature now supports PC-98 mode too just
+    like the standard mode. (Wengier)
   - Improved support for FluidSynth MIDI device by porting
     code from DOSBox ECE. Set "mididevice=fluidsynth" along
     with other required options such as "fluid.soundfont"
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.83.2
+  - The Windows clipboard copy & paste via the right mouse
+    button feature now supports PC-98 mode too. (Wengier)
   - Improved support for FluidSynth MIDI device by porting
     code from DOSBox ECE. Set "mididevice=fluidsynth" along
     with other required options such as "fluid.soundfont"
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,7 @@
 0.83.2
   - The copy & paste Windows clipboard text via the right
-    mouse button feature now supports PC-98 mode too just
-    like the standard mode. (Wengier)
+    mouse button feature now has limited support for PC-98
+    mode too in addition to other modes. (Wengier)
   - Improved support for FluidSynth MIDI device by porting
     code from DOSBox ECE. Set "mididevice=fluidsynth" along
     with other required options such as "fluid.soundfont"
--- a/src/gui/midi.cpp
+++ b/src/gui/midi.cpp
@@ -634,7 +634,7 @@ public:
 
 		midi.available=true;
 		midi.handler=handler;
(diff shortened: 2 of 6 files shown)


joncampbell123:
INT 10h is not used in PC-98 mode except in very limited circumstances as direct calls from the DOS CON driver.

Also, PC-98 does not store character data interleaved with attribute data. They are entirely separate regions of memory.

A0000-A1FFF is 16-bit per character data. One character occupies one cell if the high byte is zero, or two cells (the current and next) if high byte is nonzero (aka double wide).
A2000-A3FFF is 16-bit per attribute data, though only the low 8 bits exist.

Please revert commit https://github.com/joncampbell123/dosbox-x/pull/1511/commits/01f2caf134559816743607e8228434e73a97897f

joncampbell123:
Have you confirmed that text selection works in PC-98 mode by running something with machine=pc98?

Also be aware that double-wide characters occupy two character cells, and that DOSBox-X (and real hardware) ignore the contents of the second cell when it starts drawing the double-wide character. There is also a common convention in PC-98 games and PC-98 MS-DOS it seems to fill both cells with the same 16-bit character code anyway, and additional references to setting the most significant bit in one of them (don't recall which). That should be factored into the text selection code for PC-98 mode.

https://github.com/joncampbell123/dosbox-x/pull/1511/commits/48726247c6a53e87492a86a1fc618b97bd671de1

Wengier:
@joncampbell123 Sure, I have tested "machine=pc98" to make sure text selection (+copy & paste) by mouse now works properly in PC-98 mode with this code, at least for standard SBCS characters (and in the command line). Before this change it did not work at all, but now it works just fine as I specifically tested to make it work. Also, note that in int10_char.cpp there was already existing code for WriteChar with attribute data in PC-98 mode, so I simply added the ReadChar counterpart for it. You can verify it in this file yourself too.

Wengier:
@joncampbell123 Below is the existing code for PC-98 mode in int10_char.cpp (function WriteChar):

    case M_PC98:
        {
            // Compute the address  
            Bit16u address=((row*80)+col)*2;
            // Write the char 
            PhysPt where = CurMode->pstart+address;
            mem_writew(where,chr);
            if (useattr) {
                mem_writeb(where+0x2000,attr);
            }

            // some chars are double-wide and need to fill both cells.
            // however xx08h-xx0Bh are single-wide encodings such as the
            // single-wide box characters used by DOSBox-X's introductory
            // message.
            if ((chr & 0xFF00u) != 0u && (chr & 0xFCu) != 0x08u) {
                where += 2u;
                mem_writew(where,chr);
                if (useattr) {
                    mem_writeb(where+0x2000,attr);
                }
            }
        }
        return;

And the code I added for ReadChar is very similar to the above too, at least for standard SBCS characters.

joncampbell123:
My point is that the WriteChar call is a hack in PC-98 mode to avoid making too many drastic changes to the DOS CON driver back when developing PC-98 mode.

There's no point in the ReadChar function, because nothing will call it.

Wengier:
@joncampbell123 OK, so I have removed code for PC-98 mode in the ReadChar function. Instead, the text selection by mouse will directly read characters on the screen as written by WriteChar. I verified it works with "machine=pc98". I think the feature's support for PC-98 mode is not complete because of what you just stated, but I think it is still better than nothing as previously it did not work at all. As a result I modified the CHANGELOG to state that the support for PC-98 is limited. Maybe I will implement full support for PC-98 mode in the future then.

joncampbell123:
It's fine to read text video memory directly. There's no mechanism that I'm aware of that allows PC-98 to relocate the VRAM behind it anywhere else other than A0000-A1FFF.

PR #1513, opened 2020-05-05
New fixes

Allofich:
Continuation of https://github.com/joncampbell123/dosbox-x/pull/1510. Builds on Visual Studio and GCC.


Allofich:
The disabling of the bit-op test doesn't seem to be necessary anymore. It didn't cause a problem when building with current versions of Visual Studio 2017 or 2019. (If it's a problem it would cause an error during compiling, right?)

joncampbell123:
That was done for some particular version of VS2017 if I recall that suddenly didn't like the bitop tests. If it compiles again, that's fine.

Issue #1515, opened 2020-05-06
localDrive::GetFileAttr is not converting guest filename code page to host on Win32


yksoft1:
I saw that on Win32, localDrive::GetFileAttr is directly using the guest filename to query the file's native attributes using GetFileAttributes() without doing guest to host filename code page conversion.

In src/dos/drive_local.cpp:

```
#if defined (WIN32)
	Bitu attribs = GetFileAttributes(newname);
	if (attribs == INVALID_FILE_ATTRIBUTES)
		{
		DOS_SetError((Bit16u)GetLastError());
		return false;
		}
	*attr = attribs&0x3f;
	return true;
#else
    // guest to host code page translation

```

This may make ATTRIB command fail to get attributes for files with Japanese filenames using "drive_local" filesystem under PC-98 mode.

![image](https://user-images.githubusercontent.com/37106774/81140952-b48d0680-8f9d-11ea-86b9-00b9f875beae.png)

yksoft1:
BTW does anyone tested the new LFN features under PC-98 mode at all?

joncampbell123:
@yksoft1 Those are brand new and not fully tested yet on my end.
@Wengier The LFN updates are your work, correct? For the WIN32 case, use the filename conversion code to convert from guest (in this case Shift-JIS) to UTF-16 (wchar_t) for the Windows API. You'll have to change GetFileAttributes to GetFileAttributesW to pass the widechar string properly. Do you need assistance on how to do that?

joncampbell123:
@Wengier I'll get around to updating guest to host unicode conversion for get/set file attributes on Windows.

Wengier:
@joncampbell123 I will make a pull request very soon. You can check that.

Wengier:
@joncampbell123 I have submitted my pull request (#1518). It improved Shift-JS support in PC-98 mode (GetFileAttr, SetFileAttr, LFN DIR), as well as adding the -set command-line option to change config setting when starting DOSBox-X.

yksoft1:
`    // TODO: offer a config.h option to opt out of Windows widechar functions
`

I think it's necessary at least for non-VS2019 builds as someone might try to run them under Windows platforms without widechar support (like Windows 9x and HX-DOS extender).

Wengier:
@yksoft1 In any case, with the latest code the LFN features should alreadylyalready work normally in the PC-98 mode. Thanks for bring us attention about this.

yksoft1:
[...]

PR #1518, opened 2020-05-06
Added -set command-line option to set config option(s), improved PC-98 mode 

Wengier:
There is currently no way to specify config option(s) when starting DOSBox-X, except by putting them in the config file. To overcome this I decided to add the -set command-line option to do achieve this. Like the -c command-line option, the -set command-line can be specified multiple times for multiple options, overriding any options in the config file. For example, the command:
 
``dosbox-x.exe -set machine=pc98``

Will start DOSBox-X in PC-98 mode regardless of the config setting in dosbox-x.conf.

``dosbox-x.exe -set machine=cga -set fullscreen=true``

Will start DOSBox-X in CGA mode and also in full-screen mode.

The usage of the -set command-line is the same as "config -set", so you can specify a section or property with spaces in quotes. For example:

``dosbox-x.exe -set "dos dos clipboard device enable=true"``

Will enable support for the CLIP$ device.

I also improved Shift-JIS support in PC-98 mode as mentioned in Issue #1515, including GetFileAttr, SetFileAttr, and DIR command in PC-98 LFN mode.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,9 @@
 0.83.2
+  - Added -set command-line option to change config options.
+    It can be specified multiple times for multiple options,
+    overridding any options in the config file. For example,
+    the command "dosbox-x.exe -set machine=pc98" will force
+    DOSBox-X to start in PC-98 mode. (Wengier)
   - Re-added full drive menu items for the Windows platform.
     The "Boot from drive" item (A:, C: and D: drives only)
     should work in other platforms too. The BOOT command is
@@ -9,6 +14,7 @@
     and EGA/VGA detection failure with SuperCalc 3 and
     SuperCalc 4. This option, enabled by default, can be
     disabled or enabled from dosbox.conf.
+  - Improved long filename support for PC-98 mode. (Wengier)
   - The copy & paste Windows clipboard text via the right
     mouse button feature now has limited support for PC-98
     mode too in addition to other modes. (Wengier)
--- a/include/control.h
+++ b/include/control.h
@@ -121,6 +121,7 @@ public:
     std::string opt_editconf,opt_opensaves,opt_opencaptures,opt_lang;
     std::vector<std::string> config_file_list;
     std::vector<std::string> opt_c;
+    std::vector<std::string> opt_set;
 
     bool opt_disable_dpi_awareness;
     bool opt_disable_numlock_check;
--- a/src/dos/dos_programs.cpp
+++ b/src/dos/dos_programs.cpp
@@ -466,33 +466,29 @@ search:
 
 		_splitpath (path, drive2, dir, fname, ext);
 
-		if((!strcmp(ext,".ima")) || (!strcmp(ext,".img")) || (!strcmp(ext,".iso")) || (!strcmp(ext,".cue")) || (!strcmp(ext,".bin")) || (!strcmp(ext,".mdf"))) {
-			char type[15];
-			if(!strcmp(ext,".img"))
-				strcpy(type,"");
-			else if(!strcmp(ext,".ima"))
-				strcpy(type,"-t floppy ");
-			else
-				strcpy(type,"-t iso ");
-			char mountstring[DOS_PATHLENGTH+CROSS_LEN+20];
-			strcpy(mountstring,"IMGMOUNT ");
-			strcat(mountstring,type);
-			char temp_str[4] = { 0,0,0 };
-			temp_str[0]=' ';
-			temp_str[1]=drive;
-			temp_str[2]=' ';
-			strcat(mountstring,temp_str);
-			strcat(mountstring,path);
-			strcat(mountstring," >nul");
-			DOS_Shell temp;
-			temp.ParseLine(mountstring);
-			if (!Drives[drive-'A']) {
-				drive_warn="Drive "+str+": was not mounted successfully.";
-				MessageBox(GetHWND(),drive_warn.c_str(),"Error",MB_OK);
-				return;
-			}
-		} else
-			LOG_MSG("GUI: Unsupported filename extension.");
+		char type[15];
+		if(!strcasecmp(ext,".ima"))
+			strcpy(type,"-t floppy ");
+		else if((!strcasecmp(ext,".iso")) || (!strcasecmp(ext,".cue")) || (!strcasecmp(ext,".bin")) || (!strcasecmp(ext,".mdf")))
+			strcpy(type,"-t iso ");
+		else
+			strcpy(type,"");
+		char mountstring[DOS_PATHLENGTH+CROSS_LEN+20];
+		strcpy(mountstring,"IMGMOUNT ");
+		strcat(mountstring,type);
+		char temp_str[3] = { 0,0,0 };
+		temp_str[0]=drive;
+		temp_str[1]=' ';
+		strcat(mountstring,temp_str);
+		strcat(mountstring,path);
+		strcat(mountstring," >nul");
+		DOS_Shell temp;
+		temp.ParseLine(mountstring);
+		if (!Drives[drive-'A']) {
+			drive_warn="Drive "+str+": failed to mounted.";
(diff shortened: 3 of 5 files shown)


joncampbell123:
The point of the host_cnv_char_t is to match the best datatype for the host system.

On Windows, host_cnv_char_t is wchar_t.
On anything else (Linux, Mac OS X) host_cnv_char_t is char to represent UTF-8.

CodePageGuestToHost() will convert the guest code page to wchar_t for Microsoft's widechar functions if compiled on Windows for that reason. The wrapper is unnecessary.

The guest <-> host conversion is supposed to be applied regardless of machine type (you check only for PC-98 mode). In IBM PC mode for example the code page conversion is to convert from CP437 (the DOS character set) to the host, and in PC-98 mode, Shift-JIS to the host.

Look at other functions in drive_local.cpp like MakeDir() for a good idea how to do this.

joncampbell123:
Mind if I do the GetFileAttribute/SetFileAttribute work instead? Remove the "Improve PC-98 mode" fixes and I'll take care of that part, and merge the rest of the pull request.

Wengier:
@joncampbell123 Of course I wanted to use CodePageGuestToHost() for all machine types in that function, but unfortunately CodePageGuestToHost() did not work well for that specific function, causing crashes if I do. That is why it took me some time to submit this pull request, and I had to add a CodePageHostToGuestL() help function for PC-98 mode too (otherwise it will crash too). I supposed there is some limitation in the CodePageGuestToHost() function for this case, so I have to reply on additional code to make it work reliably.

Wengier:
@joncampbell123 The GetFileAttribute/SetFileAttribute case was not PC-98 specific though. The check for PC98 mode was done for something else (DIR in PC-98 LFN mode).

Wengier:
@joncampbell123 Please check the code again. I am not checking for PC-98 mode for  GetFileAttribute/SetFileAttribute at all.

Wengier:
The GetFileAttribute/SetFileAttribute work was only a small part of the patch. For these I did use CodePageGuestToHost() completely without any wrapper. The wrapper is for something else (as I mentioned earlier, for DIR in PC-98 LFN mode).

joncampbell123:
Why the if (IS_PC98_ARCH) on line 1289? https://github.com/joncampbell123/dosbox-x/pull/1518/commits/a491457f9d745a328c8662cbd163a343b1f175f9#diff-d6c7e14d805837d969eb6f5c84972ae4R1289

And the other one farther down? https://github.com/joncampbell123/dosbox-x/commit/a491457f9d745a328c8662cbd163a343b1f175f9#diff-d6c7e14d805837d969eb6f5c84972ae4R1329

Wengier:
@joncampbell123 The two "if (IS_PC98_ARCH)" are for FindFirst/FindNext, not GetFileAttribute/SetFileAttribute. GetFileAttribute/SetFileAttribute only needs CodePageGuestToHost() without any wrapper. But the FindFirst/FindNext fix is to make DIR show Shift-JS in PC-98 LFN mode correctly. For FindFirst/FindNext I cannot completely reply on the CodePageGuestToHost() function (otherwise it will cause crash), that is why I am adding "if (IS_PC98_ARCH)" for them.

joncampbell123:
Hold on I get what's going on here. The CodePageHostToGuest() uses a global that's overwritten when called again, so the L version uses another global.

I'll go ahead and merge this but I also want to remove the if (IS_PC98_ARCH) because code page conversion going back should always use CodePageHostToGuest(). Currently for IBM PC mode, wcstombs() is used which converts to whatever non-unicode locale the user is using, NOT CP437 or Shift-JIS.

Wengier:
@joncampbell123 In that function there are two names to convert, both SFN and LFN names (unlike in all other functions there is only one name to convert). And unfortunately CodePageHostToGuest() does not work for both names at the same time (if you tried to do both it will cause crash). Perhaps CodePageHostToGuest() needs to be fixed to do this properly, but for now I have to rely on this additional code to make it work stably.

joncampbell123:
If you look at CodePageHostToGuest() and CodePageGuestToHost() farther up the source file, you'll see the convenience functions are just a C++ overload that calls the real function pointed at the cpcnv_temp global, which it then returns if success. The real functions accept a source and destination pointer. The dest pointer is assumed to be at lease CROSS_LEN bytes large.

Wengier:
@joncampbell123 I noticed that already of course. There are at least two problems I encountered, one with the overload function, and the other with the real function. So I solved the first problem by adding a CodePageHostToGuestL() helper function, and the second problem was worked around by using "if (IS_PC98_ARCH)", although the proper way to fix the second problem is to fix the real CodePageHostToGuest() function itself.

joncampbell123:
Does it show up as string corruption with an occasional segfault? Perhaps I need to revisit those functions in terms of the find first/next functions and LFNs to see if there are any problems I need to look for.

Wengier:
@joncampbell123 Yes, that is what I encountered when I tried to completely reply on CodePageHostToGuest() in that function in the beginning. The whole DOSBox-X window may disappear with a segmentation fault message.

joncampbell123:
I will test LFNs then, copying COMMAND.COM from a Windows 95 bootdisk now.

Wengier:
@joncampbell123 You can just use the built-in COPY command for testing too in PC-98 LFN mode.

joncampbell123:
command.com with ver=7.0 shows correct code page conversion for both SFNs and LFNs.

Wengier:
I wondered if the main segmentation fault message was already fixed by calling to CodePageHostToGuestL() instead of CodePageHostToGuest(). You can experiment with changing CodePageHostToGuestL() back to CodePageHostToGuest(), and you will soon notice the crash.

joncampbell123:
I'm noticing that Windows builds do not tolerate filenames with any non-ASCII characters.

One test I do for example is I type "EDIT ▒▒▒▒.TXT" (EDIT and then ALT+177, ALT+177, etc.) and see if the filename appears on the host as four gray blocks. Code 177 in code page 437 is the classic MS-DOS gray block code.

Windows builds turn it into something like _11DB6~1.TXT which is odd.

Wengier:
@joncampbell123 I believe names like _11DB6~1.TXT are the real 8.3 names generated on Windows system. They will be represented as SFNs in DOSBox-X as is. But the true names will be shown as LFNs in DOSBox-X.

joncampbell123:
I could probably fix this for Windows if I write the find first/find next Windows functions to ignore the alternate name if cFileName is short enough to fit into a 8.3 short name.

Wengier:
@joncampbell123 I noticed that the latest commit d6bc616 will cause some files with Shift-JS names to not show up with DIR. On the other hand, the previous commit fe258a9 works in this case.

joncampbell123:
I know... I'm trying to pin down why. I don't understand why it would make any difference.

joncampbell123:
Fixed it.

Wengier:
@joncampbell123 The new is_filename_8by3w function will filter out file names like ▒▒▒▒.TXT causing them to not show up in PC-98 mode. Also, it is recommended to change line 182 in cross.cpp:

``if (*fname == '.') fname++;``

To:

``if (*fname == L'.') fname++;``

So that MinGW will work properly (VS seems to work fine).

joncampbell123:
There's no code point in Shift-JIS for PC-98 to represent ▒▒▒▒.TXT.

If you look at the console or log file, you should see:

    LOG: read_directory_next: Filename '▒▒▒▒.TXT' from host is non-representable on the guest filesystem through code page conversion

Wengier:
If you disable the is_filename_8by3w function, that file will show up in the PC-98 mode though, with a name like _11DB6~1.TXT. Perhaps you can add a config option to use Windows 8.3 names whenever available?

Wengier:
I think I will modify the code so that if the file name contains characters not representable in the current code page, AND there is already a Windows 8.3 file name available, then always use this file name so that the file will show up in the list. This situation can apply to other code pages too, not just PC-98 mode, so this change should be useful in general too.

PR #1522, opened 2020-05-07
Fix VS2019 compilation

Allofich:
I found what has been causing recent versions of Visual Studio 2019 to error-out when building: the `method` enum in the `ScreenSizeInfo` class. Fixed by moving the enum out of the class.

### DIFF
--- a/include/sdlmain.h
+++ b/include/sdlmain.h
@@ -41,16 +41,16 @@ enum PRIORITY_LEVELS {
     PRIORITY_LEVEL_HIGHEST
 };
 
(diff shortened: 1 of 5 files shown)


joncampbell123:
So that's all it takes to make their linker rapidly consume 4GB of RAM? :laughing:

Allofich:
Guess so. I wonder what the cause is? At least the workaround was simple.

yxmline:
thanks
vs2019 16.5.4 build ok

PR #1524, opened 2020-05-08
Mouse copy clipboard to use Unicode text, supporting Kana characters in PC-98

Wengier:
I wanted the mouse clipboard copy feature to handle DBCS characters in PC-98 mode, so I improved the feature to use Unicode text translations (just like the clipboard paste feature). And by adding a mapping table, it is now able to copy Japanese Kana characters in PC-98 mode in addition to SBCS characters. I am not entirely sure if an existing mapping table or function exists for direct use in this case, so I added the mapping table in the code myself. It is currently only for Kana characters, not Kanji characters, but if there is already a mapping table or function including Kanji characters suitable for this use, then the mouse clipboard copy feature will easily support Kanji characters as well in PC-98 mode.

In addition, I improved the recent is_filename_8by3w function to always use existing Windows 8.3 names if the long file name contains characters that are not representable in the current code page, so for example file names like ▒▒▒▒.txt will show up in the PC-98 mode now too (with a name like  _11DB6~1.TXT). Also further improved MinGW compatibility for the function so that all files should now show up in PC-98 mode (or else some files will be still missing in the directory list in the MinGW build).

Tested with both MinGW 64-bit and VS2015 (SDL2) and confirm the above work.


Wengier:
@joncampbell123 Do you know the best way to convert a 16-bit (unshifted) JIS code (e.g. 0x2105 for "ァ") to Shift JIS code (e.g. 0x8340 for "ァ") besides using a (long) mapping table? It is not very easy to find relevant resource online.

Wengier:
@joncampbell123 I did make such a mapping table myself for Kana characters, but for Kanji characters it will be too much work.

joncampbell123:
A copy of the Unicode consortium's ShiftJIS to Unicode mapping is in mappings/CP932.TXT.

You could build a table from that file, converting SJIS to unshifted JIS as you go, to generate a lookup table from unshifted to unicode.

Know that when reading 16-bit values from video memory, double-wide characters should be ANDed with 0x7F7F because that's apparently how real hardware works, and there is a convention to mark one of the double-wide characters with bit 15 set.

rderooy:
There is an effort to add legacy computer systems to Unicode. Not sure if that is useful for the PC-98
http://www.unicode.org/L2/L2019/19025-terminals-prop.pdf

joncampbell123:
Shift to unshift conversion is very simple actually.

The formula to convert to Shift-JIS is here: https://en.wikipedia.org/wiki/Shift_JIS

joncampbell123:
And Shift-JIS to unshifted is implemented in src/misc/shiftjis.cpp by reversing those equations.

joncampbell123:
Also know that for some reason, the low byte of double-wide chars is stored counting from 0. You will need to add 0x20 to apply the Shift-JIS equation.

joncampbell123:
Also a table to convert Shift-JIS to unicode is in mappings/CP932.TXT, as obtained from the Unicode consortium's website.

Wengier:
@joncampbell123 Thanks for the information. I did look at the Wikipedia article before but I could not apply it directly so I did not use it previously. I will implement the unshifted to shifted conversion code with the new infomration. Everything else is already taken aware of by the code except for the unshifted to shifted conversion, which is the last thing needed for a full implementation.

Wengier:
@joncampbell123 I have already implemented the conversion so that both Kana and Kanji characters are now supported via mouse copy in PC-98 mode (in addition to SBCS characters). Please check the new pull request #1529.

PR #1528, opened 2020-05-08
SVN r4338

Allofich:
Noticed a new commit in SVN that can be easily dropped in.

https://sourceforge.net/p/dosbox/code-0/4338/


Allofich:
There's also a comment about this SVN commit at the dosbox-staging fork, where someone gives it approval.
https://github.com/dosbox-staging/dosbox-staging/commit/d289e6429104c8cec4e10fea99ce2e29a15f0b61

joncampbell123:
Ha, MSCDEX.EXE probably doesn't expect to handle sectors larger than 64KB, right? :)

Issue #1500, opened 2020-04-30
Reference config errors


rderooy:
**Describe the bug**
A few mistakes in the reference config.

```
[fdc, primary]
#         int13fakev86io: If set, and int13fakeio is set, certain INT 13h commands will
#                           cause floppy emulation to issue fake CPU I/O traps (GPF) in
#                           virtual 8086 mode and a fake IRQ signal. you must enable this option
#                           if you want 32-bit floppy access in Windows 95 to work with DOSBox-X.
```
Note how it mentions setting "int13fakeio", but that is not an option for [fdc, primary] (only valid for [ide, *])

```
[sblaster]
#                                              dma: The DMA number of the soundblaster. Set to -1 to start DOSBox-X with the IRQ unassigned
#                                                     Possible values: 1, 5, 0, 3, 6, 7.
#                                             hdma: The High DMA number of the soundblaster. Set to -1 to start DOSBox-X with the IRQ unassigned
#                                                     Possible values: 1, 5, 0, 3, 6, 7.
```
DMA settings, but then it talks about IRQ...

Wengier:
I agree that the reference config file has issues in the comments like this. It needs some time to do such cleanups, probably in the next a few releases.

ThePillenwerfer:
It also says in the [parallel] section:—
```
#                for printer:
#                  printer still has it's own configuration section above.
```
when there isn't such a thing.

rderooy:
Not a reference config issue, but...
![command_003](https://user-images.githubusercontent.com/17701098/81296840-cab8c680-9072-11ea-9513-6b3b8535b22a.png)

The ```-chs``` option works, but not ```-geometry```

rderooy:
In the reference config.

```# fluid.reverb,roomsize: Fluidsynth reverb room size.```

and 

```fluid.reverb,roomsize = .61```

Note how it uses a comma instead of a period, while all other settings use a period as a seperator.

Wengier:
I have already fixed most issues mentioned above for the config reference file in my pull request #1529, including the addition of the [printer] section. But note that "int13fakev86io" (instead of "int13fakeio") is accepted in [fdc, primary] section, so this one is not really an issue.

rderooy:
@Wengier if the "int13fakev86io" is a valid option for [fdc, primary], it should probably be mentioned in the reference config.

Also the screenshot I posted of imgmake is another issue where the "-geometry" text in the help simply needs to be changed to "-chs".

[...]

PR #1529, opened 2020-05-08
Support for Kanji via mouse copy in PC-98 mode, add [printer] section + other fixes in config file

Wengier:
By implementing the full Unshift JIS => Shift JIS conversion, the mouse copy feature now supports both Kanji and Kana characters in PC-98 mode, in addition to SBCS characters. Pasting both SBCS and DBCS characters via mouse in PC-98 mode works fine too of course.

Also, as mentioned by @rderooy, @ThePillenwerfer and myself in Issue #1500, there are several problems in the config reference file, including a missing [printer] section (it is already supported in the code but never mentioned in the config reference file). So I fixed issues already mentioned there in this pull request too.


Wengier:
I think the feature is completed. Since the previous patch (#1524) already took care of rest (including ShiftJS to Unicode conversion etc), so with this patch’s full Unshift JIS => Shift JIS conversion, all JIS characters (Kana, Kanji etc) should now work fine via mouse copy in PC-98 mode in addition to SBCS characters. Tested with MinGW x64 and VS2015 (both SDL1 and SDL2) builds to confirm it works.

PR #1539, opened 2020-05-10
Implement "drive z is remote=auto" and fix MS-DOS 7+ DIR/S in general

Wengier:
I noticed in the latest code the config option "drive z is remote" was added, but the "auto" setting was not yet implemented. So I made effort to implement it. Tested with SCANDISK.EXE from MS-DOS 6.20, 6.22, 7.0 (Windows 95A), 7.1 (Windows 95B and Windows 98 first & second editions) and 8.0 (Windows ME) to confirm it works.

Also, last time I fixed "DIR /S" from MS-DOS 7.1, but by testing various versions of COMMAND.COM from Windows 95/98/Me, it seems that previously it only fixed for specific versions of MS-DOS 7+ COMMAND.COM. So this time I decided to fix the issue in general. Tested with MS-DOS 7+ COMMAND.COM from Windows 95A, Windows 95B, Windows 98 first & second editions and Windows ME to confirm it works.

There is additional minor fix for the FluidSynth driver with FluidSynth 2.x.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,8 +1,9 @@
 0.83.2
-  - Added dosbox.conf option to report drive Z: as a remote
-    drive. This can be used with Windows 95 installs to prevent
-    SCANDISK.EXE from attempting to scan and "repair" drive
-    Z: which not backed by a disk filesystem.
+  - Added config option "drive z is remote" in dosbox-x.conf
+    to report drive Z: as a remote (network) drive. It is
+    "auto" by default, which will try to prevent SCANDISK.EXE
+    from Windows 9x installs to scan and "repair" drive Z:
+    which is not backed by a disk filesystem.
   - Fixed clip_key_modifier setting not working when it is
     set to alt, ctrl or shift in the SDL2 build. (Wengier)
   - Sending "Ctrl+Alt+Del" key from the menu will now reset
--- a/dosbox-x.reference.conf
+++ b/dosbox-x.reference.conf
@@ -1441,6 +1441,10 @@ dongle    = false
 #                                                     They will not be able to request control however if the DOS kernel is configured to occupy the HMA (DOS=HIGH)
 #                       hard drive data rate limit: Slow down (limit) hard disk throughput. This setting controls the limit in bytes/second.
 #                                                     Set to 0 to disable the limit, or -1 to use a reasonable default.
+#                                drive z is remote: If set, DOS will report drive Z as remote. If not set, DOS will report drive Z as local.
+#                                                     If auto (default), DOS will report drive Z as remote or local depending on the program.
+#                                                     Set this option to true to prevent SCANDISK.EXE from attempting scan and repair drive Z:
+#                                                     which is impossible since Z: is a virtual drive not backed by a disk filesystem.
 #                           hma minimum allocation: Minimum allocation size for HMA in bytes (equivalent to /HMAMIN= parameter).
 #                                         ansi.sys: If set (by default), ANSI.SYS emulation is on. If clear, ANSI.SYS is not emulated and will not appear to be installed.
 #                                                     NOTE: This option has no effect in PC-98 mode where MS-DOS systems integrate ANSI.SYS into the DOS kernel.
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -506,8 +506,8 @@ static Bitu DOS_21Handler(void) {
                 LOG(LOG_DOSMISC,LOG_DEBUG)("DOS:INT 20h/INT 21h AH=00h recovered CS segment %04x",f_cs);
 
                 DOS_Terminate(f_cs,false,0);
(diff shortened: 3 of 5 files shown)


joncampbell123:
Looks like a lot of magic values on the stack. Can you please add comments explaining why those values are checked for?

Perhaps a list of what each version of SCANDISK.EXE and DIR /S called with each version of MS-DOS/Windows?

From personal experience, the problem with magic values is that you eventually forget why you implemented them in the first place and end up rewriting it later when you have to fix something.

Wengier:
@joncampbell123 I don't think there is a "perfect" way to check for SCANDISK.EXE and DIR /S. That is why I think it has to be worked around by doing something like this. By running various versions of SCANDISK.EXE for example, I did make a list similar to:

reg_sp: 4D28 mem_readw(SegPhys(ss)+reg_sp): 17A mem_readw(SegPhys(ss)+reg_sp)+2: C65
reg_sp: 4D28 mem_readw(SegPhys(ss)+reg_sp): 1A5 mem_readw(SegPhys(ss)+reg_sp)+2: C65

reg_sp: 6A88 mem_readw(SegPhys(ss)+reg_sp): 178 mem_readw(SegPhys(ss)+reg_sp)+2: C92
reg_sp: 6A88 mem_readw(SegPhys(ss)+reg_sp): 1A3 mem_readw(SegPhys(ss)+reg_sp)+2: C92

reg_sp: 6A18 mem_readw(SegPhys(ss)+reg_sp): 16C mem_readw(SegPhys(ss)+reg_sp)+2: C88
reg_sp: 6A18 mem_readw(SegPhys(ss)+reg_sp): 197 mem_readw(SegPhys(ss)+reg_sp)+2: C88

reg_sp: 9352 mem_readw(SegPhys(ss)+reg_sp): 155 mem_readw(SegPhys(ss)+reg_sp): BB7
reg_sp: 9352 mem_readw(SegPhys(ss)+reg_sp): 17B mem_readw(SegPhys(ss)+reg_sp): BB7

reg_sp: C3B4 mem_readw(SegPhys(ss)+reg_sp): 165 mem_readw(SegPhys(ss)+reg_sp): BE0
reg_sp: C3B4 mem_readw(SegPhys(ss)+reg_sp): 190 mem_readw(SegPhys(ss)+reg_sp): BE0

reg_sp: BB9E mem_readw(SegPhys(ss)+reg_sp): 167 mem_readw(SegPhys(ss)+reg_sp): BAF
reg_sp: BB9E mem_readw(SegPhys(ss)+reg_sp): 192 mem_readw(SegPhys(ss)+reg_sp): BAF

However, this list is INCOMPLETE because there are many other non-English versions of Windows 9x. That is why it is impossible to list all possible values, and I have to check for ranges instead of specific values. I did try the Japanese and Traditional Chinese versions of Windows 98/98SE, and found them also fall in this range. So I believe this range is likely a very good approximation for all different versions of SCANDISK.EXE.

joncampbell123:
Perhaps a later commit could identify SCANDISK.EXE by reading the name from the MCB that normally sits immediately before the PSP segment?

I'll accept this for now but it might be easier to identify SCANDISK by the 8-char name in the MCB.

joncampbell123:
Wengier your printer description in dosbox-x.reference.conf disappears every time I run update-dosbox-x-reference-conf.

Most dosbox.conf settings are defined in src/dosbox.cpp so the lack of printer option descriptions might reside there.

joncampbell123:
Hold on, I see what's going on here: C_PRINTER on my system is not defined and src/dosbox.cpp doesn't add the printer section without it.

joncampbell123:
Ok, this reveals another issue in that anything non-Windows does not compile printer support (does not define C_PRINTER). configure.ac on Linux never defines it.

Wengier:
@joncampbell123 You can check for the executable name in memory of course, but that may not be perfect either - the program name can be renamed to something else too. For example, in the Japanese version it is also named SCANDSK2. You can also name another program as SCANDISK but then it will be included if you do a name check.

joncampbell123:
A check for SCANDISK, SCANDSK2, DEFRAG, etc. is fine with me, that's what AUTO is for.

joncampbell123:
Just a hunch though: How many programs care if drive Z: is local or remote?

Perhaps "auto" could just assume a remote drive and be done with it? Drive Z: is not backed by a block device which is the whole reason for this hack in the first place.

joncampbell123:
C_PRINTER isn't going to work in Linux unless it is an enable/disable in configure.ac and configure.ac also detects FreeType2.

Fortunately most Linux systems, especially if X11 is installed, have FreeType2 on them, so it's just a matter of detecting the library.

Wengier:
@joncampbell123 The current code is primarily for solving the Windows 9x/ME install issue (so that you no longer need to run SETUP.EXE with the /is option), which is where the issue was initially raised. But if you look at the issue in general, the feature can be further improved of course.

Issue #1543, opened 2020-05-11
SYS command causes (another) disk image corruption


Wengier:
The FAT driver has been updated recently to fix the filesystem corruption with the Win9x setup. But with the latest code I found another bug in the FAT driver that will cause disk corruption, this time with the SYS command from MS or PC DOS.

Previously (before today's commits) the command "SYS . C:" will report an error "Extended Error 18" or "No more files", but with the current code the command "SYS . C:" (or "SYS A: C:" etc) will succeed with the message "System transferred", but then the disk image is not bootable and more importantly, once you restart DOSBox-X the disk image is no longer readable or mountable. So the disk image has been permanently damaged from this command and is no longer usable by DOSBox-X.

Tested with the SYS command from MS-DOS 6.0 and PC DOS 2000. Same result.

The MS-DOS 6.0 disks (including SYS.COM, IO.SYS, MSDOS.SYS etc) can be found at:

https://winworldpc.com/product/ms-dos/6x

Steps to reproduce the result (using MS-DOS 6.0 DISK1.IMG):

IMGMAKE HDD.IMG -t hd_250
IMGMOUNT A: DISK1.IMG
IMGMOUNT C: HDD.IMG
VER SET 6.0
A:
SYS A: C:
Then close and restart DOSBox-X and try to mount HDD.IMG again, and if it is mounted do a "DIR C:" and see the result.

joncampbell123:
Windows 95 SYS.COM (run from the DOSBox-X command line) seems to have no issues SYSing to a floppy disk image in drive A: though.

Wengier:
Floppy images seem to be fine, but not hard disk images. I have tested hard disk images with 100MB, 250MB and 520MB (all created with IMGMAKE), all have such problems (either no longer mountable or no longer readable). I have now included steps to reproduce this problem in my first post.

joncampbell123:
I think the problem is that INT 21h AX=440Dh CL=60h (Get device parameters) is returning dummy information instead of the actual geometry and FAT filesystem information.

So SYS.COM uses the bad information and corrupts the filesystem.

After SYS C: DOSBox-X mis-identifies the drive as FAT32 (!)

Wengier:
@joncampbell123 I see. Let me try to fill the real information and see how it works.

joncampbell123:
@Wengier Filling in the correct information might require another method to the C++ DOS drive class. Let me know if it would be easier for me to implement that.

[...]

PR #1544, opened 2020-05-12
Fix SYS disk corruption (#1543)

Wengier:
As discussed in Issue #1543, SYS command can disk image corruption on hard drives. So I fixed it. Tested to confirm it works on FAT drives.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,9 +1,10 @@
 0.83.2
+  - Fixed SYS command not working properly (Wengier) 
   - DOS kernel INT 21h function Set File Attribute no longer
-    allows changing directory and volume label attributes in
-    order to prevent filesystem corruption. This prevents
-    Windows 95 setup from creating WINBOOT.INI and then
-    changing it into a directory with the call.
+    allows changing volume label attributes and fixes directory
+    attributes in order to prevent filesystem corruption.
+    This prevents Windows 95 Setup from creating WINBOOT.INI
+    and then changing it into a directory with the call.
   - FAT driver bugs fixed where a newly created zero length
     file combined with a lseek() can corrupt filesytem
     structures (root directory and/or the second FAT table).
--- a/src/dos/dos_ioctl.cpp
+++ b/src/dos/dos_ioctl.cpp
@@ -23,6 +23,7 @@
 #include "mem.h"
 #include "regs.h"
 #include "dos_inc.h"
+#include "drives.h"
 
 bool DOS_IOCTL(void) {
 	Bitu handle=0;Bit8u drive=0;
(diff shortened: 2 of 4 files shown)


joncampbell123:
Nice! MS-DOS 6.22 SYS.COM works here as well.

joncampbell123:
I was able to complete the work by adding the IOCTL calls needed by FORMAT.COM.

I can now use FORMAT.COM from MS-DOS 6.22 to format a mounted image under the DOSBox-X DOS kernel. That includes the calls made by FORMAT.COM to modify something in the partition table if formatting a hard drive.

The only issue there is that if you use FORMAT /S it will copy COMMAND.COM from drive Z: which is not very useful outside of the DOSBox-X DOS environment. SYS C: however copies the right COMMAND.COM.

joncampbell123:
I'm pretty sure the only thing preventing Windows 95's FORMAT.COM from working so far is the use of the "lock/unlock volume" IOCTLs that the DOS kernel requires you to call before you attempt writing the disk via INT 26h (or else it halts the system with an error message).

Windows 95 OSR2 and later might use entirely different IOCTLs for FAT32, which I don't expect to be implemented anytime soon until FAT32 support is better developed in the FAT driver.

Wengier:
@joncampbell123 FORMAT /S will copy the COMMAND.COM file as specified in %COMSPEC% environment variable. So just specify the correct path for it and it should work as expected.

Wengier:
By the way, why not improve FAT32 support in the FAT driver so that FAT32 disk images can be accessed directly in the DOSBox-X shell as well? FAT32 has arguably become the most important disk format for DOS & Windows 9x/ME in the last two decades.

Wengier:
@joncampbell123 The command "FORMAT A:" using FORMAT.COM from MS-DOS **7.0** already seems to work with the current code (even though LOCK/UNLOCK is already supported in MS-DOS 7.0). But the same command from MS-DOS **7.1** does not work yet. It will call Int21 functions AX=7304h and AX=7305h/CX=FFFFh which are not yet implemented.

joncampbell123:
@Wengier Yup. Those INT 21h calls are the new read/write disk functions, especially for FAT32 drives. http://www.ctyme.com/intr/rb-3229.htm

SYS.COM in Windows 95 does not work yet because it relies on the volume locking functions to work.

joncampbell123:
I think the FAT driver could use a bit of cleanup before fully adding FAT32 since some fundamental changes happened between FAT16 and FAT32.

Besides the obvious change of the FAT table, there is also the root directory which now has it's own allocation chain instead of a fixed area on the disk. The root directory has a starting cluster number.

There are also two "boot sectors" and an extra sector with more info stored in it.

http://hackipedia.org/browse.cgi/Disk%20formats/Filesystems/FAT%2c%20File%20Allocation%20Table%20%28DOS%2c%20Windows%29/Microsoft%20Extensible%20Firmware%20Initiative%20FAT32%20File%20System%20Specification%20%282000%2d12%2d06%29%20v1%2e03%2epdf

(Wow, a "whitepaper" with actual technical documentation in it for once!)

Wengier:
@joncampbell123 SYS.COM from Windows 9x will work fine with the current code if the current reported DOS version is less than 7.0. With a reported DOS version of 7.x, Win9x SYS.COM will work properly if you add a dummy handler for the volume locking functions in dos_ioctl.cpp, e.g.

``case 0x4A:``
``case 0x4B:``
``case 0x6A:``
``case 0x6B:``
``	break;``

Since this code is very simple and I don't really think there are negative effects, perhaps you can add it to the dos_ioctl.cpp file.

joncampbell123:
That would work fine for now.

At some point I should pull out my old MSDN 6.0 / MSVC 6.0 CDs for more documentation on those IOCTLs and their function. Believe it or not back then Microsoft actually bothered to document those IOCTLs in the MSDN library at the time.

joncampbell123:
I think they also documented it in MSVC 4.2's online help, which I also have.

When I find the documentation I'll add comments to the source code to better describe what it all means and how it's supposed to work around those IOCTLs.

joncampbell123:
Windows 98 FORMAT.COM seems to have no problem with formatting if the disk is < 2GB (FAT12 or FAT16 format).

Wengier:
@joncampbell123 I think MS-DOS 7.1/Windows 98 FORMAT.COM will proceed and finish disk formatting as normal,  but when you use DIR again, everything previously on the disk are still there (so the formatting process was completely fake). On the other hand, MS-DOS 7.0 FORMAT.COM does not have this problem (the disk will be empty).

joncampbell123:
Nope. FORMAT.COM from Windows 98 worked correctly here. I tested a 800MB and 1600MB disk image.

Wengier:
Also, "SCANDISK A:" for a floppy image seems to always return the message "ScanDisk cannot examine drive A." and exit. SCANDISK for other drives will work though.

Wengier:
@joncampbell123 The latest code (d61f468) will not compile due to the function definition mismatch in the file dos_program.cpp. Tested with MinGW and VS2015.

joncampbell123:
Another possible thing to consider is that FORMAT.COM generates a new FAT filesystem. One that might not line up with the layout already loaded by the FAT driver. FORMAT.COM calls INT 21h AH=32h to return the DPB, which is always supposed to read the disk to return it.

At some point INT 21h AH=32h in DOSBox-X is going to have to reread the DPB and notify the FAT driver to re-check the BPB and reload the filesystem parameters (and flush caches).

I also checked that FORMAT.COM in Windows 95 and below "format" the hard drive by issuing INT 13h AH=4 VERIFY, not by actually formatting.

joncampbell123:
Second, this was clearly not explained well by RBIL, but the read/write/format/verify IOCTLs use C/H/S values that are relative to the drive partition.

So when FORMAT.COM does a logical read/write of C/H/S 0/0/0, it's reading/writing the boot sector of the partition, not the MBR. This was verified by watching the same FORMAT.COM do it's thing from a Windows 95 bootdisk on an already partitioned hdd image.

joncampbell123:
Checking now.

joncampbell123:
@Wengier What is the error message? It compiles find here.

joncampbell123:
By the way that commit adds the option to MOUNT a local folder to a drive letter and report it as remote so that SCANDISK.EXE doesn't complain about drives mounted like that.

joncampbell123:
@Wengier It turns out Windows 98 FORMAT.COM is using INT AX=7305h, but it never got an error message because of the way INT AX=7303h was implemented, which probably due to bad indentation was never given a break statement or an else return error case (fixing now).

Wengier:
@joncampbell123 The error messages from MSVC are:

Error	C2661	'cdromDrive::cdromDrive': no overloaded function takes 8 arguments	dosbox-x	dos_programs.cpp	224	
Error	C2661	'localDrive::localDrive': no overloaded function takes 6 arguments	dosbox-x	dos_programs.cpp	242	
Error	C2661	'cdromDrive::cdromDrive': no overloaded function takes 8 arguments	dosbox-x	dos_programs.cpp	334	
Error	C2661	'localDrive::localDrive': no overloaded function takes 6 arguments	dosbox-x	dos_programs.cpp	352	

For example in line 224:

newdrive  = new cdromDrive(drive,temp_line.c_str(),sizes[0],bit8size,sizes[2],0,mediaid,error);

There is no new "option" parameter in this code.

joncampbell123:
Heh, if DOSBox-X reports DOS 7.10, SCANDISK.EXE will try to use both INT 25h and INT AX=7305h. Unlike FORMAT.COM though SCANDISK.EXE falls back to INT 25h when INT AX=7305h doesn't work.

joncampbell123:
Ah, that part is Windows only, which is why it doesn't error out here on Linux.

joncampbell123:
Latest commit adds INT AX=7305h, but only if the reported DOS version is 7.10 or higher. Windows 98 FORMAT.COM and SCANDISK.EXE seem happy now.

Wengier:
@joncampbell123 Yes, both FORMAT.COM and SCANDISK.EXE from MS-DOS 7.1/Windows 98 work in the latest commits (except for "SCANDISK A:"). I think right now a few MS-DOS 7.1+ functions are still missing, perhaps they can be properly implemented once FAT32 disk support is improved in the FAT driver.

sikthehedgehog:
Minor nitpick on the convo, but I had to look up to figure out that
you meant INT 21h AX=7305h :P (hey, some BIOS calls take function
numbers too >.>)

Wengier:
@joncampbell123 I have now opened a new issue for direct FAT32 support in Issue #1547.

PR #1551, opened 2020-05-14
MPUIRQ=2 is valid for an XT

rderooy:
# Description
The original Roland interfaces used IRQ2, and this is valid for an XT which only has one PIC.

_Summary of changes brought by this PR._
Set the default IRQ for an XT to IRQ2 and if MPUIRQ > 7, ignore it.
Does a 80186 have one or two PICs? This patch assumes it is the same as the 80286.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,8 @@
 0.83.2
+  - Added DOS functions for "FAT32 extended" absolute disk
+    read and write used by FAT32-aware versions of FORMAT.COM
+    and SCANDISK.EXE (shipped with Windows 95 OSR2 or later).
+    The FAT driver in DOSBox-X does not yet fully support FAT32.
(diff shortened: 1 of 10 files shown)


joncampbell123:
@rderooy It's actually much simpler than that. DOSBox-X has an option "enable slave pic" to enable/disable the secondary PIC.

Please use that setting instead of assuming that cputype=8086 means XT PIC configuration. I'm pretty sure there are XT 286's out there that *also* have an XT PIC configuration as well.

joncampbell123:
Add "extern bool enable_slave_pic;" somewhere near the top of mpu401.cpp and replace "CPU_ArchitectureType == CPU_ARCHTYPE_8086" with "!enable_slave_pic".

If you like, I could make this change myself and then close this pull request.

rderooy:
@joncampbell123 sure go ahead. Otherwise I will look at it tomorrow.

PR #1559, opened 2020-05-16
Display real serial numbers for local and FAT drives

Wengier:
I noticed that commands like DIR and VOL (and external tools too) always report "0000-1234" as the volume label in DOSBox-X, so I updated the code to display real serial numbers for local drives and FAT drives whenever possible. Support for serial numbers in mounted local drives is Windows-only though.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -22,6 +22,8 @@
   - Added DOS IOCTL read/write logical device track functions
     so that FORMAT.COM is able to verify and modify the partition
     table to successfully format a hard drive image.
+  - DIR and VOL command now display real serial numbers for mounted
+    local and FAT drives (Wengier)
   - Fixed SYS command not working properly (Wengier) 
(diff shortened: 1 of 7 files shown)


Allofich:
The Visual Studio build gives the warning `'localDrive::GetSerial': not all control paths return a value`.

joncampbell123:
@Allofich Just fixed that. Try again. Latest commit.

Wengier:
@joncampbell123 It should return 0x1234 instead of 0 in such case because the former (0x1234) is the defined (fake) value returned everywhere else in the code.

joncampbell123:
MS-DOS however is said to return 0000-0000 for volumes without a serial number. 0000-1234 should be the fake value returned only by drive_local in my opinion.

Allofich:
> Try again. Latest commit.

No warning anymore with the new code.

Issue #1555, opened 2020-05-15
Win98 setup.exe - runtime error R6001


rderooy:
**Describe the bug**
With ver=7.10
Running Win98SE setup.exe from FAT32 volume shows:
```
C:\WINDOWS\OPTIONS\CABS>setup
Please wait while Setup initializes

run-time error R6001
- null pointer assignment
Setup is now going to perform a routine check on your system.

To continue, press ENTER. To quit Setup, press ESC.
```
Log shows:
```
LOG: 1122839196 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1122839202 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1122839209 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1122839216 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1122840192 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1122842109 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=1615
LOG: 1122875456 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1122875465 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1122875475 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1122875485 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1122876477 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1122878246 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=D201
LOG: 1122885907 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=1686
LOG: 1122885960 ERROR DOSMISC:DOS:INT 2F Unhandled call AX= 600
LOG: 1122885974 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=B700
LOG: 1122890627 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=160A
LOG: 1124215992 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
LOG: 1124216001 ERROR DOSMISC:DOS:INT 2F Unhandled call AX=122E
[...]

PR #1562, opened 2020-05-16
Supress "run-time error R6001" and "Assertion failed!" errors

Wengier:
As mentioned in Issue #1555, Windows 98 SETUP program will show "run-time error R6001" with DOS version 7+. I found the reason is that Windows 98 will call a DOS 7+ function, and if it succeeds, will do some weird things not yet supported by DOSBox-X. By failing the function for Windows 98 SETUP it will no longer show such error message.

Also, I noticed that when using the SRCBOOT.COM tool to save the boot sector of a FAT32 drive, DOSBox-X will pop up a message box with the "Assertion failed!" error, even though it will save the boot sector properly if I click "Ignore". So I fixed this too by suppressing the assertion error.

### DIFF
--- a/src/dos/dos_misc.cpp
+++ b/src/dos/dos_misc.cpp
@@ -213,15 +213,23 @@ static bool DOS_MultiplexFunctions(void) {
         reg_ax=0;
         return true;
     case 0x1611:    /* Get shell parameters */
+	{
 		if (dos.version.major < 7) return false;
-        strcpy(name,"COMMAND.COM");
-        MEM_BlockWrite(SegPhys(ds)+reg_di,name,(Bitu)(strlen(name)+1));
-        strcpy(name+1,"Z:\\COMMAND.COM /P");
-		name[0]=(char)strlen(name+1);
-        MEM_BlockWrite(SegPhys(ds)+reg_si,name,(Bitu)(strlen(name)+1));
-        reg_ax=0;
-        reg_bx=0;
-        return true;
+		char psp_name[9];
+		DOS_MCB psp_mcb(dos.psp()-1);
+		psp_mcb.GetFileName(psp_name);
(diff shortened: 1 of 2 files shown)


joncampbell123:
Does "get shell parameters" have anything to do with the COMSPEC environment variable?

Wengier:
@joncampbell123 I have tried to check the actual returned strings of this function in real MS-DOS 7.1, and it does not seem to include %COMSPEC% environment variable stuff. But in any case, this function itself is probably not a real cause of the error, Windows 98 SETUP seems to call this function to confirm if MS-DOS 7.x is running.

joncampbell123:
Here's what INT 2Fh AX=1611h returns on an actual Windows 98 system running from a hard disk image.

guest os_000

joncampbell123:
http://www.ctyme.com/intr/rb-4525.htm

DS:DX = "C:\WINDOWS\COMMAND.COM"
DS:SI = "C:\WINDOWS /P /D /K AUTOEXEC"

Wengier:
@joncampbell123 The "C:\WINDOWS\" part will have something to do with the %windir% or %winbootdir% variables. If not set (as in my case), they will not be part of the strings.

joncampbell123:
Just for reference, Windows 95 returns the same thing:
guest os_000

Wengier:
According to my case, the returned strings are:
DS:DX = "COMMAND.COM"
DS:SI = "/P /D /K AUTOEXEC"

joncampbell123:
@Wengier Windows 95/98 bootdisk?

Wengier:
@joncampbell123 Yes, but customized with the SHELL= command in CONFIG.SYS, e.g. 

SHELL=COMMAND.COM /P

Instead of something like:

SHELL=C:\WINDOWS\COMMAND.COM /P

PR #1563, opened 2020-05-17
Support for directory renaming on FAT drives

Wengier:
I noticed that you can rename both files and directories on local drives, but you can only rename files but not directories on FAT drives, so I decided to add directory renaming on FAT drives too. For example, "REN DIRA DIRB" will rename a directory named DIRA TO DIRB on FAT drives just like on local drives.

I also grouped recent changes (especially those regarding FAT32 support) in CHANGELOG to make it more readable.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,13 +18,14 @@
     for "FAT32 extended" absolute disk read and write.
   - Improved FAT32 support to the FAT driver, including direct
     support for FAT32 drives in the DOSBox-X command shell. Files,
-    directories and volume labels on FAT32 drives can be accessed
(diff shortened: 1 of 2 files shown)


joncampbell123:
Hold on... checking if MS-DOS and Windows 9x permit renaming directories...

Wengier:
@joncampbell123 On MS-DOS 7+/Windows 9x you certainly can. Furthermore, you could already rename directories on mounted local drives previously.

joncampbell123:
The MS-DOS 6.22 REN (rename) command doesn't seem to accept renaming directories.

Windows 95 COMMAND.COM does, however.

PR #1565, opened 2020-05-18
Further fix for Fluidsynth default values and support changing the new "lfn" option

Wengier:
The previous pull request #1548 by @rderooy tried to fix the default values of Fluidsynth for different platforms (Windows or non-Windows), but I have noticed an issue with the fix: the dosbox-x.reference.conf file will use either the Windows value or the non-Windows value, thus it is not portable. So I improved his fix by adding a value "default" for these options, which will adjust accordingly at DOSBox-X startup for different platforms. This will make the config file itself portable.

In addition, I made it possible to change the newly added "lfn" config option with the "config -set" command.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,13 +5,13 @@
   - FAT driver cleaned up and fix to avoid edge cases that
     can corrupt directory entries and leave lost clusters
     on the disk.
-  - INT AH=36h fixed to convert free space but maintain a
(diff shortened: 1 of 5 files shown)


joncampbell123:
From time to time I run a script to regenerate dosbox-x.reference.conf from the defaults in src/dosbox.cpp

Issue #1547, opened 2020-05-13
Support for FAT32 disk image access from the DOSBox-X shell


Wengier:
As mentioned in the discussion of pull request #1544, right now DOSBox-X's support for the FAT32 file system in disk images is very limited, i.e. it does allow FAT32 disk image to boot with a real DOS, but there is no direct support for it from the DOSBox-X shell at all. As a result, DOSBox-X cannot create or access any disk images that are greater than 2GB (or any FAT32 disk images that are between 512MB and 2GB), which (significantly) limits the usefulness of its FAT support. Also, without direct support for FAT32 drives the MS-DOS 7.1+ functions for FAT32 file system cannot be properly implemented either.

Since FAT32 has become a critically important file system for DOS and Windows 9x/ME in the last two decades, and (as stated in dosbox-x.com) two of the main focuses of DOSBox-X are to give the user all the options they need to emulate everything from original IBM PC hardware with 64KB of RAM all the way up to late 90's hardware, and also as a great platform for emulating Windows 3.x, 95, 98 and software written for those versions of Windows, this feature will be extremely helpful for both of these two goals.

joncampbell123:
The latest commit 7c3a01cea8aa10332923698dacef67a16b831e63 has basic functional FAT32 support, including directory enumeration, file open and file create support.

However some IOCTLs are lacking and Windows 98's SCANDISK.EXE will try to fix the filesystem as if FAT16 because of it, so don't run SCANDISK.EXE until that's fixed.

Wengier:
@joncampbell123 Great job! I will be sure to check it out tomorrow.

joncampbell123:
I just implemented INT 21h AX=7302h and now SCANDISK.EXE is happy, except it thinks the free space is being incorrectly reported.

joncampbell123:
FORMAT.COM is now happy in the latest commit when formatting FAT32.

There's probably still some bugs to iron out but it seems stable so far.

joncampbell123:
I got free disk space reporting updated to support FAT32 partitions properly.

The only thing left to figure out is why SCANDISK.EXE always complains about incorrectly reported free space on FAT32 drives.

Wengier:
@joncampbell123 I have tested a 1GB FAT32 image with the latest code. It does indeed work, but I have found following issues in the code (besides the free disk space issue):

[...]

PR #1569, opened 2020-05-19
IMGMOUNT command (with argument) to list mounted FAT/ISO drives and drive numbers

Wengier:
I noticed that MOUNT command without parameters will list mounted drives, but IMGMOUNT command without parameters will show its usage. Apparently this is not consistent  (both "MOUNT /?" and "IMGMOUNT /?" will show their usages though). I fixed this so that IMGMOUNT command (without parameters) will list mounted FAT/ISO drives and mounted drive numbers.

I also fixed the issue that entering "C:" as the directory name opens a file (or subdirectory) in it instead of listing the files in C: for FAT drives in MS-DOS 7.x EDIT as mentioned in Issue #1547.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -33,7 +33,9 @@
     FAT drives just like on local drives (Wengier)
   - DIR and VOL commands now display real serial numbers for mounted
     local drives (Windows only) and FAT drives (Wengier)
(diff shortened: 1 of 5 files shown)


Wengier:
SUBST command (without parameters) now lists mounted local drives (similar to real DOS).

PR #1570, opened 2020-05-19
Extended SFN-only mode for SFN calls

Wengier:
In the previous code (besides FindFirst/FindNext) the SFN-only mode was only added for the SFN call of opening files through the use of the force variable. I renamed this variable to a more descriptive name of "force_sfn" and also extended to other SFN calls including UNLINK/MKDIR/RMDIR too in order to make it more complete.


joncampbell123:
Thanks for the clarification. I was wondering what "force" was for. Merging now.

PR #1587, opened 2020-05-22
Config file cleanup: move PC-98 related settings to the new [pc98] section

Wengier:
I noticed that in the dosbox-x.reference.conf file there are more than 20 PC-98 related settings. They all begin with "pc-98 " - most of them in the [dosbox] section, and another in the [dos] section. So I decided to clean them up by making a separate [pc98] section, and move all these PC-98 related settings to this section in the config file.

I also moved the [sdl] section in dosbox-x.reference.conf to the top, instead of [log]. The [log] section is only useful for developers and very technical users, whereas the [sdl] section is apparently useful for all users. Putting the [log] section on top can scare away some users I think.

In addition, for the auto setting the isRemote() function in drive_local.cpp and drive_virtual.cpp now check for the DOS executable name "SCANDISK", in addition to the check on the stack. The stack check is still useful even if the executable name has been renamed to some other names so it is kept as is.

Tested to ensure the code works.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 0.83.2
+  - Added [config] section in dosbox-x.conf to resemble DOS's
+    CONFIG.SYS file, although it currently only supports BREAK,
+    LASTDRIVE and REM commands (Wengier)
   - Moved PC-98 related config options from [dosbox] and [dos]
-    sections to a new [pc98] section. (Wengier)
+    sections to the new [pc98] section. (Wengier)
   - Config option "dpi aware" now supports the "auto" setting
     to auto-decide on the best setting for the platform. This
     fixes very small window issue on e.g. Surface tablets.
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.83.2
+  - Moved PC-98 related config options from [dosbox] and [dos]
+    sections to a new [pc98] section. (Wengier)
   - Config option "dpi aware" now supports the "auto" setting
     to auto-decide on the best setting for the platform. This
     fixes very small window issue on e.g. Surface tablets.
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,8 +2,11 @@
   - Added [config] section in dosbox-x.conf to resemble DOS's
     CONFIG.SYS file, although it currently only supports BREAK,
     LASTDRIVE and REM commands (Wengier)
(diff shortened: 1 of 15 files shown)


Wengier:
I have now added a new (stub) [config] section to the dosbox-x.reference.conf file, which is designed to resemble the DOS CONFIG.SYS file, although it currently only supports the REM, BREAK and LASTDRIVE commands.

joncampbell123:
Please mention in CHANGELOG that the change moves pc-98 settings to their own and that existing dosbox.conf files will need to move those settings as well i.e. that will break existing dosbox.conf files that use it.

joncampbell123:
Glad to see a [config] section. I've thought about doing that as well, but also to support DEVICE= lines to emulate device drivers, though... since CONFIG.SYS happens before AUTOEXEC.BAT how would you imgmount or mount C: to make that happen...

Wengier:
@joncampbell123 Sure, CHANGELOG now mentions that "Existing dosbox-x.conf/dosbox.conf files will need to move those settings ("pc-98 ...") to the new [pc98] section in order to keep these setting continue to work". Since these settings are **not** renamed, so this should be an easy process, considering how many users are actually using the PC-98 mode.

Wengier:
@joncampbell123 I would like to see a DEVICE= command in [config] section too, e.g. to load the IFSHLP.SYS device driver for use with WFW 3.11 and Win9x. As for the MOUNT/IMGMOUNT commands issue, I think this can probably be solved with the INSTALL= command, which allows you to load executables which will include MOUNT.COM and IMGMOUNT.COM.

Wengier:
The INSTALL command has been supported in CONFIG.SYS since DOS 4.0 I think.

joncampbell123:
I'll accept the moving of pc-98 settings as long as I can get the dosbox.conf handling to remap the settings from the old location onto the new location i.e. the pc-98 settings can exist in either section and work. Writing of course will place it in the new location.

joncampbell123:
Since the setting names never changed, I think the remap is important to avoid breaking other people's dosbox.conf files including the test setups I've made already.

Wengier:
@joncampbell123 Yes, I am currently working on this. I will submit a new pull request later today.

Wengier:
I have already finished the remapping of these settings from other sections to the [pc98] section. Please check out my pull request #1593.

Issue #1588, opened 2020-05-23
New default dosbox-x config dir not XDG compliant


rderooy:
**Describe the bug**
With the latest build, the default dosbox config directory has changed from ~/.dosbox to ~/.dosbox-x

If we want to change the default directory, I think it should be done properly, and it should be XDG compliant. As such I am asking to change it to ~/.config/dosbox-x before the next release.

There is an enviroment variable that can overwrite the default location, but by default it is not set ($XDG_CONFIG_DIRS).

http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

Wengier:
I have changed user directory from ~/.dosbox-x to ~/.config/dosbox-x as suggested in my pull request #1593. The environment variable will not be used to reduce complexity (DOSBox-X will load dosbox-x.conf from the current directory first after all).

PR #1593, opened 2020-05-24
Auto-redirect PC-98 related settings to [pc98] section and improve handling of [config] section

Wengier:
The [pc98] section was recently added to the config file, and PC-98 related settings were moved to this section from other sections. But this could break existing config files for PC-98 related options. So I fixed this issue in this pull request by automatically redirecting existing PC-98 related options to the new [pc98] section from the other sections if the config file's [pc98] section is empty when DOSBox-X starts. Thus existing config files will continue to work for these settings.

I also improved the new [config] section to support SET, DEVICE/DEVICEHIGH, INSTALL/INSTALLHIGH commands. For DEVICE/DEVICEHIGH commands, currently it will simply use
the built-in DEVICE.COM program to load the specified device driver(s). The built-in LH command will be used for HIGH options. A example of these commands:

```
INSTALL=MOUNT.COM C: C:\WINDOWS
DEVICE=C:\IFSHLP.SYS
SET TEMP=C:\TEMP
```

A new file named CONFIG.SYS will now appear in the Z: drive for the contents of the [config] section (similar to the AUTOEXEC.BAT file in the Z: drive). A new command-line option -noconfig is added to skip the [config] section. This section will also be skipped if the secure mode is enabled.

I also moved the user directory from ~/.dosbox-x to ~/.config/dosbox-x in order to be XDG complicant, as stated by @rderooy in Issue #1588. But DOSBox-X will always load the dosbox-x.conf file from the current directory first.

Last, I improved the handling of quotes for some commands.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,15 +1,19 @@
 0.83.2
   - Added [config] section in dosbox-x.conf to resemble DOS's
-    CONFIG.SYS file, although it currently only supports BREAK,
-    LASTDRIVE and REM commands (Wengier)
+    CONFIG.SYS file. It currently only supports REM, BREAK, 
+    SET, DEVICE/DEVICEHIGH, INSTALL/INSTALLHIGH and LASTDRIVE
+    commands. The file CONFIG.SYS will appear on the Z: drive.
+    It is possible to bypass the [config] section with the
+    -noconfig command-line option (Wengier)
   - Moved PC-98 related config options (starting with "pc-98 ")
     from [dosbox] and [dos] sections to its own [pc98] section.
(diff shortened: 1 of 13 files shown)


Wengier:
The files= config option has also been moved from the [dos] section to the new [config] section because as described it is equivalent to the CONFIG.SYS FILES= option. DOSBox-X will also automatically redirect existing files= options from the [dos] section to the [config] section if the config file's [config] section is empty.

joncampbell123:
It works fine here, thanks.

I made an additional commit to move the <sstream> include out of the #ifdef WIN32 so that it compiles here on Linux.

PR #1595, opened 2020-05-25
Allow only case-changing for LFN rename and support for "DOS=HIGH, UMB" in [config] section

Wengier:
The REN command could rename files when the old name and new name are different, but it did not work if the old name and new name are different only by case, e.g. "REN NAME name". So I updated the code to allow this in LFN mode.

Also, the [config] section now supports the "DOS" option, so "DOS=HIGH/LOW, UMB/NOUMB" as in DOS's config.sys is now supported. The previous "dos in hma=true/false" option in [dos] section will be automatically converted to "DOS=HIGH/LOW" if the [config] section does not exist in the config file.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,7 @@
 0.83.2
   - Added [config] section in dosbox-x.conf to resemble DOS's
-    CONFIG.SYS file. It currently supports REM, BREAK, FILES,
-    SET, DEVICE/DEVICEHIGH, INSTALL/INSTALLHIGH and LASTDRIVE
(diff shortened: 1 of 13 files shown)


Wengier:
Since it is almost the end of the month, I think this will be my last pull request to add features for this month. I will only make pull request if I find bugs for the rest of the month.

Wengier:
Some additional fixes are made, such as moving some of the [config] section handing code out of the Win32-only block to work for all platforms.

joncampbell123:
Ready to merge?

joncampbell123:
I'll take that as a yes. Merging now.

PR #1602, opened 2020-05-26
Fixed "config -wc(d)" not writing certain [config] options

Wengier:
I noticed an issue with the "config -wc(d)" command when writing the [config] section, i.e. it did not write commands like set/device/devicehigh/install/installhigh (because they are handled slightly differently), so I fixed this issue in this pull request. As a result,  "config -wc(d)" should now write all supported commands to the [config] section.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,5 @@
 0.83.2
+  - Fixed issues with the "config -wcd -all" command (Wengier)
   - Added DTASEG, DTAOFF, and PSPSEG as hex value constants to
     the debugger interface to aid in debugging DOS programs.
   - Added [config] section in dosbox-x.conf to resemble DOS's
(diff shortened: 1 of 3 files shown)


Wengier:
@rderooy I have already added special handling to the SET/INSTALL(HIGH)/DEVICE(HIGH) config.sys commands in my pull request so that "config -wcd -all" should now add them to the config file template even if they don't exist. It also allows "config -get config" to show these commands in addition to the other options, just like the content of the CONFIG.SYS file. But it is impossible to change these extra options with "config -set" (this can never be supported because they are simply incompatible with each other).

"LASTDRIVE=A" was the existing setting in DOSBox-X before I added the [config] section.  DOSBox-X does not support external (character or block) device drivers so LASTDRIVE=A will work fine at this time. On the other hand, if you change it to "LASTDRIVE=Z" and try to run Windows 98 Setup, it may complain that SUBST.EXE & JOIN.EXE are running.

Issue #1596, opened 2020-05-25
config -all -wcd still writes to wrong location


rderooy:
**Describe the bug**
With the new default config directory, I was expecting config to write to it. But it does not.

```
config -l
DOSBox-X 0.83.2 configuration directory:
/home/robert/.config/dosbox-x/
```

But running ```config -all -wcd``` still writes to ~/.dosbox

rderooy:
Actually, it seems -wcd no longer writes to the config directory, but rather the current directory.

rderooy:
I also think config -wcd should print the path where it wrote the file.
```
config -all -wcd
Writing config file dosbox-x-0.83.2.conf
```

The help for ```config``` states for the -wcd option
```
-wcd
 Write to the default config file in the config directory
```

It is also very sensitive to the order of the options.
```config -all -wcd``` functions different then ```config -wcd -all```

Wengier:
I have fixed both issues with the "config -wcd -all" command, i.e. , the issue that it writes to the current directory instead of the config directory, and the issue that "config -wcd -all" does not work the same way as "config -all -wcd".

As for the config -wc or -wcd options, it will indeed print out the path of the file written, if the path is anything other than the current directory. If there is no directory mentioned, then it means that it writes to the file in the current directory.

rderooy:
@Wengier 

One more issue with writing the default config file. If the directory does not yet exist, you will get an error.

![command_000](https://user-images.githubusercontent.com/17701098/82987111-4dd89700-9ff7-11ea-9034-2855ad2b318b.png)

After manually creating the directory, it succeeds.

Wengier:
@rderooy In my pull request #1604, I added the ability to auto-create the ~/.config/dosbox-x directory if it does not yet exist. For now I will assume the parent directory already exists.

rderooy:
I think the ~/.config directory should exist in most cases, at least with the Gnome and KDE desktops for sure, and lots of Linux applications use it as it is the default directory as defined by XDG, so I'm reasonably sure other desktop environments should also have it.

PR #1604, opened 2020-05-27
CONFIG command fixups

Wengier:
As mentioned by @rderooy in Issue #1596 and the discussion of pull request #1602, the "config -wcd -all" command may fail if the directory does not yet exist, and it does not add config.sys options like SET/INSTALL(HIGH)/DEVICE(HIGH) to the config file template if they don't exist yet, so I fixed them. It also allows "config -get config" to show SET/INSTALL(HIGH)/DEVICE(HIGH) commands in addition to the other options, just like the command "type config.sys". But it is impossible to change these commands with "config -set" because they can appear multiple times thus they are not regular options like the others.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -36,8 +36,11 @@
     the MBR lacks executable code, or the first partition
     is Windows 98-style LBA FAT16 or FAT32.
   - IMGMOUNT now assumes ISO type by default if the image file
-    extension is ".iso". No need for "-t iso" in this case, but
-    it can be overridden by e.g. "-t hdd". (Wengier)
+    extension is .iso/.cue/.bin/.mdf. No need for "-t iso" in
+    this case; it can be overridden by e.g. "-t hdd". (Wengier)
+  - IMGMOUNT command (no parameters) now lists mounted FAT/ISO
+    drives and mounted drive numbers, also SUBST command (no
+    parameters) now lists mounted local drives (Wengier)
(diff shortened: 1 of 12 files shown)


Wengier:
When mounting images files with the .iso/.cue/.bin/.mdf extensions, IMGMOUNT will now assume ISO type by default, so you no need to manually type "-t iso" in the command in this case. But it can be overridden by e.g. "-t hdd" (if applicable).

Allofich:
Regarding this commit
https://github.com/joncampbell123/dosbox-x/commit/8ba1327ce751057f0f0e732ca27183188cc492ea

I see warnings in the Mac OS X GCC build (using Travis CI) like:

```shell_misc.cpp:1202:107:` warning: address of stack memory associated with local variable 'spath' returned [-Wreturn-stack-address]
                        if (DOS_FileExists(which_ret)) return strchr(which_ret, '\"')&&DOS_GetSFNPath(which_ret, spath, false)?spath:which_ret;```

for each of these lines returning `spath`.

The warning means that the address of `spath` is part of the return value, but `spath` is declared within this function and will no longer exist after the call is exited. `which_ret` is OK.

Wengier:
@Allofich Thanks for pointing this out. Since the code worked fine as is, I did not always pay too much attention to the compiler warnings. But in any case, I believe I have fixed this in my new pull request #1610.

PR #1606, opened 2020-05-28
4DOS fixes and -helpdebug option to list debug-related options

Wengier:
I noticed that in certain situations 4DOS DIR command may not list all files in the directory properly when LFN is enabled because of issue related to the filefind handle, so I fixed it. As a result, 4DOS DIR should now work correctly in these situations.

I also added the -helpdebug command-line option to list all debug-related command-line options such as -early-debug, -keydbg, -log-con and -log-fileio (these options are only useful for developers and very technical users, but not normal users). The standard -? or -help option will no longer list them, but the new -helpdebug command-line option will be listed by -? or -help for these who indeed want to use the debugging feature.

Finally, some small cleanups are made to the options of the built-in DIR command: options such as /-A and /-P can be used to override /A and /P etc if the latter are specified in the %DIRCMD% environment variable.

Tested to ensure the build works properly.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,5 @@
 0.83.2
   - Fixed issues with the "config -wcd -all" command (Wengier)
-  - Added DTASEG, DTAOFF, and PSPSEG as hex value constants to
-    the debugger interface to aid in debugging DOS programs.
   - Added [config] section in dosbox-x.conf to resemble DOS's
     CONFIG.SYS file. It currently supports REM, BREAK, NUMLOCK,
     FCBS, FILES, DOS, DEVICE/DEVICEHIGH, INSTALL/INSTALLHIGH,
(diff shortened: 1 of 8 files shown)


joncampbell123:
What does 4DOS do differently with LFNs or directory enumeration?

Wengier:
@joncampbell123 4DOS (8.00) DIR has two modes, i.e. SFN-only mode and LFN-enabled mode, and their output format are very different. SFN-only mode seems to have no issue previously, but In LFN-enabled mode problem might occur in certain situations which is fixed in this pull request.

joncampbell123:
I'm seeing a merge conflict with dos_classes.cpp, can you confirm which version is correct?

Wengier:
@joncampbell123 The version with just "int i;" (@Allofich changed to "unsigned int i;" in his previous pull request, which is fine) is correct. I moved the two "sSave" lines slightly down in this pull request, so these two lines here are now duplicate.

So the merged version should be "unsigned int i;" instead of the two versions above.

PR #1617, opened 2020-05-30
Fixed ^G unexpected behavior in the shell and added help messages for commands

Wengier:
I noticed that the ^G character (beep) can cause unexpected behavior in the command shell: it will not move the cursor by itself, but if you later press keys such as ESC and BackSpace to delete the entire input line, then the command prompt (e.g. Z:\\>) can be deleted as well. So I fixed this by moving the cursor filled with a white space when the ^G is pressed. As a result, keys such as ESC and BackSpace will no longer delete the command prompt in such case, and ^G will still function properly.

Also, I added help messages for some commands in the Z: drive (such as MIXER and RESCAN). So "command /?" will show the help messages for these commands.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,12 +1,16 @@
+DOSBox-X by joncampbell123
+
+Based on DOSBox by the DOSBox Team
+
 The DOSBox Team
 ---------------
 
 Sjoerd v.d. Berg <harekiet>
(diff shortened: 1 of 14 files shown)


joncampbell123:
Ready to merge?

Wengier:
@joncampbell123 Yes, you can merge it now.

PR #1620, opened 2020-06-01
Add COUNTRY command for country-specific date & time formats

Wengier:
I noticed that DOSBox-X did not support the set country code function, and all date & time formats (such as the DIR and DATE/TIME commands) were fixed to the U.S. format. Apparently this is not very flexible for non-US users. So I solved this by adding the COUNTRY command which allows to change the country code within DOSBox-X, and then date and time will be shown according to the formats associated with the specific country code. For example, "COUNTRY 61" will set the country code to 61 (International English) which will use the DD-MM-YYYY date format instead of the U.S. MM-DD-YYYY date format.

I also added comment to the config file that FAT32 disk images will be supported with a reported DOS version of 7.1 or higher to reflect the recent change to the FAT driver. It uses the current reported DOS version, not the DOS version at start though.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.83.3
+  - Added COUNTRY command to set country code for country-
+    specific date and time formats. (Wengier)
   - DOS 440Dh IOCTL function 67h (get access flag) added to
     allow FDISK.EXE to determine FAT filesystem type.
   - Unknown DOS 440Dh IOCTLs warnings now indicate whether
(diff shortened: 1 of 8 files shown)


joncampbell123:
@Wengier Just curious, why do you write code that compares *integers* to NULL?

src/shell/shell_cmds.cpp line 1029:

    if (sec==NULL&&msec==NULL)
        sprintf(retBuf, "%2u%c%02u%c", hour, sep, min, *ampm);
    else
        sprintf(retBuf, "%u%c%02u%c%02u%c%02u%s", hour, sep, min, sep, sec, dos.tables.country[9], msec, ampm);

joncampbell123:
I fixed the NULL vs 0 warnings in the latest commit.

Wengier:
@joncampbell123 I agree that comparing integers to NULL can cause warnings so it can be fixed for cleaner code.

PR #1622, opened 2020-06-03
Implement IOCTL functions "set device parameters" and "set serial number"

Wengier:
I noticed that IOCTL functions "set device parameters" and "set volume serial number" were not implemented in the code, so I made efforts to implement them. The function "set volume serial number" is tested to work on all FAT drives. And the function "set device parameters" seems to work so far during my limited testing.

Also fixed an issue with the built-in LOADFIX command.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,8 +1,13 @@
 0.83.3
   - Added COUNTRY command to set country code for country-
-    specific date and time formats. (Wengier)
+    specific date and time formats. For example, the command
+    "COUNTRY 61" sets the country code to 61 (International
+    English) which use the DD-MM-YYYY date format instead of
+    the default (U.S.) MM-DD-YYYY date format. (Wengier)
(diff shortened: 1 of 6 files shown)


joncampbell123:
Nice!

Allofich:
There are a few places where 16-bit values are being explicitly read and assigned to 8-bit variables.

```
bpb.v.BPB_NumFATs = mem_readw(ptr+0xc);
bpb.v.BPB_Media = mem_readw(ptr+0x11);
bpb.v.BPB_SecPerClus = mem_readw(ptr+9);
```

These create data conversion warnings in the Visual Studio build, which have been common in this project because of things like assigning a `BitU` to a smaller type, but I haven't seen values being explicitly assigned to smaller types like this. Are those lines correct? If so, we can just silence the warnings by making explicit casts.

Wengier:
I have fixed the issue in both functions (40h and 60h) in my pull request #1629.

PR #1629, opened 2020-06-04
Re-port the improved save state feature with save slot support

Wengier:
As we already know the current save state feature in DOSBox-X is extremely experimental. It only supports save and load states with no additional features, and even that does not work most of the time. So I decided to port the more recent save state feature to DOSBox-X instead. While it is still not perfect, it is certainly more developed, and it supports up to 10 save slots to select from, which is apparently very useful.

The unmodified patch only supports using the shortcuts to save state, load state, and switch to the previous or the next save slot (no GUI menu support because it was for original DOSBox), but I improved it by adding GUI menus so that you can use the menu to save and load states as well as selecting a save slot (from 1 to 10) via the menu.

While the new save state feature does change a lot of files, changes in most of the files are simply added to the end, without touching the original contents. As a result, I don't think the new code will affect the existing code for the most part.

Also, the code used by the old save state and the new save state feature are very different, so they can in fact work independently from each other. For this reason I have kept the old save state feature as is for now, but I can remove it if it is deemed no longer necessary.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,9 +1,12 @@
 0.83.3
+  - Re-ported the save state feature for saving and loading
+    states with support for up to 10 save slots. They can be
+    selected from the "Capture" menu. (Wengier)
   - Added COUNTRY command to set country code for country-
     specific date and time formats. For example, the command
     "COUNTRY 61" sets the country code to 61 (International
-    English) which use the DD-MM-YYYY date format instead of
-    the default (U.S.) MM-DD-YYYY date format. (Wengier)
+    English) which uses the DD-MM-YYYY date format instead
+    of the default (U.S.) MM-DD-YYYY date format. (Wengier)
   - DOS 440Dh IOCTL function 67h (get access flag) added to
     allow FDISK.EXE to determine FAT filesystem type. Also
     implemented function 40h (set device parameters) and
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,7 @@
 0.83.3
-  - Re-ported the save state feature for saving and loading
-    states with support for up to 10 save slots. They can be
-    selected from the "Capture" menu. (Wengier)
+  - Re-ported and improved the save state feature for saving
+    and loading states with support for up to 10 save slots.
+    They can be selected from the "Capture" menu. (Wengier)
   - Added COUNTRY command to set country code for country-
     specific date and time formats. For example, the command
     "COUNTRY 61" sets the country code to 61 (International
--- a/include/cpu.h
+++ b/include/cpu.h
@@ -438,6 +438,10 @@ public:
         desc.Load((PhysPt)(table_base+selector));
         return true;
     }
+	
+	virtual void SaveState( std::ostream& stream );
+	virtual void LoadState( std::istream& stream );
+
 
 protected:
     PhysPt table_base;
@@ -489,6 +493,9 @@ public:
 		ldt_value=value;
 		return true;
 	}
+
+	virtual void SaveState( std::ostream& stream );
+	virtual void LoadState( std::istream& stream );
 
 private:
 	PhysPt ldt_base;
--- a/include/dma.h
+++ b/include/dma.h
@@ -131,6 +131,9 @@ public:
 	}
 	Bitu Read(Bitu want, Bit8u * buffer);
 	Bitu Write(Bitu want, Bit8u * buffer);
+
+	void SaveState( std::ostream& stream );
+	void LoadState( std::istream& stream );
 };
 
 class DmaController {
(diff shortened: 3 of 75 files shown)


rderooy:
Considering the old save-state feature is so limited, I would vote to remove it.

joncampbell123:
I have one request: Until save states are stable the menu should make it clear the save/load state system is experimental.

DOSBox-X has it's own load/save state code, which also reads/writes a ZIP file using code I wrote, though not listed in the menu for that reason. If you make the load/save state code better than mine I'll remove my own experimental hackery.

joncampbell123:
I went through the code to see if I can remove full core, and now I remember why I haven't removed it yet.

Two parts of the code rely on Full core: The recursive pagefault code and the dynamic core rely on it as a fallback.

joncampbell123:
Hold on, I'm fixing load/save state issues I see in VGA memory:

- It uses vga.mem.linear_orgptr. Why? Use vga.mem.linear, which is the video memory base (aligned).
- It ASSUMES that VGA memory is at least 512KB, which is WRONG for non-SVGA cards.

joncampbell123:
Ken's Labyrinth seems to work with your load/save support!

joncampbell123:
The code as written can enter PIC events with a NULL handler if it does not recognize the handler function on save state. The latest commit checks for that case and prints a warning instead of calling a NULL pointer, which makes load/save work without crashing in PC-98 mode.

joncampbell123:
Load/save seems to work in general but it does seem to cause subtle state corruption issues after loading. Wolfenstein 3D will crash if you exit a level after loading state. It's not XMS or EMS related, I checked. Xargon after load state will randomly stop playing music after you die, sound effects are wrong, and some dialogs are empty on display.

joncampbell123:
Duke Nukum seems to work fine with load/save.

joncampbell123:
Heretic doesn't handle it well at all.

joncampbell123:
Prehistorik seems to work OK.

joncampbell123:
Electroman generally works... don't use load/save state during the title screen.

joncampbell123:
Jazz Jackrabbit seems to work OK.

joncampbell123:
Blood doesn't handle it well. src\resource.cpp(479): Error seeking to resource!

joncampbell123:
Ecstatica seems to handle it OK.

joncampbell123:
Pinball dreams seems to handle it OK.

joncampbell123:
Quake isn't hard crashing anymore but errors out with another message suggesting something with file handles. Also Gravis Ultrasound IRQs are not being restored on state load.

joncampbell123:
Save/load state seems to bring out every possible way Daggerfall can crash, lol.

joncampbell123:
Catacombs 3D exits unexpectedly when you exit the level IF you loaded state.

joncampbell123:
@Wengier Looking at the code, it saves/restores file handles and states and file positions but I don't think they restore the file behind each file handle. Perhaps that is what is confusing the games. These tests were done from a folder mounted from a local drive.

joncampbell123:
One more thing to look at (or maybe I can figure it out): If aspect ratio correction is enabled, loading state messes up the aspect ratio.

joncampbell123:
Blues Brothers Jukebox seems to work OK with load/save.

Touhou Project 2 seems to work OK though since the FM board emulation is not yet tied into the load/save state system, loading can leave stuck FM notes.

joncampbell123:
Tyrian 2000 seems to have no issues.

joncampbell123:
Duke Nukem 3D seems to work OK with load/save states.

joncampbell123:
One more thing: Save states currently work by storing a lot of internal state directly into the file. Sound Blaster for example dumps the entire struct as-is to the file. That's fine for use on the same machine but that opens up some issues with long term storage and transfer between systems.

Right away I can see a possible issue with, say, 32-bit DOSBox-X save states crashing 64-bit DOSBox-X. Or between x86 DOSBox-X (little endian) and PowerPC DOSBox-X (big endian).

There needs to be a disclaimer saying that save states are for temporary use and are not intended for long term storage. They can change between DOSBox-X versions and different builds. Make it clear so that some fool doesn't invest his time collecting his save states long term only to find out they don't work anymore except with the specific version that made them.

rderooy:
If the save-state files don't have it, they should probably have some header information such as
dosbox edition:dosbox-x
doxbox version:0.83.3
dosbox arch:x86_64
dosbox hostos:Linux

Even if you don't need it now, better have a bit too much. Always useful for debugging. You can then restrict loading a save-state if something does not match and you know it is incompatible.

Wengier:
@joncampbell123 Much of the code in the save state patch that storing the internal state in various files was ported directly with no or only small changes. I did however fix some other code in it to make the code compile and run on DOSBox-X because patch was originally targeted for DOSBox which may have different structures. This code should certainly works better than the old save state feature, but I fully agree that it can be further improved to work even better in DOSBox-X...

Wengier:
@rderooy The version information included in my latest pull request will look like:

DOSBox-X 0.83.3 (SDL2)
Windows 64-bit build
Jun 3, 2020 10:56:46am

BTW: 0.83.3 is always the DOSBox-X version, not DOSBox version. Even DOSBox 0.75 is not ready yet.

PR #1630, opened 2020-06-04
Display program name of the save state etc

Wengier:
The new save state feature I ported earlier will display if a save slot is empty in the menu, but it did not display the program name of a save slot. Thus I improved the feature so that the program name will be displayed if it is not empty. With this it becomes very easy to see which program each save slot was saved for. 

I also did some other small cleanups to the feature, such as applying the new program name to DOSBox-X.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,10 +1,13 @@
 0.83.3
-  - Save/load state now saves PC-98 FM interrupt state so
-    that reloading state does not cause hung music.
-  - Save/load state now supports Sound Blaster Goldplay mode.
-  - Save/load state fixed not to crash when used within a
-    guest operting system (when the DOS kernel has been
-    shut down using the BOOT command).
+  - Improvements and fixes to the save/load state feature:
+    It now supports Sound Blaster Goldplay mode;
+    It now saves PC-98 FM interrupt state so that reloading
+    state does not cause hung music;
+    It is also fixed not to crash when used within a guest
+    operting system (when the DOS kernel has been shut down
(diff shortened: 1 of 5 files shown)


joncampbell123:
One thing that should be made clear by next release:

Save states currently work by storing a lot of internal state directly into the file. Sound Blaster for example dumps the entire struct as-is to the file. That's fine for use on the same machine but that opens up some issues with long term storage and transfer between systems.

Right away I can see a possible issue with, say, 32-bit DOSBox-X save states crashing 64-bit DOSBox-X. Or between x86 DOSBox-X (little endian) and PowerPC DOSBox-X (big endian).

There needs to be a disclaimer saying that save states are for temporary use and are not intended for long term storage. They can change between DOSBox-X versions and different builds. Make it clear so that some fool doesn't invest his time collecting his save states long term only to find out they don't work anymore except with the specific version that made them.

Wengier:
@joncampbell123 While I did fix quite a few things in the save state patch during porting to make the code compile and work better on DOSBox-X, much of the code in the patch that storing the internal state in various files was ported directly with no or only small changes. Since the patch was originally targeted for DOSBox which may have different structures, and also it involved changes to a lot of files, I did put quite a bit of efforts to port it, but I fully agree that it can be further improved to work even better in DOSBox-X.

By the way, perhaps we can already delete the old save state feature, which is not very usable anyway.

joncampbell123:
I will delete the old save state feature, yes. At this point the ported save/load state code is better.

PR #1631, opened 2020-06-04
Save & verify DOSBox-X version/build for save states

Wengier:
sAs discussed earlier, states saved by DOSBox-X in one platform may not work in another, so I improved the save states feature to save the DOSBox-X version and build. An example of DOSBox-X_VERSION file to be saved will be:

```
DOSBox-X 0.83.3 (SDL2)
Windows 64-bit build
Jun 3, 2020 10:56:46am
```

When you load a saved state, DOSBox-X will verify if the current environment matches this (everything except the build date/time). If not, it may refuse to load the state. But you can always force load the state by enabling the "Force load state" menu option.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.83.3
+  - Added GUI menu option under "DOS" to change the long
+    filename setting (enable, disable, or auto). (Wengier)
   - Improvements and fixes to the save/load state feature:
     It now supports Sound Blaster Goldplay mode;
     It now saves PC-98 FM interrupt state so that reloading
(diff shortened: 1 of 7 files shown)


Wengier:
@joncampbell123 I also have a question: What is the difference between "Toggle Menu" and "TogMenu" in the Windows SDL1 build? Both appear in the mapper editor, but in different places. I am not entirely sure if they are duplicates. Only "TogMenu" will appear in other builds on the other hand.

joncampbell123:
@Wengier Let me check the code, it might be a simple mistake.

PR #1634, opened 2020-06-06
Windows installer: major improvements

Wengier:
I noticed that official releases of DOSBox-X always came with quite a few Windows builds (Visual Studio x86/x64/ARM/ARM64 SDL1/SDL2 builds and MinGW32/64 SDL1/SDL2/lowend/SDLDraw builds) in different sub-directories, but there is never a Windows installer release. Apparently, this is not very friendly for Windows users, and non-technical users may not know how to start the program at all in this case. I also noticed that there is a directory called "windows-installer" in the repository, but the code in there was very preliminary. So I decided to go ahead to make significant improvements to it.

The new Windows installer will be a unified installer for both 32-bit and 64-bit Windows, and it will be much more usable than the previous one. All different Windows builds (both VS and MinGW builds) are included in the installer, but the installer will ask the user to select a **default** build from these builds to be installed as the dosbox-x.exe program in the DOSBox-X directory (e.g. "C:\DOSBox-X"). And if the user selects a full installation, then all different Windows builds will be copied to the sub-directories as well. The files installed to C:\DOSBox-X include the following (compact installation):

changelog.txt
COPYING.txt
dosbox-x.exe
dosbox-x.reference.conf
FREECG98.BMP
README.txt
shaders (subdirectory)

For full installation, the various Windows builds will be copied to sub-directories in the Win32_builds and Win64_builds directories. Still, 64-bit Windows builds will only be copied if the running Windows is 64-bit.

The file dosbox-x.reference.conf will also be automatically copied to dosbox-x.conf for use with DOSBox-X by the installer. If the dosbox-x.conf file already exists in the directory, the installer will ask the user whether to overwrite the existing config file.

A Windows installer for DOSBox-X 0.83.2 I built is available from:

http://cndos.fam.cx/dosware/DOSBox-X-setup.exe

It should already work as is. Tested in Windows 10, 64-bit.

To build the Windows installer from source, the pre-compiled Windows binaries in sub-directories named Win32_builds (32-bit Windows builds) and Win64_builds (64-bit Windows builds) are required (only dosbox-x.exe is needed). All of them are already included in the official Windows releases of DOSBox-X. The structure is like the following:

Under "Win32_builds":
ARM
mingw
mingw-lowend
mingw-sdl2
mingw-sdldraw
Win32

Under "Win64_builds":
ARM64
mingw
mingw-lowend
mingw-sdl2
mingw-sdldraw
x64

Apart from Inno Setup, all other files need to build the installer (e.g. readme.txt, dosbox-x.reference.conf) are automatically obtained from the repository.


Wengier:
It should already work, but I think I will further simplify the building process tomorrow.

Wengier:
@joncampbell123 May I ask which path the following zip files are located when you use the build scripts to create them in order to make releases? I want to make the build script of the Windows installer automatically extract dosbox-x.exe from these zip files and then automatically compile the source files to the Windows installer executable, so I need to know the exact location of these zip files:

dosbox-x-windows-\*-windows.zip (e.g. dosbox-x-windows-20200531-220809-windows.zip)
dosbox-x-mingw-win32-\*.zip (e.g. dosbox-x-mingw-win32-20200531151931.zip)
dosbox-x-mingw-win64-\*.zip (e.g. dosbox-x-mingw-win64-20200531160500.zip)

joncampbell123:
You may ask, indeed.

windows-*-windows.zip is compiled from VS2019 which places it in the bin directory.
The make-windows-release.pl Perl script is run from the root of the directory (of the repository) to make the ZIP archive

MinGW builds are run from the parent directory of the DOSBox-X source directory using the make-mingw.sh script. The script runs through each configuration, copying off each result into another directory before zipping up the completed bundle. A copy is in the build-scripts directory.

Wengier:
@joncampbell123 Thanks for the information. I see from the make-windows-release.pl script there is ``my $subdir="release/windows"``. So can I assume the zip files can be found in the following paths?

(repository root)/release/windows/dosbox-x-windows-\*-windows.zip
(repository root)/build-scripts/dosbox-x-mingw-win32-\*.zip
(repository root)/build-scripts/dosbox-x-mingw-win64-\*.zip

joncampbell123:
./make-windows-release.pl makes only (repository root)/release/windows/dosbox-x-windows-*-windows.zip

the make-mingw.sh script is run from the MinGW shell in the parent directory (above where you cloned DOSBox-X) and it clears and resets the repository for each variant it builds. The ZIP files end up in the same directory (parent of DOSBox-X repository directory)

joncampbell123:
Mac OS X builds are made the same way as MinGW on Mac OS X.

A Windows XP VM is used the same way for the MinGW HX-DOS builds. Windows XP is used because ironically MinGW runs faster in a virtual machine under XP than natively on the host Windows 10 filesystem.

Wengier:
@joncampbell123 I see. So I will make the build script of the Windows installer using the following paths for the zip files (containing both VS and MingW builds) then:

(repository root)/release/windows/dosbox-x-windows-*-windows.zip
(repository root)/../dosbox-x-mingw-win32-*.zip
(repository root)/../dosbox-x-mingw-win64-*.zip

joncampbell123:
Sounds about right.

Wengier:
@joncampbell123 I have already rewrote the building script (as well as adding all required building tools) for the installer to make the building process fully automatic, as long as the zip files containing VS and MinGW builds as mentioned earlier are already in the said directories.

Wengier:
@joncampbell123 I think the Windows installer is already pretty usable by now. Therefore, if there is no major issue found, I would strongly recommend to include the installer in the next release, so that Windows users will be able to get started with DOSBox-X more easily.

PR #1639, opened 2020-06-07
Support for cues files with MP3/OGG/WAV/FLAC audio tracks

Wengier:
I noticed that both DOSBox ECE and DOSBox-staging support mounting cue files with compressed audio tracks, but DOSBox-X does not yet support this feature. So I decided to port this feature from DOSBox-staging to DOSBox-X so that it is now possible to mount cue files with MP3/OGG/WAV/FLAC audio tracks in DOSBox-X too. Tested to work with both Visual Studio and MinGW.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.83.3
+  - Support for mounting cues files with MP3/OGG/WAV/FLAC
+    compressed audio tracks. (Wengier) 
   - Rewrote the Windows installer for DOSBox-X, as well as
     the building script for the installer. All required
     build tools including the Inno Setup Compiler are now
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -29,6 +29,10 @@
 typedef void (*MIXER_MixHandler)(Bit8u * sampdate,Bit32u len);
 typedef void (*MIXER_Handler)(Bitu len);
 
+template <class T> T clamp(const T& n, const T& lower, const T& upper) {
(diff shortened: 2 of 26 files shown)


joncampbell123:
Looks good! I'll do additional commits to handle any compile failures.

Will a future pull request update the VS2019 project files to enable this?

joncampbell123:
It looks like cdrom_image support just includes the decoders library directly. That's kind of messy, it might be better to compile the decoders library into an .a file and then link to the rest of the program.

Wengier:
@joncampbell123 Yes, for now it includes the decoder library directly to avoid the need of updating the Visual Studio project files (I did not want to modify these files directly since it is possible to cause unexpected issues). But if you prefer, you can surely update the Visual Studio project files to include the library files later.

joncampbell123:
A follow-up on Track 1 repeating the first sector in Windows 95.

The reason it repeats is because Windows 95 periodically sends a command to read sector 16 (command 0x28) even while playing a CD. That read is repositioning playback back to the start.

If I start CD audio playback then use CTRL+ALT+DEL to bring up the Task Manager window (which runs as Win16 and halts all 32-bit processes), CD audio playback proceeds normally.

joncampbell123:
The latest commit makes CD reads fail if a data read is carried out against an audio track. Windows 95 periodically attempts to read data sector 16 in the background for some reason, which according to SCSI MMC standards should immediately fail if playing an audio CD and should fail against non-data sectors. Track 1 can play normally now.

PR #1641, opened 2020-06-08
Allow key mapping to mouse buttons

Wengier:
DOSBox-X ECE has this feature, but DOSBox-X did not, so I decided to add this feature to DOSBox-X too. As a result, the events for mouse buttons (left, middle, right) will now appear in the mapper and the users can map them to keys if they wish.

Also fixed the path to the default FluidSynth font file in Windows platform.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 0.83.3
-  - Support for mounting cues files with MP3/OGG/WAV/FLAC
+  - Mouse buttons (left, middle, right) can now be mapped
+    to keys in the keyboard mapper. (Wengier)
+  - Support for mounting .cue files with MP3/OGG/WAV/FLAC
     compressed audio tracks. (Wengier) 
   - Rewrote the Windows installer for DOSBox-X, as well as
     the building script for the installer. All required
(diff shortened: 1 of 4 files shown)


joncampbell123:
Nice!

I'm investigating a bug with CD audio and WAV files. I made three WAV files (16-bit stereo) from MP3s using FFMPEG. One of them loops the first second when played in DOSBox-X. The other two are fine.

Wengier:
> Nice!
> 
> I'm investigating a bug with CD audio and WAV files. I made three WAV files (16-bit stereo) from MP3s using FFMPEG. One of them loops the first second repeatedly when played in DOSBox-X as a CD audio track. The other two are fine.
> 
> The one that loops is 6 minutes 46 seconds, in case it's a bug with integer overflow and duration.
> 
> Edit: No, it only happens to track 1, regardless of WAV file. Hm...

Interesting. I used SJGPlay for DOS to test the CD tracks when I ported the feature yesterday, and did not notice this. I tested 3 audio tracks, one MP3, one OGG, and one WAV.

joncampbell123:
It might be a problem with IDE emulation. It only happens with CDPLAYER.EXE in Windows 95 through IDE emulation.

PR #1656, opened 2020-06-14
Improvements to overlay drives

Wengier:
I have made further improvements to the recently supported overlay drives. For example, long filename support should now work more reliably on such drives than before, also fixed some other issues I found.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,8 +5,9 @@
     function that FAT filesystem mounting uses.
   - Added suppprt for mounting overlay drives using MOUNT
     command with "-t overlay" option. Ported from DOSBox
-    SVN and adopted for DOSBox-X, this feature allows the
-    user to redirect new and changed files to a different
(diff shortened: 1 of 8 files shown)


Wengier:
Shift-JIS characters (Kana, Kanji, etc) are now supported for overlay drives in PC-98 mode, by using wide characters instead of regular ones.

joncampbell123:
Where is host_cnv_use_wchar defined?

Wengier:
@joncampbell123 host_cnv_use_wchar is defined in line 56 of drive_local.cpp. This definition was already there for quite some time I think.

Wengier:
@rderooy The new message does show up in Windows, but I when I tried on Linux it did not show up indeed. I will submit a new pull request today including a fix for this Linux-specific issue.

joncampbell123:
@Wengier Right. My mistake. For some reason grep didn't find it at all.

Issue #591, opened 2018-03-09
Need a script to automate MinGW-based HX DOS extender releases


joncampbell123:
Needs to include compiled binary, HDPMI32, HX DOS and HXGUI binaries, and a copy of MSVCRT.DLL known to work taken from an older Windows system.

yksoft1:
You may also need CRTDLL.DLL.
(There is a modified HX DOS distribution that contains an OpenGL driver for Voodoo, but I don't think Dosbox-X can be used practically on a PC that still compatible with Voodoo card).

joncampbell123:
The copy of MSVCRT.DLL I used from Windows 98 did not need CRTDLL.DLL, neither did the one from Windows XP.

Wengier:
@joncampbell123 I have already made the HX-DOS build script fully automated by including all **required** HX DOS Extender files and MSVCRT.DLL (in /build-scripts/mingw/hxdos) in my latest pull request #1666. The updated build script (make-mingw-hx-dos.sh) will also automatically run PESTUB.EXE to patch the HX-DOS build's dosbox-x.exe file before zipping it so that the program DPMILD32.EXE is no longer needed for running it in DOS.

Therefore it becomes extremely easy to run the HX-DOS build: simply run dosbox-x.exe directly in a real DOS system after unzipping, as the package has already taken care of everything needed.

BTW, you may want to slightly modify the directories defined by environment variables $copydir and $hxdir in make-mingw-hx-dos.sh in case the defined one does not match the actual build environment on your computer. They are defined as follows (as in the repository):

copydir="build-scripts/mingw/dosbox-x-mingw-hx-dos"
hxdir="build-scripts/mingw/hxdos"

Wengier:
@joncampbell123 I noticed that DOSBox-X 0.83.3 is released, but HX DOS binaries are not yet included in the HX-DOS version. Is there any issue with the current build script? Please let me know if you find any.

joncampbell123:
Now that I think about it, I have an older version of the script one level up from the source directory that I ran. Let me try again.

joncampbell123:
It looks like the script needs a little more work, which can then be done in the next release.

Wengier:
[...]

PR #1666, opened 2020-06-16
Cleanups to README.md file

Wengier:
The content of the README.md file will appear when you open the DOSBox-X GitHub page, but I found that the majority of this content is very technical, such as compiling, debugging and some technical details of the source code.  Apparently such information is primarily targeted at developers, not end-users, so that they will not appear very friendly for end-users in general. Thus, I decided to make a cleanup to README.md file, adding contents such as "Notable features in DOSBox-X" in the page, and moved technical contents (compiling, debugging, other technical details of the source code) to separate files named README.source-code-description and README.debugger. README.md will then focus more on general information about DOSBox-X, whereas the new pages are all directly linked from the README.md file for those who are indeed interested in such technical information.


Wengier:
I have now made the HX-DOS build script fully automated by including all required HX DOS Extender files and MSVCRT.DLL (in /build-scripts/mingw/hxdos). The updated build script (make-mingw-hx-dos.sh) will also automatically run PESTUB.EXE to patch the HX-DOS build's dosbox-x.exe file before zipping it so that the program DPMILD32.EXE is no longer needed for running it in DOS. This resolves Issue #591.

PR #1678, opened 2020-06-21
Update FLAC/MP3/WAV decoders to latest version

Wengier:
As pointed out by @dreamer recently, at this time the FLAC/MP3/WAV decoders in /src/libs/decoders were no longer the latest version, which may miss opus support and other fixes. So I now ported the code from the most recent development branch from DOSBox-staging to update the current code in /src/libs/decoders.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,7 @@
 0.83.3
+  - DOSBox-X will try to automatically run the program
+    with LOADFIX if the error message "Packed file is
+    corrupt" is detected. (Wengier)
   - MOUNT command can now mount a drive to the UNC
     root path in the form \\ComputerName\SharedFolder
     on Windows systems (Wengier)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,7 @@
 0.83.3
+  - MOUNT command can now mount a drive to the UNC
+    root path in the form \\ComputerName\SharedFolder
+    on Windows systems (Wengier)
   - Improved the HX-DOS build package to make it fully
     automated: all required HX DOS Extender files will
     be included in the package, and dosbox-x.exe can
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -87,6 +87,7 @@ bool enable_dbcs_tables = true;
 bool enable_filenamechar = true;
 bool enable_share_exe_fake = true;
 bool rsize = false;
+bool packerr = false;
 int dos_initial_hma_free = 34*1024;
 int dos_sda_size = 0x560;
 int dos_clipboard_device_access;
@@ -97,6 +98,7 @@ int maxfcb=100;
 int maxdrive=1;
 int enablelfn=-1;
 bool uselfn;
+extern bool infix;
 extern bool int15_wait_force_unmask_irq;
 extern bool startup_state_numlock;
 
@@ -560,6 +562,7 @@ static Bitu DOS_21Handler(void) {
     char name2[DOSNAMEBUF+2+DOS_NAMELENGTH_ASCII];
     
     static Bitu time_start = 0; //For emulating temporary time changes.
+    if (reg_ah!=0x4c) packerr=false;
     switch (reg_ah) {
         case 0x00:      /* Terminate Program */
             /* HACK for demoscene prod parties/1995/wired95/surprisecode/w95spcod.zip/WINNERS/SURP-KLF
@@ -1420,6 +1423,7 @@ static Bitu DOS_21Handler(void) {
                 }
 
                 MEM_BlockRead(SegPhys(ds)+reg_dx,dos_copybuf,towrite);
+                if (reg_bx==2&&towrite==22&&!strncmp((char *)dos_copybuf,"Packed file is corrupt",towrite)) packerr=true;
                 if (DOS_WriteFile(reg_bx,dos_copybuf,&towrite)) {
                     reg_ax=towrite;
                     CALLBACK_SCF(false);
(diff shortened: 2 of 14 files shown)


kcgen:
hi @Wengier, 

These files are all separately authored, so would be a good opportunity to give the  `--author` feature a go if you want!   You would have to unstage the commit, then recommit them in groups with the author flag (Ie, David Reid for all of the dr_* files, Sean Berett for stb, Yan Collet for xxhash, and myself for mp3.cpp and the seek tables). 

+52 −8  src/libs/decoders/dr_flac.h
+132 −43  src/libs/decoders/dr_mp3.h
+130 −64  src/libs/decoders/dr_wav.h
+30 −29  src/libs/decoders/mp3.cpp
+5 −5  src/libs/decoders/mp3_seek_table.h
+112 −110  src/libs/decoders/stb.h
+248 −171  src/libs/decoders/stb_vorbis.h
+496 −202  src/libs/decoders/xxh3.h
+232 −149  src/libs/decoders/xxhash.h

Happy to help. Regards!

Wengier:
@kcgen Thanks! It's my first day to commit files using the command-line tool to this repository. The earlier commit was done in the morning before I played with the command-line tool, so I decided to revert that commit with “git revert” and then re-update the files by groups using the —author option. It seems to be a success. Thanks again!

kcgen:
Good stuff!

Side comment, I recently installed an addon in vscode that shows "blame" history for the current line of code the cursor is on.. and it's amazing seeing some lines have remained intact for 18 years, and it also shows the author's commit message.  So retaining the authorship (and absorbing commits and original comments when possible) all feeds into this rich history.  

@dreamer helped introduce me to these very enriching sideroads in the world of git :-)

Wengier:
I have now implemented the algorithm to automatically re-run the program with LOADFIX if the error "Packed file is corrupted" error message is detected in this pull request. This solves Issue #1681.

Wengier:
This pull request also fixes the issue that MOUNT command cannot mount the UNC root path in the form \\ComputerName\SharedFolder to a drive. It did work with UNC paths like \\ComputerName\SharedFolder\directory, but not the root path. Now UNC root paths can be mounted too.

kcgen:
> I have now implemented the algorithm to automatically re-run the program with LOADFIX if the "Packed file is corrupted" error message is detected in this pull request. This solves Issue #7 and #1681.

@Wengier  - this is a great nice quality-of-life improvement! @dreamer, we should put this in our queue. Using `LOADFIX` is ones of those subtleties that experienced dosbox users know to try when older EGA / CGA games fail, but new or casual users would definitely be stumped unless they've thorougly read the documentation.

Wengier:
@kcgen I am currently working on the porting of long filename feature as discussed with @dreamer earlier. It will need some time to make it fully working as in DOSBox-X right now.

kcgen:
Thanks for the heads up @Wengier!

dreamer:
@Wengier I think I was more interested in other parts of LFN patch (ability to paste in terminal, mouse work and correct DOS file shortening), but I think I saw a long list of productivity software that was helped by this feature (no games though), so we might include it as well - but we will need to be extra careful not to re-introduce security bugs in dosbox.

Over last few months, we were gradually fixing code, that was susceptible to the buffer overflows due to limits on file name / path name length (we're still not finished btw) - most of those fixes were a precaution only (but there were few real buffer overflows). What are the new limits on the path name, file name, and extension name length after introducing long filename support? Does it affect negatively any older games? There's plenty of code dosbox-staging inherited from DOSBox that depend on max path length == 80, max file name length == 12, max extension name length == 3.

Also - I don't know if it helps, but your work was imported to branch `vogons/lfn-r4329-2`: [093ef](https://github.com/dosbox-staging/dosbox-staging/commit/093ef992fa2310d74e9e517bd72ab440453f9e7e)

Wengier:
@dreamer I see, although I was indeed under the impression that you did specifically ask for LFN support/patch previously. I would not consider things like mouse work to be (strictly) part of the long file name feature. For example, even though LFN support, ability to paste in terminal, and mouse work have all been incorporated to DOSBox-X in recent months, they were not all incorporated at the same time. I incorporated things like the mouse copy and paste support separately because I do consider them different (they were in different Vogous topics too).

I don't know if there are specific games that are helped by the feature (I don't play DOS games a lot), but I am quite aware that DOS game emulators like NESTICLE and ZSNES do make use of long file name support whenever possible, including when running in DOSBox-X with the feature enabled. In such case, long file names will be shown for directories and names of the game ROMs, instead of something like PROGRA~1 and SOMEGA~1.ROM. 

As a side note, the code for features like LFN support in DOSBox-X is newer than the code in my own fork earlier, so there are later improvements over the code in branch vogons/lfn-r4329-2 that you mentioned, most notably to enable the feature for FAT drives.

As for the buffer overflow handling, I think the max path length will remain the same if the LFN feature is not enabled. Otherwise, the LFN path length will be used for LFN calls (only), and the name limits for non-LFN calls (used by older games etc) will not be affected. The maximum limit of a long filename is 255 characters, as can be seen from the following page: https://jpsoft.com/help/filenames.htm, although it may be increased up to 4 to allow the use of quotes.

Wengier:
@kcgen I noticed that the game Thexder as mentioned in #1681 does work in DOSBox-staging with the default setting even without LOADFIX. So if you have noticed any other specific program or game that still reports "Packed file is corrupt" in DOSBox-staging, please let me know so that I can also implement the algorithm in DOSBox-staging and verify it is indeed implemented properly there.

PR #1690, opened 2020-06-24
Cleanup Linux build instructions

rderooy:
- The no-avcodec build option no longer exists
- Add Linux SDL2 build option
- Mention make install
- Mention RPM build option


joncampbell123:
build-debug-no-avcodec no longer exists because configure.ac now properly auto-detects whether FFMPEG is present. If it's not present, then it doesn't use it.

PR #1693, opened 2020-06-25
Update (c) year for dosbox and dosbox-x assets and update FSF address

rderooy:
I did not touch copyright year for other assets


rderooy:
I'm done, there are some old FSF addresses inside contrib and SDL, but I don't want to touch those.

Issue #1675, opened 2020-06-20
Lutris support - executable as CLI argument not working


gombosg:
**Describe the bug**
Lutris runs Dosbox runners as `dosbox %s` where `%s` is the game executable. Vanilla Dosbox supports this, it just runs the game executable from that folder without any config whatsoever.

**To Reproduce**
Dosbox-X doesn't support that e.g. if I do
`./dosbox-x /gamedir/GAME.BAT`
then it says
`WARNING: Unknown option gamedir/GAME.BAT (first parsing stage)`
(EDIT: same if file name is given in quoted)

**Expected behavior**
It should start the game.

**Environment (please complete the following information):**
 - Fedora 32
 - Fresh compile from master 8cf8188
 - No config

**Additional context**
This is supposed to be working judging from the manual page in the `intro` command and looking at how vanilla Dosbox works.

Wengier:
I have already fixed the issues in my pull request #1695.

gombosg:
Thanks, works, closing!

PR #1695, opened 2020-06-26
Fixed command-line program handling (Issue #1675)

Wengier:
As noticed in Issue #1675, there are some bugs in DOSBox-X's command-line parameter handling which cause command-line like "./dosbox-x /dir/app.bat" to not work as expected. Thus I fixed the issues so that such an argument should now work in both Windows and Linux.

Also slightly updated the documentation to mention the parameter "--triplet arm64-windows" when using vcpkg to install library for ARM64 build.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.83.3
+  - Fixed command-line like "./dosbox-x /dir/app.bat"
+    not able to launch the specified program (Wengier)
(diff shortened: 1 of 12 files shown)


Wengier:
I have improved some comments and some content in the Windows installer documentation. The default setting for clip_key_modifier is now “shift” instead of “disabled”.

PR #1696, opened 2020-06-28
Fixed DIR /P bug

Wengier:
I have noticed a bug in DIR /P handling, so this pull request fixes the bug.


Wengier:
By the way, the official name for Mac OS X Sierra 10.12 and later is "macOS", so I updated the documentation to mention this name too.

PR #1700, opened 2020-06-30
IMGMAKE help message fixes

Wengier:
There are now some new options to the IMGMAKE command, but because of some examples listed in the bottom the help messages can no longer be shown in one screen, so “IMGMAKE -?” will miss a few lines on the top. So I decided to add an “-examples” option to show the examples (with new examples), which will no longer be shown by the -? option. As a result, “IMGMAKE -?” will now show all the help message in one screen instead of only showing part of them in the screen. Also made some small fixes to the help messages displayed by it.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,15 +1,16 @@
 0.83.3
   - IMGMAKE can now generate FAT12, FAT16, and FAT32
-    filesystems.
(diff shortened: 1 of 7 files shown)


Wengier:
It now fixes the issue to create HDD images larger than 2GB in Visual Studio builds (Issue #1701).

Issue #1709, opened 2020-07-02
Copy and Paste menu item


rderooy:
**Is your feature request related to a problem? Please describe.**
The current copy-and-paste support is not the most intuitive.

As such I want to suggest a Edit menu. Either directly in the top bar, or as a sub-menu in Main. With support for Mark, select-all and paste similar to that of the Windows Command Prompt.

For this you would not need to change the clip_key_modifier.

**Additional context**
![Capture](https://user-images.githubusercontent.com/17701098/86379548-3f7f3a00-bc8b-11ea-9b8e-d9c6447bccd8.JPG)

Wengier:
@rderooy I think the Edit menu is not really required for copying and pasting in Windows Command Prompt. In case you don't know, you can enable "QuickEdit mode" in "Command Prompt Properties", and then you will be able to directly use the mouse to copy and paste without the need of the Edit menu. It will be very similar to the "clip_key_modifier=none" setting as in DOSBox-X. But an additional Edit menu can be considered of course.

rderooy:
@Wengier I know, was just using it as an example. And I think it may be easier to discover the Edit menu than setting clip_key_modifier in the config file.

Wengier:
I think it is probably easier to make a toggle menu item so that when enabled, then no keyboard modifier is needed for the copy and paste operation.

Wengier:
@rderooy I have already done this in my latest pull request.

Issue #1683, opened 2020-06-23
Hot keys and keys assigned to the mapper do not work


maximus105:
**Describe the bug**
In mode "cga_mono" or "cga" Host + R key does not work. When you press this combination, DOSBox-X freezes. Also, after assigning a Mono CGA Bright key and saving it, after restarting DOSBox-X, the assigned key combination does not work (it simply is not saved in mapper). Other commands in DOSBox-X assigned to mapper may also not work.

**Environment (please complete the following information):**
 - Windows 10 x64 2004
 - DOSBox-X 0.83.2 x64 SDL2

[dosbox]
machine = cga_mono
or
machine = cga

[render]
frameskip = 0
alt render = false
aspect = true
char9 = true
doublescan = false
scaler = advinterp2x forced
xbrz slice = 16
xbrz fixed scale factor = 0
xbrz max scale factor  = 0
autofit = true
monochrome_pal = white bright

[vsync]
vsyncmode = on
vsyncrate = 60

[cpu]
cycles = 240

Wengier:
@maximus105 I just did a quick check. It is not that the keyboard mapper does not work in this case (it does work), but there are  apparently some bugs with the graphics reset code for the CGA mode so that when you reset the system the graphics are not being updated in this mode. It may appear like a freeze, but the system itself did not freeze, it is just that the graphics is no longer being updated.

@joncampbell123 You may want to check why the graphics reset code does not work when you try to reboot the guest system (throw int(3)) in CGA mode.

joncampbell123:
@Wengier Will do. I'm currently implementing the requested feature that MinGW and Mac OS X builds enable SDLnet so that localhost null modem connections are enabled.

By the way, GitHub just sent me an email saying that you created an unauthorized branch in DOSBox-X because of a bug in the website. That would explain the issues I had merging a pull request that day. You're a hacker, congratulations :)

Wengier:
@joncampbell123 ??? GitHub said I created an unauthorized branch in DOSBox-X because of a bug in the website? Honestly, I have no idea what this means. I don't even know what "unauthorized branch" it is talking about. Can you describe it in more details for this? Thanks. By the way, I was testing the Git command features in the last a few days myself indeed to get more familiar with it (I was not so familiar with the command-line tools previously for example), perhaps I did something unexpected during this period that I was not aware myself?

joncampbell123:
[...]

Issue #1707, opened 2020-07-01
[Documentation] How to install


rubyFeedback:
Hello,

I was looking at:

https://github.com/joncampbell123/dosbox-x/blob/master/INSTALL

But it seems as if cmake is used rather than GNU configure
so INSTALL is perhaps outdated.

The bigger issue is that I failed to use cmake for this:

cmake -DCMAKE_INSTALL_PREFIX=/usr .
-- The C compiler identification is GNU 10.1.0
-- The CXX compiler identification is GNU 10.1.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc - works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ - works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at CMakeLists.txt:19 (message):
  Option not implemented

That was when I looked at the INSTALL file but it does not help my use case. :-)

Nice to see dosbox is still sort of alive though!

Wengier:
What operating system (e.g. Windows or Linux) are you using? Are you trying to simply install and run the program or you want to build the program from source code by yourself?

Wengier:
If you just want the latest version of DOSBox-X, you can always get one for your platfrom from the Release page:

https://github.com/joncampbell123/dosbox-x/releases

PR #1710, opened 2020-07-02
Fix Issues #1683, #1709, #1712, #1714

Wengier:
As mentioned in Issue #1683, the keyboard shortcut for Mono CGA Bright did not work after DOSBox-X is exit completely and re-run, so I fixed the issue. As a result, the Mono CGA Bright key should now work properly.

I also fixed Issue #1709 by adding a menu item toggle ("Quick right mouse button copy/paste") to allow quick clipboard copy and paste via the right mouse button without any key modifier.

In addition, according to Issue #1707 the current content of the text file "INSTALL" seems to be somehow outdated and also confusing, so I updated the file to also point out that the latest versions of DOSBox-X can be found at the Releases page, i.e. 
https://github.com/joncampbell123/dosbox-x/releases

I believe these issues (#1683 and #1709) have already been fully fixed, but @joncampbell123 will confirm this of course.


Wengier:
I have further expanded the INSTALL documentation file to briefly explain how to get the latest Windows and Linux packages from the Releases page. I just follow what we have, and nothing else.

Wengier:
This pull request now fixes Issue #1712 also, i.e. to mount a directory with a command-line like "dosbox-x ." at startup.

Wengier:
This pull request is now supposed to fix Issue #1714 too.

PR #1715, opened 2020-07-04
Fix make install on Mac and BSD

SeekingMeaning:
# Description

Replaced `install -D -d` and `install -d` with `mkdir -p`

`-D` flag is not portable to non-GNU systems

N/A

N/A

N/A

**Additional information**

`-p` flag is to recursively create parent directories if they do not exist


rderooy:
This change will break the Linux install, if the directory before it does not exist.

SeekingMeaning:
Okay, I'll change it to `mkdir -p`

Issue #1545, opened 2020-05-12
DOS IOCTL does not support external device drivers


Wengier:
As mentioned in Issue #1217, I found that DOSBox-X's DOS IOCTL functions do not support external device drivers that are accessed by their name, such as the device named "IFS$HLP$" as provided by Windows for Workgroups 3.x or Windows 9x/ME's IFSHLP.SYS. Loading IFSHLP.SYS with the built-in DEVICE command (or another device loader like DEVLOAD) will succeed, but trying to open or communicate with the device name via the IOCTL functions will all fail. The IFSHLP.SYS device driver is useful for both WFW 3.11 and Windows 9x, because without it is impossible to enable network support in WFW 3.11, and it is also a requirement to start Windows 9x's WIN.COM if run from the DOSBox-X shell.

joncampbell123:
Agreed.

Two things need to happen.

One, the internal DOS device registration needs to make a fake device driver header and add it to the linked list visible through the List of Lists. It should be possible to see ALL device drivers considered by DOSBox-X through the List of Lists, not just the fake CON, LPT entries placed there.

Two, DOS device driver handling needs to refer to that device driver linked list when asked to talk to a device by name. Currently it only consults an internal list.

Considering that the callback system allows calls from guest to specific host native code, one callback could be registered per device driver and the fake device driver header interrupt and strategy routines would just contain that callback instruction for it to work regardless.

joncampbell123:
By the way, future device driver emulation should also create a fake CD-ROM device driver entry as well if MSCDEX.EXE emulation is switched on, since on real DOS MSCDEX.EXE talks to a device driver to provide CD-ROM functions.

PR #1726, opened 2020-07-09
Improvements to IMGMAKE command

Wengier:
The IMGMAKE command will show an error if you try to create a disk image file which already exists. I noticed that there is a -force option in DOSBox Optionals to overwrite the existing disk image file even if it already exists. I believe this is a useful addition so I added this to DOSBox-X too. Also, IMGMAKE will now use the default image file name "IMGMAKE.IMG" if no image file name is specified. In addition, I made some improvements to the help message for this command.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,8 +4,9 @@
     and paste via the right mouse button without any
     key modifier that may have been specified by the
     config option "clip_key_modifier". (Wengier)
-  - Fixed mounting disk images with spaces via the
-    "Drive" menu. (Wengier)
+  - Fixed mounting a directory with command-line like
+    "dosbox-x .", and fixed mounting disk images with
+    spaces via the "Drive" menu on Windows. (Wengier)
   - IMGMAKE warning replaced to indicate a general
     incompatibility between MS-DOS/SCANDISK and
     cluster sizes 64KB or larger.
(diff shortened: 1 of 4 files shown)


joncampbell123:
Following up with the build system: I finished the update and my Linux system now has GCC 9.3 by default.

Development from this point on will be done with GCC 9.3, though I will stay at C++11 and keep a minimum off GCC 4.8.5 around for at least a year until nobody uses GCC 4.x anymore.

Wengier:
@joncampbell123 I am glad to hear that. By the way, do you have time to look at Issue #1545? I am sure you are more familiar with the hardware/device driver part than me.

joncampbell123:
That's going to take slow incremental work. The device driver selection code needs to enumerate the driver linked list, and then it's just a matter of constructing the various calls to the driver interrupt routine with the request structures.

Wengier:
@joncampbell123 I see. And thanks for the link with details about DOS device drivers.

joncampbell123:
@Wengier The good news is that I have a few books on my bookshelf that cover MS-DOS device drivers to varying degrees. Including "Advanced Assembly Language" that helped me write my first character device driver way back in the day. It was just a dumb driver that sent written data to the Sound Blaster DSP using DSP command 0x10 "direct write" :)

Wengier:
That is a good thing for sure. I have not actually written a DOS device driver using the assembly language myself yet (I did use the assembly language for some other things). But in any case, once this feature is implemented, it will be become a very useful and unique feature of DOSBox-X - apparently no other DOSBox forks have supported this yet (many of them are very game centric of course).

PR #1727, opened 2020-07-10
Support SHELL= in [config] with built-in 4DOS 8.00 command shell as alternative shell

Wengier:
I have now added support for SHELL= option in [config] section of dosbox-x.conf to resemble SHELL= in DOS's config.sys file. With this you can now specify a non-default command shell such as the free but powerful 4DOS 8.00 shell for 4DOS features and capabilities. In order to make this work a slightly modified version of the 4DOS 8.00 (consisting of 4DOS.COM, 4DOS.HLP, 4HELP.EXE) is now added to the Z: drive. For example, you can now specify "SHELL=4DOS.COM" to launch this shell at start. There is now a [4dos] section added in the dosbox-x.conf file to act as the 4DOS.INI file if you use this shell. On the other hand, the current command shell will be used as is if SHELL= is not specified, or if you specify "SHELL=COMMAND.COM".

### DIFF
--- a/dosbox-x.reference.conf
+++ b/dosbox-x.reference.conf
@@ -1886,6 +1886,9 @@ cd-rom spinup time      = 0
 cd-rom spindown timeout = 0
 cd-rom insertion delay  = 0
 
+[4dos]
+rem = This section is the 4DOS.INI file, if you use 4DOS as the command shell
+
 [config]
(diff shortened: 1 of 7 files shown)


joncampbell123:
4DOS was awesome back in the day, glad to see it added.

PR #1728, opened 2020-07-10
Document the country codes and default codepages

rderooy:
# Description

Add some comments for clarification.

Also, why does it default to CP 437 for countries that would have used 850 (or 858)? Is this to ensure the dosbox-x info box on startup does not get messed up?

Right now things are rather inconsistent.
- Lots of countries or locales are missing from this windows autodetect list
- It always tries to use CP 437 when you use the autodetect, or set a keyboardlayout in your config
- It instead defaults to using CP 850 when you use the keyb command, even for "keyb us".

I'm also not sure how autodetect works on Linux and Mac. Is that perhaps in the SDL code?

Documents and websites used as reference:
https://www.aivosto.com/articles/charsets-codepages-dos.html
http://www.borgendale.com/keyboard.pdf
https://www.science.co.il/language/Locale-codes.php

### DIFF
--- a/src/dos/dos_keyboard_layout.cpp
+++ b/src/dos/dos_keyboard_layout.cpp
@@ -1272,11 +1272,59 @@ public:
 					layoutname = "sg";
 					wants_dos_codepage = (IS_PC98_ARCH ? 932 : 437);
 					break;
+				case 2057: // UK, CP 850, Alt CP 437
+					layoutname = "uk"; 
+					wants_dos_codepage = (IS_PC98_ARCH ? 932 : 437);
+					break;
+				case 2060: // Belgium-French, CP 850, Alt CP 437
+					layoutname = "be";
+					wants_dos_codepage = (IS_PC98_ARCH ? 932 : 437);
+					break;
+				case 2064: // Swiss-Italian, CP 850, Alt CP 437
+					layoutname = "sf"; // Uses Swiss-French
+					wants_dos_codepage = (IS_PC98_ARCH ? 932 : 437);
+					break;
+				case 2067: // Belgium-Dutch, CP 850, Alt CP 437
+					layoutname = "be";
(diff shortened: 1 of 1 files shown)


joncampbell123:
Code page 437 is the initial code page of the IBM PC, and a good default if it doesn't have a country code. Code page detection is not part of SDL, and is entirely added on by DOSBox-X.

rderooy:
Does not really answer the questions :-)

Also this method of "detecting", is just making assumptions. These Windows locale codes say nothing about the actual keyboard layout. You can have set one locale, but have a totally unrelated keyboard layout.

To give an example, I live in Luxembourg but use a US keyboard, not the official Swiss-French keyboard. In addition, Belgium and German keyboards are also common here. Another example is the Netherlands, the existence of an official Dutch keyboard layout is pretty much unknown to the majority of the population. If you go into a shop to buy a PC you will only find US keyboard layouts. To get a Dutch keyboard you need to special order it from a major PC manufacturer (I know IBM and Apple had them in the past).

rderooy:
Was trying to find a way to "detect" the current keyboard layout on Linux that works with both Xorg and Wayland. The only thing I found so far is dependent on the desktop environment. For instance

$ gsettings get org.gnome.desktop.input-sources mru-sources
[('xkb', 'us+euro'), ('xkb', 'us+intl'), ('xkb', 'be'), ('xkb', 'de'), ('xkb', 'fr'), ('xkb', 'us')]

This is an ordered list with the most recent (current) keyboard layout you used on the left (I have several keyboards defined for testing).

But KDE, XFCE, etc would need their own methods... Some of those may be caught by a more generic Xorg solution using xkb such as:

$ setxkbmap -query

KDE has "kkbswitch" to query the keyboard layout, but I don't know if it works with Wayland.

Obviously these are command line tools, and not suitable as such...

Issue #489, opened 2018-02-04
Suggestion: a HOSTRUN command (runs command on Windows host system)


emendelson:
I've found this useful in my SVN-based project. It's a command that runs a command on the Windows host system, like this: HOSTRUN notepad d:\path\filename.txt

In a project that I use for running DOS applications, I use the name START, but that won't work if the guest is a Win95/98 system that has its own internal START command, so I've renamed it HOSTRUN here. Here's the diff (and I'm only offering it for anyone who might find it useful, even if it doesn't go into DOSBox-X). And it's probably obvious that I don't understand this code, but simply took bits and pieces from other people's code until it worked.

https://github.com/emendelson/dosbox-x/commit/4db8647a1dfc590733b8cd05894a19fd1b25fd46

joncampbell123:
I like the idea, but I have some reservations about it.

Generally when people run commands in DOSBox they expect the command to stay within DOSBox. People generally don't expect the DOS session to spawn programs on their host OS. So to prevent malicious dosbox.conf setups that format your hard drive in Windows, it would be off by default.

I propose that your version could be adopted with the requirement for the --async switch to indicate that you want it to run independently of DOSBox-X.

A future idea I do want to do with the command is that unless --async is given, the subprocess is spawned with STDIN/STDOUT/STDERR redirected to DOSBox-X and the program's console I/O is routed to the DOS console in DOSBox-X. A dummy DOS program runs until the actual process terminates.

You could run Microsoft's recent C++ compiler within DOSBox-X like any other DOS program that way. It looks like it's on the DOS console even though it's actually running the program on the host and handling STDIO to the console.

Other rules:
- If the subprocess terminates, then DOSBox-X returns to the command prompt like normal DOS termination
- CTRL+C and CTRL+BREAK would send the proper termination signal to the subprocess
- Closing DOSBox-X will terminate the subprocess right away
- Console I/O will be handled unless --async given even if the subprocess is a graphical Windows program. It's possible for graphical Windows programs to emit to STDERR even if you don't see it.
- Modifications will be added in the future to bring this command to Linux, Mac OS X, etc. (CreateProcess in Windows vs fork/exec in Linux)

emendelson:
All those ideas sound vastly useful - and far beyond anything I had thought of when I added this to a project that mostly runs WordPerfect for DOS and uses the command to run file-conversion routines, etc., in the host.

[...]

PR #1729, opened 2020-07-11
START command to run command on Windows host (enabled by the -winrun command-line option)

Wengier:
I have noticed that there are requests about interacting with the host system from within DOSBox-X (such as Issue #489 and others elsewhere), so I added the START command to do this (currently only for Windows). This command is disabled by default for security reasons, but can be enabled by either the "startcmd" config option (in [dos] section) or the -winrun command-line option (which will also enable the CLIP$ device).

The specified command(s) will be run on the host system (instead of inside DOSBox-X), and the /MAX, /MIN, /HID options can be used to run the program maximized, minimized, or hidden (they can be shortened to +, -, _ respectively). Also, START will run commands listed in the "startincon" config option (separated by space) in the Windows Command Prompt and wait for a key press before exiting.

For example, ``START + NOTEPAD.EXE" will run NOTEPAD.EXE maximized, and "START DIR" will run Windows's DIR command in the Command Prompt. The START command itself will wait for program (Notepad and DIR in these examples), so it will return to the DOSBox-X shell when the program ends, but the user can press Ctrl+C in the DOSBox-X window to terminate the program right away of course (or by closing the DOSBox-X window).


joncampbell123:
Can you confirm START is disabled unless you specify -winrun?

Wengier:
> Can you confirm START is disabled unless you specify -winrun?

@joncampbell123 Yes, START.COM won't even appear on the Z drive unless you enable it manually.

PR #1736, opened 2020-07-13
Current directory as default directory for IMGMAKE image file generation on mounted local drives and delete generated file if error occurs

Wengier:
Currently IMGMAKE command always generates the disk image file on the DOSBox-X program directory if no path is given in the image file name (e.g. "FD.IMG"), which is likely not expected if the current directory within DOSBox-X is a mounted local directory. So I updated the code to use the current directory inside DOSBox-X (instead of the DOSBox-X program directory) as the default directory for the image file generation if the current drive of the integrated DOS is a mounted local drive or mounted overlay drive. On the other hand, if the current directory is not a mounted local or overlay drive, then the default directory for the image file generation will still be the DOSBox-X program directory, and there will also be no change if an absolute path (e.g. D:\DIR\FD.IMG) is specified instead (the host path will be used as before). The same applies to the recently-implemented START command.

Also, as mentioned in an earlier thread, IMGMAKE command currently keeps the image file even if an error occurs. As a result, the generated image file is useless since it is not mountable by the IMGMOUNT command anyway. So I decided to add code to remove the generated image file in such case so that no unmountable disk image files will be generated on the disk.

Also fixed the Visual Studio SDL1 code compilation.


Wengier:
The updated code has been tested in Windows (Visual Studio 32 and 64-bit, SDL1 and SDL2, as well as MinGW64) and Linux and confirmed to work on these platforms.

Issue #1658, opened 2020-06-15
DOSBox-X won't generate or read any config files


Slider-Whistle:
**Describe the bug**
DOSBox-X seems to be completely unable to read or write to any config file.

**To Reproduce**
1. Calling dosbox-x -printconf will return $HOME/.dosbox/dosbox-0.83.1.conf (forgive the slightly outdated version). However, editing said file won't make any changes to DOSBox-X's behaviour.
2. Configuring DOSBox through its inbuilt interface will seem to do things, but no config file is written to and the program consequently forgets all changes on relaunching.
3. "CONFIG.COM -all -wcd dosbox.conf" does not generate any files.

**Environment (please complete the following information):**
 - Slackware current x86_64
 - Compiled and installed DOSBox-X from this link: https://slackbuilds.org/repository/14.2/system/dosbox-x/?search=dosbox-x

rderooy:
@Slider-Whistle The latest dosbox-x version has had a few updates in this regard. The default location on Linux is now ~/.config/dosbox-x
```
$ dosbox-x -printconf
/home/robert/.config/dosbox-x/dosbox-x-0.83.3.conf
```
sorry for the high version number, I'm running a development build that I compiled myself.

2) I tried it just now, I changed SDL autolock to true, and made sure to use the save option (Configuration -> Save) before closing the Configuration Gui. And the config file was properly updated.

3) config.com has been fixed already such that at least ``config.com -all -wcd`` works properly.

Slider-Whistle:
Never noticed the little menu at the top of the DosBox-X Configuration window, it was a huge help you brought that up. I know it's probably copying the look of Windows 3 or something there, but that's missing some major visibility cues.

I think it could probably be helped by adding little underlines under the first characters (whether you can actually press alt+C to open up "Configuration" or not), making the main body of the window white while the top part is gray, and most importantly, moving the "Save" option next to "Close", maybe labelled as "Apply" for just an extra bit of familiarity. Maybe I'm just blind though.

Selected "Save" through the configuration menu's menu labelled "Configuration" and it worked, just like you said. Thanks again for the help.

As for the config.com thing and the file -printconf points to, it sounds like both have already been addressed.

So if I want the config file to start being read on boot, I should try upgrading to the latest stable version?

rderooy:
@Slider-Whistle Please upgrade to the latest version if you can. It should fix some issues in this regard.

[...]

PR #1738, opened 2020-07-15
Some improvements to the Configuration GUI (e.g. "Save..." button next to the "Close" button)

Wengier:
I noticed that there was a suggestion in Issue #1658 to put a "Save" button next to "Close" button for the Configuration GUI so that users will remember to click the "Save..." button before clicking the "Close" button (if they do want to save the changes to the config file), so I did this. Previously the user might assume that the setting were automatically saved just by clicking "Close", which is in fact not the case.

I also slightly updated the INSTALL file to mention the macOS packages found in the Releases page.


Wengier:
I have improved the Configuration GUI's handling of "4dos" and "config" sections by adding a separate handler for them, as they can not be handled in the exact same way as the other sections.

Wengier:
It has been tested and confirmed to work in Windows and Linux.

Wengier:
I have expanded the INSTALL file for more information about the releases and also moved the library and source code configure information from the INSTALL file to the README.source-code-description file, as the former is focused on the installation part and the latter is focused on the source code compilation part.

PR #1740, opened 2020-07-17
Upgrade INSTALL to INSTALL.md for improved readability

Wengier:
The current INSTALL file is a plain text file which does not support any formatting, so it does not look very good compared to pages like README.md (which also links to it). Thus I decided to convert the INSTALL file from a plain text file to INSTALL.md page which should now look much better, similar to the README.md page. Further installation instructions are also added to it so that most things related to the DOSBox-X installation needed by users are now mentioned in the documentation.


Wengier:
It now fixes Issue #1741 (the game "Disney's Duck Tales: the Quest for Gold" not starting).

Wengier:
LS is now an external command appearing on Z: drive as LS.COM instead of a built-in shell command. This avoids name clash with Links LS 97 which has the main executable called LS.EXE.

Wengier:
I have added a new feature to this pull request - now there is a "Reboot guest system" menu item (under "Main") to reboot the guest OS or the integrated DOS kernel directly. Previously there was only a "Reset guest system" menu item, which will reset the whole virtual machine and then boot into the integrated DOS, but there was no way to simply reboot the guest OS. There was also no way to reboot the kernel of the integrated DOS from the menu either (rather than using the RE-DOS command).

The new "Reboot guest system" menu item will do both according to whether the guest OS is running: if the guest OS is currently active, then it will reboot the guest OS; if the integrated DOS is currently active then it will reboot the kernel of the integrated DOS. On the other hand, the previous "Reset guest system" menu item has been renamed to "Reset virtual machine" which will restart the virtual machine and boot into the integrated DOS within DOSBox-X as it was previously done (no change in this function).

PR #1745, opened 2020-07-21
Fix rebooting El Torito floppy guest system and auto-assume "-fs none" for IMGMOUNT to mount drive numbers

Wengier:
The new "Reboot guest system" feature was recently added to reboot the emulated DOS (guest DOS or integrated DOS) directly, but I have noticed that it did not work for guest OS mounted via El Torito floppy drive. So this pull request fixes this issue. Also, IMGMOUNT command now lists "El Torito floppy drive" as the disk name for this drive number.

Furthermore, the IMGMOUNT now automatically assumes "-fs none" for mounting drive numbers, so that you no longer need to specify "-fs none" manually in such case. For example:

``imgmount 0 disk.img``

Will mount drive number 0 to the disk.img image file, without having to enter ``imgmount 0 disk.img -fs none`` for this.


joncampbell123:
Please update the documentation to indicate that IMGMOUNT assumes -fs none for drive numbers. Make sure to mention that this is a DOSBox-X variation and other DOSBox forks including the main DOSBox project may still require -fs none in that case.

Wengier:
@joncampbell123 Unlike DOSBox (and other DOSBox forks), IMGMOUNT command in DOSBox-X had already done a number of things automatically, e.g. it automatically assumes ISO type for .iso/.cue files, and it also has much better HDD geometry auto-detections. Most of these things are already covered in the Wiki, and I will add the "-fs none" for this case too.

Issue #1762, opened 2020-07-27
Spanish keyboard problems again: can't type accented upper-case Ó


emendelson:
Using current code, both SDL1 and SDL2 builds have problems with the Spanish keyboard, These were reported to me by a user of my WordPerfect system, but the user tested with the SDL1 and SDL2 builds of the current code, not my patched version. He reports that with KEYB SP 850 in [autoexec], these problems occur:

SDL1: AltGr does not work (has no effect). The -/_ key produces the wrong characters (I don't know which characters appear, but I've asked). Ó accentuated uppercase does not appear.

SDL2: AltGr does not work (has no effect). Ó accentuated uppercase does not appear.

I haven't tested this, but the report seems to be reliable.

rderooy:
I just tried it with the SDL2 version here, and AltGr works to produce the Euro symbol when I press AltGr-e, also various other keys on the right hand side of the keyboard, and under the number keys, have additional symbols when I use AltGr.

Dead keys also work, so when I press ' and then e, I get é or with shift É. Likewise, I can produce a lower-case ó, but in this case not the upper case variation. Ó. Not sure why that is.

p.s. unless you have a good reason you should use codepage 858 as it is basically the same as 850, but includes the Euro symbol.

SDL1 and non-us keyboards requires that you set
```
[sdl]
usescancodes=true
```
This is also mentioned here https://github.com/joncampbell123/dosbox-x at the bottom of the page in "International Keyboard Layouts". When I do that here, it seems to behave the same as the SDL2 binary.

rderooy:
@Wengier @joncampbell123 International keyboard layout issues are a common problem, particular with the SDL1 build. Perhaps usescancodes should default to auto, which would mean that if (SDL1 && !keyboard=us) usescancodes=true

What do you think?

emendelson:
@rderooy I will ask my user to change from 850 to 858 and let me know what happens. This is very strange. He is a very reliable reporter.

@joncampbell123 @wengier My user tells me that this build (made from my patched version of the April 2020 code) works perfectly, with the accented upper-case Ó and everything else:

http://www.columbia.edu/~em36/wpdos/DOSBox-X-17April.exe

That was a long time ago. I would like to try to narrow down the differences between my version from that date and the current version. I know that the main differences would probably be in dos_keyboard_layout_data.cpp, but where else should I look?

[...]

PR #1767, opened 2020-07-29
Add CP 850/858 for host to guest code page mapping

Wengier:
As mentioned in Issue #1762, DOSBox-X currently only supports  code page 437, 866 and 932, but not 850 or 858, which are very common for users. So I added support for these code pages to DOSBox-X.

Also allowed to change the "usecodepages" and "startwait" config options via the config -set command.

Tested to ensure the code works.


Wengier:
I have added code page 852 (Latin II) and 855 (Cyrillic) too, which are useful for Central and Eastern European users.

rderooy:
I tested the PR locally on Linux with both SDL1 and SDL2 binaries. It worked great. I can now list files with a Ó in the filename when using an appropriate codepage, and can type the character when using an appropriate international keyboard layout.

Hoping this PR still makes it for the next release version.

joncampbell123:
@Wengier Speaking of release, this month's release is going to be a little late due to how busy I've been with work the past few weeks. Expect a release of perhaps August 2nd or 3rd.

Wengier:
@joncampbell123 I see. I think I can understand it because I have to make time to do work myself too.

PR #1770, opened 2020-07-30
Updated keyboard, codepages and regional settings

rderooy:
Update README.md to reflect recent changes to the code.

This would ideally be applied when a new release is made, to reflect the release version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,53 +334,49 @@ URL: https://github.com/yksoft1/dosbox-x-vanilla-sdl/tree/emscripten (look for c
 International keyboard layouts and codepages
 --------------------------------------------
 
-DOSBox, and by extension, DOSBox-X was developed around
-the US keyboard layout.
-This is primarily due to limitations around SDL1. SDL is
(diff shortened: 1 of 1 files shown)


Wengier:
@rderooy I appreciate the updates, but I think the section ("International keyboard layouts and codepages") should probably be moved to the Wiki instead, which is the actual DOSBox-X user guide.

rderooy:
@Wengier I had thought of that. It was simply because README.md already had a section on keyboards that I kept it.

I was actually thinking on a "Input" page on the wiki, handling keyboards, mice and joysticks. But that will not be done in the next few days, so I would suggest to add this commit for now, and then for the next round, when that page will hopefully exist, it can be removed from README.md.

Wengier:
@rderooy I think you can mention the country auto-detection with “country=0” too, although it only works on Windows, and I don’t know if the detection is 100% reliable.

Wengier:
@rderooy I have already created a DOSBox-X regional settings page in the DOSBox-X Wiki:

PR #1771, opened 2020-07-31
Move guide on keyboard layout and region settings from README.md to user guide in Wiki

Wengier:
I have already created a new "Guide:Regional settings in DOSBox‐X" page in the DOSBox-X Wiki for customizing the regional settings in DOSBox-X, so I am removing the duplicate content from the section "International keyboard layouts and codepages"/"Regional settings" of the README.md file. Also change the reference to "Alt+Enter" key to the "Host+F" key in the comment of the dosbox-x.reference.conf file because the latter (not the former) is the shortcut for toggling between fullscreen mode in DOSBox-X.


Wengier:
The new Wiki page on DOSBox-X regional settings is available from the following URL. It already has more content than the one in README.md, but it can still be further expanded.

Wengier:
I have added a menu (under "Main") to enable or disable the existing mouse wheel movements to arrow feature.

PR #1772, opened 2020-07-31
Document country settings

rderooy:
No functional changes, just documentation to better understand what is happening.

Also, what is country 88 meant to be? It does not appear in any table I have been able to find. Generally the codes are based on the international dialling codes of countries, and there is no 88 dialling code as it would conflict with several 3-digit dialling codes starting with 88 (Bangladesh (880) and Taiwan (886)).

### DIFF
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -254,7 +254,6 @@ void DOS_SetCountry(Bit16u countryNo) {
 		case 81:  // Japan
 		case 82:  // South Korea
 		case 86:  // China
-		case 88:  // ???
 		case 354: // Iceland
 		case 886: // Taiwan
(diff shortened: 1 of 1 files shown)


Wengier:
@rderooy I think 88 is a short alias of the country code 886 in this case. Unlike international dialing codes, when used as an alias country code no conflict will be caused apparently.

BTW: I have added a few new sections to the "Guide:Regional settings in DOSBox‐X" Wiki page:

rderooy:
Well, 886 is already listed. And 88 does not appear in any COUNTRY.SYS tables I have been able to find. So I think it should just be removed. Any objection?

Wengier:
@rderooy The country code 88 seemed to be unassigned by ITU/CCITT in 1964. So it can be removed I guess.

PR #1777, opened 2020-08-02
Fix load state

Wengier:
I have noticed that commit 5bcab09 ("Fix VS2019 compiler warnings") broke the save state feature because of type mismatch in the file opl.cpp and dbopl.cpp. This fix brings the save and load state feature back working. It will also save and display the timestamps for the saved states.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,9 @@
 0.83.4
   - Fixed loading saved states in fullscreen mode with
-    the "aspect=true" setting. The timestamp for the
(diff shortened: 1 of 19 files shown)


joncampbell123:
@Wengier  Ready for me to merge?

Since I'm almost two days late I think it's time to make a build tonight.

Or, do you need me to wait to finish your changes?

Wengier:
@joncampbell123 I think it is ready for be merged now. It is already past mid-night in my location, so please go ahead to make a build.

Issue #1428, opened 2020-03-03
Glide wrapper patch


arromdee:
gulikoza's Glide patch (https://www.vogons.org/viewtopic.php?t=16462 ) that works with a Glide wrapper.

A number of years ago I used this patch and it had a much better framerate in Tomb Raider than the Dosbox-x glide support, and worked fullscreen.  My computer has newer parts now, but I still don't get maximum framerate.

My current system is Mageia Linux 7.1, 64 bit.

rderooy:
I had a look at this. It seems the latest glide code can be found as part of DOSBox ECE.
https://dosboxece.yesterplay.net/

Most of the code is pretty straight forward to merge it seems, you just need to take these files verbatim:
src/hardware/glide.cpp
include/glidedef.h
include/glide.h

Add a few config options to dosbox.cpp, add a little code to sdlmain.cpp and update Makefile.am

Where I got stuck, was the last file. The original code adds itself to the Mem_GetPageHandler() in src/hardware/memory.cpp. And that function has changed quite a bit. Looking through the change history, I found that @joncampbell123 had done this to speed up the code, such that not every page request was going through a bunch of range checks. So adding range checks back in seems the wrong thing to do.

Wengier:
I have already added support for the Glide wrapper in my latest pull request. The original code only supports SDL1, so it needs to be modified somewhat for it to also work with SDL2. Furthermore, the library file glide2x.dll (Windows), libglide2x.so (linux), or libglide2x.dylib (macOS) will be required for Glide to work. Tested with 3dfx version of Tomb Raider and confirmed to work.

arromdee:
Unfortunately, using the glide wrapper and openglide crashes in openglide.  I mentioned this in #1788 and also [reported this](https://github.com/voyageur/openglide/issues/13) to voyageur's openglide, but don't have high hopes of anyone ever being able to look at it.

It also turned out that my initial problem of low level emulation not being fast enough was not a problem because I didn't have cycles=auto, but it's still awkward to use it because it can't be used fullscreen or in a large window as per #1215.  Pure software Voodoo emulation on Tomb Raider is unusable, because any text in the emulated glide window (such as "Lara's Home" on the start screen) is replaced by colored blocks.  I also get sound stuttering which is not helped by setting a high blocksize and happens even though my CPU usage is nowhere near 100%.

Wengier:
[...]

PR #1782, opened 2020-08-03
Update documentation to list DOSBox-X 0.83.4 version packages (instead of 0.83.3 version)

Wengier:
The latest version of DOSBox-X (0.83.4) has been recently released. So I have updated the documentation (particularly INSTALL.md) to show DOSBox-X 0.83.4 packages instead of the previous 0.83.3 ones accordingly.


Wengier:
It now includes support for Glide wrapper as mentioned in Issue #1428.

Wengier:
I have increased the number of save slots from 10 to 100. It seems that 10 save slots are still not enough for some people, but 100 should be enough. Each page in the save slot menu (under "Capture") still shows 10 slots, but you can now go to the previous or next page (up to 10 pages) for more save slots.

PR #1790, opened 2020-08-06
Add "Quick launch program..." menu to quickly run program inside DOSBox-X

Wengier:
I have noticed that Daum does have a menu item to quickly launch a program (.exe, .com, .bat) selected from the file browser to quickly run inside the virtual machine, so I have added the feature to DOSBox-X too. It is currently Windows-only, so that users can  easily run programs from the GUI menu selection instead of typing commands. The directory will be auto-mounted before the program is launched, and will be auto-unmounted after the program has finished running.

Tested in both Windows and Linux to confirm the code works (or compiles).


Wengier:
I have updated the xxhash decoding library from 0.7.4 to 0.8.0, as in dosbox-staging.

Nikos-1974:
It will be nice if dosbox-x ask you after program execution this: ....do you want to stay in the mounted directory Yes/No ?
If user select yes dosbox-x will stay in mounted directory it No the will unmount the path :+1)

Wengier:
@Nikos-1974 I have added the option to not auto-mount the drive after the program finishes in my latest pull request. As for other ideas you mentioned - maybe later.

Wengier:
The "Quick launch program..." menu is now available for non-Windows platforms as well in addition to Windows systems.

PR #1794, opened 2020-08-08
Fix Glide reset VM and allow quick launch to not auto-unmount drives

Wengier:
I noticed that resetting virtual machine will probably not work when the Glide feature is currently active. So I fixed this in the code. Also, I improved the new quick launch feature to ask the user whether to auto-unmount the drive after the program has finished execution.


Wengier:
The save state feature now saves and restore the mounted path for mounted local or overlay drives. This fixes the "Save state corrupted" error when trying to loading states but the local drives are not yet mounted.

Issue #1789, opened 2020-08-06
Joystick (XBox 360 Controller Wireless) does not work


maximus105:
**Describe the bug**
Joystick (XBox 360 Controller Wireless) does not work at all in DOSBox-X with standard configuration, while in DOSBox-Staging v0.75.0, DOSBox ECE r4356, DOSBox SVN-Daum everything works fine.

**Environment (please complete the following information):**
 - Windows 10 x64 2004
 - DOSBox-X 0.83.4 x64 SDL1

Wengier:
Have you used the key mapper to map the joystick buttons? What happens when you try to do so?

maximus105:
In the mapper, the assigned joystick buttons are preserved (they are highlighted in white), but when you press them on the joystick, nothing happens (they do not blink in the mapper).

Wengier:
My Logitech Dual Action Gamepad does work with the key mapper however. When I press some button on the joystick that is saved it will blink in the mapper. I have only one joystick to test at this time though. Perhaps you have more than one joysticks?

maximus105:
No, just one joystick.

Wengier:
I don’t have XBox 360 Controller Wireless to test myself, so I cannot say what exactly happened. Either it is unsupported in some way, or there is something wrong in the system configuration.

joncampbell123:
I have a wired XBox 360 controller, and here on Linux it seems to work just fine, except for some reason SDL1 is mapping LT as one of the axises of the second analog joystick.

maximus105:
I don't quite have an XBox 360 Controller Wireless. I have a Sony Dualshock 3 connected via Bluetooth to a PC via program SCP Server, it works like an XBox 360 Controller in XInput mode. All games under Windows 10 x64 work fine. Also it works fine in DOSBox-Staging v0.75.0, DOSBox ECE r4356, DOSBox SVN-Daum.

maximus105:
[...]

PR #1795, opened 2020-08-10
Improve save state feature's handling of drive assignments and fix Issue #1789

Wengier:
I have further improved the save state feature's handling of changed drive assignments. As a result, when the drive assignments in the saved states are different from the current state, DOSBox-X can automatically restore the drive assignments in most (although not all) cases without showing the "Save state corrupted" error, including all drive types such as regular FAT and ISO drives if the image files still exist.

Also fixed Issue #1789 related to the joystick in the SDL1 build, and added direct download links in the INSTALL.md page to the release files.

Tested in both Windows and Linux to confirm the code works (or compiles).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,10 +21,13 @@
     This fixes the "save state corrupted" error while
     trying to load saved states but the drives are not
     yet mounted in DOSBox-X. (Wengier)
-  - Increased save slots from 10 to 100. Each page in
-    the save slot menu (under "Capture") still shows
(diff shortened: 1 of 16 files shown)


Wengier:
I have added support for automatically re-mounting other FAT drive types for loading states too, including El-Torito floppy drives, RAM drives and VHD images, as long as the image files still exist on the host (for RAM drives, the drive will be re-mounted to the default empty drive). As a result, virtually all drive types should be supported by this feature by now.

PR #1799, opened 2020-08-13
Add option to remove saved state and silence Glide compile warnings

Wengier:
I have added the ability to remove saved states in the selected save slots in case the users find that some saved states are no longer necessary and want them erased from the slots. There will be a confirmation before this operation. Also silenced most (if not all) of the GCC compiler warnings for the glide.cpp file.

Tested in both Windows and Linux to confirm the code works.


Wengier:
I decided to add the new default option to the Windows installer (in case the dosbox-x.conf config file already exists in the install directory), which allows the installer to automatically upgrade the dosbox-x.conf file to the new version format while keeping all the user-customized settings already made. As a result, the config file will always be up-to-date after a version upgrade, including all options of the latest DOSBox-X version and also will keep all the changes already done by the user.

Wengier:
The default value for "output" is now "default", which will default to the value according to your platform when DOSBox-X runs. This makes the dosbox-x.reference.conf config file portable (instead of a specific value like "surface" or "opengl", or "direct3d" varying by the platform, which is apparently not portable).

Wengier:
I have updated the update-build-timestamp.pl script to include DOSBox-X's Git commit SHA1 (e.g. 5d157b4) so that VER /R will show it, which can make bug reporting from users easier.

PR #1808, opened 2020-08-20
Show mounted drive numbers in menu (under "DOS")

Wengier:
The previous code added support for showing details for each drive (A-Z) from the "Drive" menu. This pull request further added the ability to show detailed information for mounted drive numbers (0-5) from the "DOS" menu. Unlike drive letters (A-Z), drive numbers will work even when booting into a guest system, so this function allows the user to see the currently inserted disk (and information like swap position or IDE controller) when running a guest system too.


Wengier:
The "drive information" menu item now shows whether a drive is read-only for the selected drive (A-Z).

Issue #1802, opened 2020-08-16
Descent II CD track #1 stops playing in game


DSN9990:
**Describe the bug**
CD track 1 (not the ones from the main menu) of Descent II stops playing ingame when it reaches 0:40 (or maybe 0:39/0:41).
All other tracks appears to be fine and play properly.

**To Reproduce**
1. Run Descent II in Dosbox-x
2. Start a new game
3. Leave Track 1 and wait until it stops playing

**Expected behavior**
Track 1 should play properly, along with the rest.

**Environment (please complete the following information):**
 - Windows 10 64bit
 - DOSBox-X 0.83.4

-------------------------------------- d o s b o x - x . c o n f  ----------------------------------------
 [sdl]
output=direct3d
autolock=false
mouse_emulation=always

[dosbox]
vmemsize=1
memsize=16
vesa modelist cap=16
allow low resolution vesa modes=true
allow explicit 24bpp vesa modes=true
allow high definition vesa modes=true
allow unusual vesa modes=true
allow 32bpp vesa modes=true
allow 24bpp vesa modes=true
allow 16bpp vesa modes=true
allow 15bpp vesa modes=true
allow 8bpp vesa modes=true
allow 4bpp vesa modes=true
allow 4bpp packed vesa modes=true
allow tty vesa modes=true
ignore extended memory bit=true

[cpu]
core=dynamic
cputype=pentium
cycles=max

[pci]
voodoo=auto
glide=true

[dos]
hard drive data rate limit=0
ver=5.0
---------------------------------------------------------------------------------------------------------

DSN9990:
Why is causing this issue? This occurs even in Earth 2140, another game by Interplay.
Only part of the first track plays and then it stops playing.

DSN9990:
Why is causing this issue? This occurs even in Earth 2140, another game by Interplay.
Only part of the first track plays and then it stops playing.

joncampbell123:
Does it read data from the CD at any point?

CD-ROM drives can either play audio or read data, they can't do both. However reading the wrong type of sector doesn't cause that.

Case in point, there was a bug I fixed awhile back regarding Windows 95 and CD audio playback. Windows 95 does a background poll (once a second) using a READ CD command for data, even if you're playing an audio CD. I think it does that because DOSBox-X doesn't emulate the ATAPI lock/unlock door and notification commands.

joncampbell123:
[...]

Issue #1809, opened 2020-08-20
Non-working/Buggy DOS games list


DSN9990:
Games (FIXED):

1. Screamer Rally - hangs with blinking cursor DOS prompt when running setup program, but installs fine.

2. Grand Theft Auto - sound setup fails, does not save configuration, also install hangs when I try to install the game again (after the game is installed).

3. Descent II - CD track 1 stops and does not continue playing anymore.

Games (NOT WORKING/BUGGY):

1. Cannon Fodder (CD) - Gives an SoundBlaster initialization error both in install program when trying to test sound and when trying to run the game (after the intro ends).

Config:

```
[sdl]
fullresolution=desktop
output=direct3d
autolock=false
mouse_emulation=always

[dosbox]
machine=svga_s3
vmemsize=1
memsize=16
vesa modelist cap=16
allow low resolution vesa modes=true
allow explicit 24bpp vesa modes=true
allow high definition vesa modes=true
allow unusual vesa modes=true
allow 32bpp vesa modes=true
allow 24bpp vesa modes=true
allow 16bpp vesa modes=true
allow 15bpp vesa modes=true
allow 8bpp vesa modes=true
allow 4bpp vesa modes=true
allow 4bpp packed vesa modes=true
allow tty vesa modes=true
ignore extended memory bit=true

[cpu]
core=dynamic
cputype=pentium
cycles=max

[dos]
hard drive data rate limit=0
ver=5.0

```

Now using latest (0.83.5 64Bit SDL1).

DSN9990:
Screenshots:

Formula Karts:
![dosinst_001](https://user-images.githubusercontent.com/69558812/90756040-45b88c80-e2b2-11ea-8ac0-a7b5b82acf0f.png)

Screamer Rally:
![command_000](https://user-images.githubusercontent.com/69558812/90756092-5406a880-e2b2-11ea-8983-f2c2da219755.png)

and more to go...

DSN9990:
@joncampbell123 Can you test/reproduce this?

joncampbell123:
@DSN9990 If I can find the game online, I'll try.

However first... did you run LOADFIX before running the game?

DSN9990:
No. I tried now and it did not work also. Same problem.

Here it's the link to the game: [https://www.old-games.ru/game/download/5701.html]

Wengier:
[...]

PR #1811, opened 2020-08-23
Fix Issue #1802 (CD audio code re-port) and Grand Theft Auto installer

Wengier:
Issue #1802 pointed out some issue(s) exist in the CD audio code, ported previously from dosbox-staging. On the other hand, I noticed that the DOSBox ECE's CD audio code works fine, so I decided to re-port the relevant code (including SDL_Sound) from DOSBox ECE instead. As a result, the CD track should no longer stops while playing in game, and MP3/OGG/FLAC tracks will still be supported.

I also fixed the issue with the Grand Theft Auto installer which may get stuck with "COMMAND" shown in the titlebar when trying to re-install the game to the same directory, as mentioned in Issue #1809.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Although based on the DOSBox project, DOSBox-X is now a separate project because
 
 * Support for 3dfx Glide and Voodoo chip emulation
 
-* Support for cue sheets with FLAC, Opus, Vorbis, and MP3 CD-DA tracks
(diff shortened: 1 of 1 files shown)


rderooy:
@Wengier seems this was the fix for the CD audio issue: https://github.com/dosbox-staging/dosbox-staging/commit/9b3e6cdde8f52c862cc8d57c8c811dc5bb935ed0

PR #1832, opened 2020-09-08
Move font to contrib/fonts

rderooy:
Cleanup the root directory by moving "font" to "contrib/fonts" and update all references I could find to it.

I have tested that Linux "make install" and make-rpm.sh still works. But could not test it on Windows or mac OS.


Wengier:
The modification looks fine.

PR #1834, opened 2020-09-09
Various icon and macos specific files moved to contrib

rderooy:
Linux builds are not effected by this. But windows and macos are, and I did not compile test those as I'm not setup to compile them.


Wengier:
@rderooy I agree with the cleanup in principle, but I think the change can break some Windows build scripts because of location change of some files, such as make-windows-release.pl and DOSBox-X-Setup.iss. Please double check them. Thanks!

rderooy:
@Wengier I had gone though the tree with fgrep to ensure I caught every reference to the files I moved. This includes DOSBox-X-Setup.iss, where I had updated some pointers. I just had another look, and I don't see anything further.

speaking of windows-installer, I was thinking of moving that also under contrib.

Wengier:
@rderooy That's fine, you can move the Windows installer under contrib as well.

Wengier:
I just finished my other changes, so perhaps I can move the Windows installer myself.

Wengier:
@rderooy The Windows installer was already moved to the new place. Thanks for the suggestion! If you have further suggestions like this, you can either try to do yourself or let me know. Thanks!

Issue #1826, opened 2020-09-04
dosbox-x: error while loading shared libraries: libpng15.so.15: cannot open shared object file: No such file or directory


CHJ85:
Hi there. I am trying to run Dosbox-X on Ubuntu. Had to compile it because the only package I could find was for Red-Hat/Fedora.
Now I can't run it because there's missing library files. Turns out these are not available in any Debian/Ubuntu package. So I had to go look for it in the Fedora repositories. That file did not work. Likely because it wasn't made for Debian based distros.
So please on your next version, either package Dosbox-x for other distros too or use dependencies that are available on every distro. Or better yet, just package it as a Flatpak. That way it'll run on every distro.
Btw. Why would you even package it for Fedora in favor of the Debian base? Kinda weird. I mean, I don't think a lot of gamers use Fedora. They usually stick to Ubuntu, Pop OS or Linux Mint. You know.

rderooy:
@CHJ85 If you successfully compiled it yourself, that means your system must have had the header files for libpng15 installed. Otherwise it would not be looking for libpng15.so.15

For instance, if your system had the header files for libpng16 installed, it would have linked with that and it would be looking for libpng16.so.16

Since you have the header files for libpng15, which I assume you got from your distro repository, you must also be able to install the actual matching libraries from your distro repository.

rderooy:
In the case of Fedora, I have the following two packages installed. the -devel one has the header files to allow me to compile DOSBox-X.

libpng-devel-1.6.37-3.fc32.x86_64
libpng-1.6.37-3.fc32.x86_64

CHJ85:
Yea I figure I didn't compile it correctly. So I kinda just gave up this project before I corrupt my system somehow.
But again, why can't you guys just do flatpak instead?

CHJ85:
Also, I tried the "alien" way of simply converting the RPM to DEB, but that gave me the same error.

dreamer:
@CHJ85 many Linux distros moved away from libpng15 years ago due to security concerns. Some distros (e.g. Fedora) provide it for compatibility reasons, but it's rather rare.

Perhaps DOSBox-X should do it the same way we handle packages in [dosbox-staging](https://github.com/dosbox-staging/dosbox-staging)? We provide simple tarballs built against Ubuntu 18.04 LTS - it's old enough to provide ABI compatibility with most distros, and we encourage Linux packagers to simply create native builds suited for specific distros.

rderooy:
[...]

Issue #1203, opened 2019-08-14
Distribute Dosbox-x in a Flatpak


LeandroStanger:
http://docs.flatpak.org/en/latest/introduction.html#reasons-to-use-flatpak
https://github.com/flathub/flathub/wiki/App-Submission

tpimh:
@egrath has this flatpak build of dosbox-x: https://github.com/egrath/dosbox-x-flatpak
Not published on flathub yet.

rderooy:
A flatpak version is now available on flathub
https://flathub.org/apps/details/com.dosbox_x.DOSBox-X

rderooy:
In addition to the flatpak, there is also a snap package, that is current
https://snapcraft.io/dosbox-x
https://github.com/snapcrafters/dosbox-x

The maintainer is a user named @popey apparently.
It uses the regular DOSBox icon, and not the DOSBox-X icon.

Note: I have not tried if it works, and I don't know how flatpak compares to snap.

popey:
@rderooy Thanks for the ping! I have updated the icon for dosbox-x in the snap store. The snap has about 7K weekly active devices installed, most of which are on the latest release.

![Screenshot from 2020-10-06 11-55-19](https://user-images.githubusercontent.com/1841272/95192996-fe874c00-07ca-11eb-9779-551eca1e9ec0.png)

rderooy:
@popey sure!

Are you using a SDL1 or SDL2 build of DOSBox-X? And are you enabling features like fluidsynth?

Looking through the git repo, I noticed you seem to be using a rather basic desktop file. There is a more extensive one included with dosbox-x in the contrib/linux directory.

**edit**

Sorry, I did not mean the desktop file, but the metainfo.xml file, but I don't see that one.

**edit**

Looking at your yaml file, it seems like you should be using SDL2, but your actually using SDL1. And you are also not enabling fluidsynth.

PR #1861, opened 2020-09-16
Add initial flatpak support

rderooy:
# Description
This adds flatpak manifest files for SDL1 and SDL2 builds.

This has been tested locally, against a local repo, which is automatically created if you use the scripts.

The SDL1 build is "flaky". In that sometimes, it fails with a broken pipe during the compilation of DOSBox-X. The point of failure however varies each time, and it can fail twice in a row and then suddenly succeed the third time, or it can complete the first time.

The SDL2 build has not failed once for me. And if we want to add DOSBox-X to the most popular repository (flathub), my suggestion would be to use the SDL2 version.

I still need to see about what to do about SoundFonts. A flatpak cannot access any 'random' files under /usr so it cannot system installed soundfont files. The only solution is to add a soundfont to the flatpak, but they are rather large. So instead of adding it to the flatpak during compilation, I will look into adding it via "extra-data" at install time.

Another issue is that FluidSynth support does not work right now with FluidSynth v2.

- https://github.com/joncampbell123/dosbox-x/issues/1826
- https://github.com/joncampbell123/dosbox-x/issues/1203

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1090,6 +1090,7 @@ src/xBRZ/Makefile
 include/Makefile
 contrib/macos/dosbox-x.plist
 contrib/linux/dosbox-x.spec
+contrib/linux/dosbox-x.appdata.xml
 make-rpm.sh
 ])
 AC_OUTPUT
--- a/contrib/linux/com.dosbox_x.DOSBox-X-sdl2.yaml
+++ b/contrib/linux/com.dosbox_x.DOSBox-X-sdl2.yaml
@@ -1,0 +1,96 @@
+app-id: com.dosbox_x.DOSBox-X-sdl2
+runtime: org.freedesktop.Platform
+runtime-version: '20.08'
+sdk: org.freedesktop.Sdk
+
+command: dosbox-x-sdl2
+
+finish-args:	# flatpak permissions
+  - --device=all	# needed for gamepads and serial/parallel
+  - --share=ipc	# needed for X11
+  - --share=network	# needed for networking (NE2000, IPX)
+  - --socket=x11
+  - --socket=wayland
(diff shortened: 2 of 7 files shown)


rderooy:
I had a look at extra-data to download a soundfont during install time. However, like many things with flatpak you have no control over it. Your download just gets unceremoniously dumped into the (chroot) /var/extra directory and you cannot do any actions on it, like rename or move. So to use extra-data for the soundfont, either the user needs to adjust their config file to point to the correct location, or dosbox-x needs to be updated to check the alternative location.

Here is a manifest fragment for extra-data (needs to be inserted into 'sources'
```
- type: extra-data
        filename: FluidR3_GM_GS.sf2
        url: https://archive.org/download/fluidr3-gm-gs/FluidR3_GM_GS.sf2
        sha256: 545b2833936f15f04df5f0c5c4096b3ba6ced46ec7031f61991cae46f8681986
        size: 151001312
```

rderooy:
The instructions for adding a project to flathub can be found here:
https://github.com/flathub/flathub/wiki/App-Submission

I am willing to try to make a flathub compatible package if there is interest in maintaining such a package.
I can submit it to flathub, or that could be done by @joncampbell123 
Even if I submit the PR to flathub, I will want to add @joncampbell123 and probably @Wengier to the project such that they can update it when a new release is made.

But I won't have time to work on this until next week at the earliest.

Wengier:
@rderooy I believe flatpak support is the way to go for Linux, so it is certainly a good thing to have such a package. Thanks for adding this initial support. As @joncampbell123 is probably busy right now, you can certainly submit it to flathub and adding us to the project so that we can all try to update the package when we have time.

PR #1880, opened 2020-09-26
Add an AVX2 version of the render cache

linkmauve:
# Description

* Clean up the SSE2 checks.
* Check for AVX2 support at runtime.
* Add an AVX2 codepath for `RENDER_DrawLine_scanline_cacheHit()`.

**Does this PR address some issue(s) ?**

It removes an unused extern variable, and lowers the runtime cost of `RENDER_StartLineHandler()` from 8.3% to 6.2% of the total execution time on the DOS prompt.

**Does this PR introduce new feature(s) ?**

It allows AVX2 to be checked at runtime, enabling some more efficient SIMD.

**Are there any breaking changes ?**

No.

**Additional information**

None that I could think of.

PR #1882, opened 2020-09-26
Fix AVX2 when not building with -mavx2 or similar

### LINKED ISSUE #1881: [Linux][SDL2] Building process is broken in render.cpp
**Describe the bug**
After AVX2 was added to main code, I cannot get Dosbox-X to be build. Every single time, build process stop while processing render.cpp.

Error log start with:

```
render.cpp:172:67: warning: AVX vector return without AVX enabled changes the ABI [-Wpsabi]
  172 |                 __m256i v = _mm256_loadu_si256((const __m256i*)src);
      |                                                                   ^
In file included from /usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/include/immintrin.h:53,
                 from /usr/include/SDL2/SDL_cpuinfo.h:86,
                 from /usr/include/SDL2/SDL.h:38,
                 from ../../include/sdlmain.h:10,
                 from render.cpp:35:
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/include/avx2intrin.h:433:1: error: inlining failed in call to 'always_inline' 'int _mm256_movemask_epi8(__m256i)': target specific option mismatch
```

**To Reproduce**
1. Grab latest git code
2. Use this configure line: ./configure --enable-core-inline --disable-debug --enable-avcodec --prefix=/usr --enable-sdl2
3. Run make and wait.

**Expected behavior**
Build process going on.

 - Archlinux
 - commit SHA1: b5310f1c3671bfe1b0226fd75a8703053fa4ecdb

Attaching crash log.
[dosbox-x-crash.log](https://github.com/joncampbell123/dosbox-x/files/5286470/dosbox-x-crash.log)

linkmauve:
In generic builds, we have to tell the compiler to change the target for
a specific function and to make sure to call it only if we know the
extension is supported by the CPU.  This is achieved using this target
function.

I’ve tested this code on both gcc and clang, with -march=native and
without, with different optimisation levels, and it seems sound this
time, sorry for breaking the generic build.

From another project, MSVC doesn’t have this restriction.

Fixes #1880.
Fixes #1881.

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -100,7 +100,7 @@
 #include <list>
 
 /*===================================TODO: Move to it's own file==============================*/
-#if defined(__SSE__) && !(defined(_M_AMD64) || defined(__e2k__))
+#ifdef __SSE__
 bool sse2_available = false;
 bool avx2_available = false;
 
@@ -114,7 +114,7 @@ bool avx2_available = false;
     __asm mov d, edx
 # endif /* _MSC_VER */
 
-void CheckSSESupport()
+static void CheckX86ExtensionsSupport()
 {
 #if defined(__GNUC__) && !defined(EMSCRIPTEN)
     sse2_available = __builtin_cpu_supports("sse2");
@@ -1357,8 +1357,8 @@ void DOSBOX_SetupConfigSections(void) {
         0
     };
 
-#if defined(__SSE__) && !(defined(_M_AMD64) || defined(__e2k__)) && !defined(EMSCRIPTEN)
-    CheckSSESupport();
+#ifdef __SSE__
+    CheckX86ExtensionsSupport();
 #endif
     SDLNetInited = false;
 
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -125,21 +125,37 @@ static void RENDER_EmptyLineHandler(const void * src) {
 }
 
 /*HACK*/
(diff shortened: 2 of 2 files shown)


yxmline:
https://github.com/joncampbell123/dosbox-x/commit/f99de613c609713864417c275bdc7e90802fcb57
render.obj : error LNK2001: unresolved external symbols "bool avx2_available" (?avx2_available@@3_NA)
build\dosbox-x\vs2015..\bin\x64\Release SDL2\dosbox-x.exe : fatal error LNK1120: 1 external command that cannot be resolved
vs2019 x64 build sdl2 fatal,x86 sdl2 build is ok
my cpu is i7 3770 Avx2 is not supported

linkmauve:
Thanks for testing on MSVC, this should now be fixed.

Wengier:
@linkmauve I have checked the new code for all architectures (x86/x64/ARM/ARM64, SDL1 & SDL2) in Visual Studio 2019. All seem to build fine except for x64 SDL2. Could you please fix x64 SDL2 build too?

dosbox.cpp(124,5): error C4235: nonstandard extension used: '__asm' keyword not supported on this architecture

linkmauve:
I’ve only tested in that configuration. ^^'

I couldn’t get autotools to work until the end on AArch64 for instance.

PR #1885, opened 2020-09-28
Add support for out-of-tree builds

linkmauve:
# Description

There were very few changes required to support this feature of
autotools, namely replacing `pwd` with $srcdir, and using it everywhere
required.

```
../configure: line 11293: vs2015/config_package.h: No such file or directory
```

The ability of being in any arbitrary directory and running the `configure` script from there, thereafter building everything in the current directory.

This also allows multiple build directories to exist at once for a given source checkout.

None that I know of.

**Additional information**

N/A


Wengier:
Thanks for the fix!

PR #1889, opened 2020-09-28
Replace integer aliases with stdint integers

linkmauve:
# Description

This simplify the code for newcomers who are accustomed to C, using the standard `cstdint` types instead of `Bit<size><signedness>` types. The `Bits` and `Bitu` aliases have been kept, I aliased them to `intptr_t` and `uintptr_t` respectively for now, but doing a mass-renaming to them didn’t seem correct, as many uses of them were better served with `size_t`/`ssize_t` or even `ptrdiff_t`.

Readability mostly.

None.

I hope not.

**Additional information**

Instead of reviewing each commit separately, you may want to use `sed s/Bit8u/uint8_t/g (src|include)/**/*.(c|cpp|h)` in zsh (which is how I generated these commits) and compare the results with each commit.


Wengier:
Great work @linkmauve, and thanks for the improvements! Meanwhile, since it involves changes to a lot of files and the scheduled release date for the 0.83.6 version is coming in 2 days, I would recommend the code to be merged shortly after that just to make sure it does not break anything. Thanks again for the work!

linkmauve:
Sure, I have no problem with waiting a few days (or weeks if you ever need) before merging something. :)

If you need to make changes before this gets merged, I’ll happily rebase it, as I said it was just a few `sed` commands.

Wengier:
@linkmauve Yes, I would certainly appreciate for any quick bugfix before the release!

Wengier:
@linkmauve DOSBox-X 0.83.6 is already released, so I am merging this now. Keep up the good work!

PR #1899, opened 2020-10-02
Remove extra VS2013 files

Jookia:
# Description

This removes the old VS2013 project file and config.h.backup.

No

No

This will break building with VS2013 if it works.

**Additional information**

It's a bit confusing to have it around when adding new features or refactoring


joncampbell123:
Before you do this ask the one contributer I can think of, @yksoft1 that still might use these.

Jookia:
I'm interested to hear from them. Though I do want to point out that even if these are in use, they're out of date compared to the vcproj.

aybe:
Personally, I'd merge as [its last modified date is Feb 23, 2016](https://github.com/joncampbell123/dosbox-x/commits/fe0974b9a37b085d81a946f2ad9ad2d76008143a/vs2015/dosbox-x.vcproj), project substantially changed since then that it's obvious it's not in use anymore.

Jookia:
Yes, that's my thinking as well. Keeping it would at least mean maintaining it and updating it to build.

joncampbell123:
Good point. If @yksoft1 still needs to compile with VS2013 he's going to have to re-backport the project and solution files again. I'll take the lack of updates as a sign he doesn't need it anymore. He may be happy with MinGW compiles, I'm not sure. I'd like @yksoft1 to jump in if possible and comment on it.

aybe:
Okay, let's wait then, there's no hurry after all.

joncampbell123:
No response from @yksoft1 

Time for them to go.

PR #1903, opened 2020-10-02
Compile fix

Allofich:
With current master (db1d59c), the Release x64 bit configuration (SDL1) fails to build in the latest VS2019 (Version 16.7.5).

The first in the list of compiler errors that appears is:

dosbox-x\src\output\direct3d\hq2x_d3d.h(31,88): error C2061: syntax error: identifier 'uint8_t'

By including stdint.h, where uint8_t is declared, into this file, compilation is fixed for me.

### DIFF
--- a/src/output/direct3d/hq2x_d3d.h
+++ b/src/output/direct3d/hq2x_d3d.h
@@ -23,6 +23,7 @@
 
 #include "config.h"
 #include <math.h>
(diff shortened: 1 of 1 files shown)


tbraun-de:
I can confirm the issue with the VS2019 build, and can also confirm that the change from this PR fixed the issue for me, too.

PR #1905, opened 2020-10-02
Fix warnings in the SDL2 build

linkmauve:
# Description

This fixes some warnings, this is best read commit per commit.

The POD one especially could cause the constructor pointer to be overwritten with `NULL` since the memory layout of non-POD objects is undefined.

A few warnings when building with SDL2 on Linux.

None.

When passing -defaultdir and `chdir()` fails, dosbox-x now exits with an error instead of staying in the current directory.

**Additional information**

N/A


Wengier:
Thanks for the fixes!

PR #1908, opened 2020-10-03
Fix ref-nmake build

Jookia:
# Description

This makes the ref-nmake build work again.

No

No

No

**Additional information**

I'm not sure if this is too many OBJs at once, they all seem to be separated in to multiple variables somewhat arbitrarily.


Wengier:
Thanks for the fixes!

PR #1911, opened 2020-10-03
MinGW: Support compiling with pcap

Jookia:
# Description

This lets MinGW support compiling with pcap. The configure.ac will refuse to build it if libpcap is not installed system-wide. On Windows we just need the headers, so we can always compile with it.

No.

pcap on Windows when compiled with MinGW?

None

**Additional information**

**I haven't tested running this as I'm getting other MinGW errors.** I may need to append more commits to get pcap running with MinGW, but this commit is currently required regardless.


Jookia:
Okay, I can confirm this works. The other MinGW errors are related to SDL2.

Wengier:
I appreciate the improvement. Keep up the good work!

PR #1914, opened 2020-10-03
Fix some Visual Studio warnings

Allofich:
Visual Studio builds currently have a lot of warnings, many of which come from the FluidSynth code. Some also come from newly added warnings in Visual Studio, and as usual there are the usual type-conversion warnings.

This PR reduces the number of warnings from the 649 of current master branch to 567.

One new warning, C4996, (https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=vs-2019#posix-function-names) appears for some functions like `chdir`.

The warning suggests changing these to `_chdir`, which is a name change for compliance with "C99 and C++03 constraints on reserved and global implementation-defined names".

However, the warning's page also notes:

"To fix this issue, we usually recommend you change your code to use the suggested function names instead. However, the updated names are Microsoft-specific. If you need to use the existing function names for portability reasons, you can turn off these warnings."

so I decided to just disable this warning.

### DIFF
--- a/include/render.h
+++ b/include/render.h
@@ -57,7 +57,7 @@ typedef struct {
 	Bitu last;
 } RenderPal_t;
 
-typedef struct {
+typedef struct Render_t {
 	struct {
 		Bitu width, start;
 		Bitu height;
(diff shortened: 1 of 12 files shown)


Wengier:
The function _chdir is indeed Windows only, so I agree with just disabling this warning for Visual Studio.

PR #1922, opened 2020-10-04
SVN commits

Allofich:
Filling out missing intermediate commits from SVN

Commits added:
https://sourceforge.net/p/dosbox/code-0/4257/
https://sourceforge.net/p/dosbox/code-0/4258/
https://sourceforge.net/p/dosbox/code-0/4261/
https://sourceforge.net/p/dosbox/code-0/4262/
https://sourceforge.net/p/dosbox/code-0/4264/
https://sourceforge.net/p/dosbox/code-0/4265/

Commits that were skipped were either because of conflicts (modified code is different so the SVN changes can't be applied as they are and may not be relevant) or because they were irrelevant to DOSBox-X.


Wengier:
Looks like the SVN is adding a new command to the Z drive.

PR #1927, opened 2020-10-05
Update wiki url

rderooy:
Change links to point to the new wiki at https://dosbox-x.com/wiki


Wengier:
@rderooy The welcome banner was slightly broken because of the Wiki URL change. I have already fixed it in the latest code.

Wengier:
By the way, we can continue to work on the draft Wiki (https://github.com/wengier/dosbox-x-wiki/wiki) as before, and by now it becomes pretty easy to periodically transfer the content from the draft Wiki to the new Wiki (https://dosbox-x.com/wiki). The previous main Wiki (https://github.com/joncampbell123/dosbox-x/wiki) can perhaps note that it is recommended to use the new Wiki site.

rderooy:
That is indeed a good idea. I assume no redirect is possible from the
github wiki link?

Wengier:
I indeed assume so at this time. Feel free to let me know if you (or anyone else) find a way to redirect the Wiki.

PR #1937, opened 2020-10-08
Add windowposition config option

tomba4:
# Description

Normally, at least on Windows, the window open with SDL1 is positioned randomly on the screen, which is pretty annoying. It can be overriden by setting SDL_VIDEO_WINDOW_POS, but it's cumbersome.

SDL2 is even worse, as SDL_VIDEO_WINDOW_POS does nothing and the window always ends up in the same, weird place on the screen (top left corner is around the middle of the screen on my main 4k monitor, resulting in bottom right corner outside of the visible area).

Added "windowposition" option to allow users to configure the exact placement of the window when DOSBox starts. If the config value is empty, no change will be made.

No

 "windowposition" configuration option.

No.

**Additional information**

I saw the DOSBox code for the first time only yesterday, so please bear with me if I missed something obvious.

### DIFF
--- a/dosbox-x.reference.conf
+++ b/dosbox-x.reference.conf
@@ -11,6 +11,7 @@
 #                        If you end up with small window on a large screen, try an output different from surface.
 #  windowresolution: Scale the window to this size IF the output device supports hardware scaling.
 #                        (output=surface does not!)
+#    windowposition: Set the window position at startup.
 #            output: What video system to use for output.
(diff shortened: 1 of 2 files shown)


tomba4:
Note to me: next time build *both* SDL1 and SDL2...

Wengier:
This is a pretty useful feature for users I think. Thanks for implementing it!

PR #1958, opened 2020-10-20
Use reverse DNS naming convention

rderooy:
Use reverse DNS naming for .desktop and .metainfo.xml as your
supposed to do nowadays according to freedesktop.org.

This also happens to match what is mandated by flatpak, so it
simplifies the flatpak post install script a little.

+ cleanup the mess that flatpak builds can leave behind when you
do a clean.


Wengier:
Thanks for the cleanup!

Issue #1966, opened 2020-10-28
Emulated pentium cpu stepping


rderooy:
**Please describe your question.**
Setting ``cputype=pentium``, and running a CPUID tool: http://web.inter.nl.net/hcc/J.Steunebrink/chkcpu.htm

Reports a Pentium CPU with FDIV bug. Should we not emulate a CPU without FDIV bug?
![command_000](https://user-images.githubusercontent.com/17701098/97411214-4009a380-1900-11eb-8c60-e31573e71394.png)

I'm guessing it is because of the reported CPU stepping. Both B1 and C1 steppings have the bug, while the D1 stepping is fixed.

joncampbell123:
What's wrong with emulating the FDIV bug? :)

(probably not going to happen)

Wengier:
@joncampbell123 Is there an actual Pentium CPU without the FDIV bug? If so, perhaps there can be an advanced config option to control whether to emulate the said bug (probably false by default)?

rderooy:
Ohh absolutely. It mainly effected the original 60 and 66MHz CPU's. And
Intel provided a free replacement CPU to those that wanted one.

joncampbell123:
I think there were some 100MHz versions made with the bug too. At least I remember mine back in the day having something of that problem, though it didn't bother me at all. At most it might have caused the textures in Quake to warp a little more when viewing walls at extreme angles :)

rderooy:
Yeah, there are some early 3.3V models that also had the bug.

Wengier:
There is now an advanced config option "report fdiv bug" (in [cpu] section) which when set to true will report this bug.

PR #1967, opened 2020-10-28
Change P5 stepping to D1

rderooy:
# Description
Original Pentium B1 and C1 steppings have the FDIV bug, and reporting such a stepping can therefore make software think we have the FDIV bug, so change to the fixed D1 stepping.

https://github.com/joncampbell123/dosbox-x/issues/1966


rderooy:
Note: I'm not aware of any cases where the currently reported B1 stepping causes problems. But since we don't emulate the FDIV bug, it seems prudent to reported a fixed stepping level.

Wengier:
Thanks for the fix! I will also add an advanced config option to optionally allow it to be emulated.

PR #1970, opened 2020-10-28
Set cpu-id to a 486sx when fpu=false

rderooy:
# Description
Right now the CPU-ID is always reported as a 486DX when emulating a 486 class CPU, even if the FPU is disabled.

This fixes it, by reporting the CPU-ID of a 486SX CPU when the FPU is disabled.


Wengier:
Thanks for the fix!

PR #1965, opened 2020-10-26
Recent SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/4374
https://sourceforge.net/p/dosbox/code-0/4375
https://sourceforge.net/p/dosbox/code-0/4376 (only part of this could be applied, the other modified code non-existent in DOSBox-X)
https://sourceforge.net/p/dosbox/code-0/4378
https://sourceforge.net/p/dosbox/code-0/4379
https://sourceforge.net/p/dosbox/code-0/4382
https://sourceforge.net/p/dosbox/code-0/4383 (correction to 4382, combined here)
https://sourceforge.net/p/dosbox/code-0/4384

Skipped:
https://sourceforge.net/p/dosbox/code-0/4381 (conflicts with DOSBox-X code. @joncampbell123 maybe take a look?)

Also, the next commit
https://sourceforge.net/p/dosbox/code-0/4385
has some conflict with DOSBox-X-added things. @joncampbell123, this also looks like you would know best what's going on.

Also, @joncampbell123, you might be interested that https://sourceforge.net/p/dosbox/code-0/4384 is supposed to fix "quite a few demos."

PR #1973, opened 2020-10-31
SVN r4386

Allofich:
https://sourceforge.net/p/dosbox/code-0/4386/

(https://sourceforge.net/p/dosbox/code-0/4385/ is skipped because of some conflicts, as I mentioned in
https://github.com/joncampbell123/dosbox-x/pull/1965)

@joncampbell123 
DOSBox-X has some minor differences in the mpu401.cpp, where it uses `mpu.clock.tempo_rel` (multiplies by it, then divides by 0x40).

The `PIC_AddEvent` in `MPU401_Event`, which is modified by this commit, was using `mpu.clock.tempo_rel`, so I made it still do so in this PR.

But another similar-looking `PIC_AddEvent` that was modified by this commit was not using `mpu.clock.tempo_rel`, so I also kept it that way, but should it use `mpu.clock.tempo_rel`? And, the SVN commit then added one more `PIC_AddEvent`, but should that use `mpu.clock.tempo_rel`? Right now it does not.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -203,6 +203,10 @@
     itself after an E_EXIT occurs. (Wengier)
   - Updated MPXPLAY to latest version 1.65. (Wengier)
   - Integrated SVN commits (Allofich)
+    - r4386: Correct MPU-401 clock-to-host function to
+    operate independently of playing. Fixes hang at
+    exit when playing Roland music in demos by The
+    Phoney Coders. Clamp tempo to valid range.
     - r4384: ICW1 on the PIC clears the Interrupt Mask
     Register. Fixes Antagony and quite a few demos that
     expect IRQs to be unmasked.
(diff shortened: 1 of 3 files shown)


joncampbell123:
If you're referring to mpu401.cpp line 167, then it should probably multiply by tempo_rel and divide by 0x40 in the same way it does on line 574 at the tail end of MPU401_Event().

Where is the additional PIC_AddEvent going to go?

joncampbell123:
Also considering the nature of the patch, I wonder if this will solve some issues with certain demoscene entries and PC-98 games that seem to have trouble detecting the MPU-401 or using it.

There are a handful of MS-DOS demoscene entries in the 1991-1993 time frame that are able to use the MPU-401, but if they do, they hang waiting for some condition at program exit.

Allofich:
>Where is the additional PIC_AddEvent going to go?

The other one is within `MPU401_WriteCommand`, under case 0x95.

I've gone ahead and made all three of the `PIC_AddEvent` calls do a multiply by tempo_rel and divide by 0x40, on the assumption that this is correct.

If the case 0x95 one is different for some reason, I can change it back.

Wengier:
Thanks for merging the code!

PR #1985, opened 2020-11-06
Enhanced MODE command to change screen lines/columns

Wengier:
I have enhanced MODE command so that you can change the number of columns and lines in the screen with the syntax "MODE CON COLS=c LINES=n" (c=80 or 132, and n=25, 43, 50, or 60), similar to a real DOS system. The command "MODE CON" will on the other hand show the current number of columns and lines as in MS-DOS. Also did minor code cleanup in the Configuration Tool.


joncampbell123:
The new code seems to work fine, GCC though does spit out quite a few compiler warnings.

Though it's not the highest priority to do so, compiler warnings should be dealt with carefully. Otherwise it's just added noise on top of warnings and errors that might be important to pay attention to.

joncampbell123:
I see Linux builds now have the OpenGL shader selection menu item. It's enabled if you start DOSBox-X with OpenGL output, but if you start with surface output and then change to OpenGL it does not enable.

joncampbell123:
I have to type in the path manually on the command line but even on Intel UHD graphics chipsets Linux and MesaGL are able to run shaders.

It would be 100% complete for Linux users if it was still able to render the OpenGL menu bar as it does (bypassing the shader for that case).

glslshaderswork

Wengier:
@joncampbell123 True. At this time the OpenGL shaders are only initialized when you start DOSBox-X with OpenGL output, but not if you start with a different output and then change to OpenGL. This issue can certainly be fixed to make it more flexible.

But it appears that the SDL draw menu is not compatible with OpenGL menu bar at this time as discussed earlier. Not exactly sure how to resolve this.

joncampbell123:
Regarding the menus, perhaps there's a way for one compiled fragment (the screen) to run with shaders, while turning off shaders during the menu draw routines. I'll have to play around with that on my system.

Issue #1996, opened 2020-11-13
mode with cols=132 causes dir/w to mess up


rderooy:
**Describe the bug**
Setting ``mode con cols=132 lines=60`` and running ``dir/w`` causes the output to be messed up. ``ls`` on the other hand has no issue, and neither does dir when using 4dos.

![command_000](https://user-images.githubusercontent.com/17701098/99059835-66396f80-259f-11eb-9d38-744f8773c047.png)

Also, question. Are you not supposed to be able to break a DIR output? I did a dir /s of a very large directory tree and neither CTRL-C nor CTRL-BREAK had any effect. Likewise with the ``tree``command.

Wengier:
@rderooy Thanks for reporting the bug with DIR /W. I think I have already fixed it in my pull request.

And you are right, it will be nice to allow break from a long DIR output as well.

PR #1998, opened 2020-11-14
Fix issue #1996 and improve TTF output

Wengier:
As reported in Issue #1996, there is a bug with DIR /W when the number of columns is greater than 80. So I have fixed it.

Also made various improvements to the TTF output, such as mouse copy/paste support, the ability to switch to a different output (either manually or automatically), and the ability to change the number of lines and columns in this output with the MODE command.


joncampbell123:
@Wengier Do you mind if I turn the ttf.rows and ttf.lins options so that the default is zero and the TTF output follows the programmed character width and height of the VGA output?

It would allow it to work with both 40x25, 80x25, or any custom text mode programmed by the DOS application.

Wengier:
@joncampbell123 Sure, you can of course change it in any way you want, including other fixes such as the SDL drawn menu issue. I have been considering exactly the same thing for the ttf.rows and ttf.lins options, but have not started to do it yet.

joncampbell123:
Will do. I've already modified it to make room on my system for the SDL drawn menu, though the menu is not being drawn automatically.

PR #2007, opened 2020-11-19
Some fixups to the PC98 and video output

Wengier:
I have made some additional fixups to the PC98 mode, as well as other small updates to the TTF output to make it work better. The code in the vga_draw.cpp that was temporarily disabled has been reactivated since it now works with a fix applied.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -43,6 +43,7 @@ A list of features ported from DOSBox Staging:
 A list of features ported from vDosPlus by Wengier:
 
 * Long filename support (improved for FAT drives since then by Wengier and joncampbell123)
+* TrueType font (TTF) output support (originally by Jos Schaars and heavily improved since then by Wengier)
 * Improved support for automatic drive mounting (Windows)
 * Support for clipboard copy and paste (improved since then by Wengier)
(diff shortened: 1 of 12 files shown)


joncampbell123:
Regarding PC-98 mode, you're right. The text layer is 80 columns. The funny thing is that there *is* a 40-column mode, which is just the 80-column mode with each cell stretched out, and the 40-column mode skips every other character cell. But as far as I know, only games use the 40-column mode, so it's perfectly fine to assume 80 anyway. The 40-column mode will just look like every cell is doubled up.

Wengier:
@joncampbell123 I see that line 3002 of bios.cpp apparently assumes that there are 80 columns in PC-98 mode. For the rows there is a pc98_DOS_console_rows function to return the number of rows, but for the columns it seems to be fixed to 80 right now. When (if) 40-column support is eventually added for PC-98 mode, this code will be updated accordingly too to support it of course.

joncampbell123:
It's such a strange mode that not even the DOS kernel (where the console driver lives) supports it. If it's used at all, it's in some games atop the graphics layer, which the TTF output does not render.

PR #2018, opened 2020-11-22
Improve PC-98 cursor and TTF output

Wengier:
I have improved the cursor in the DOS shell so that it will move among double-byte characters instead of by byte in PC-98 mode (any output). Also improved TTF output such as allowing to specify a output to switch to instead of always switching to the surface output, along with some cleanups to the feature.


joncampbell123:
Text selection... I thought I missed something in ttf_cell. :+1:

Wengier:
@joncampbell123 Yes, the new structure makes it much easier to modify than before. Well done!

PR #2038, opened 2020-11-29
Look for TTF fonts in program directory and default font path too

Wengier:
This allows TTF fonts and the DOSBox-X config file (dosbox-x.conf/dosbox.conf) to be searched in the DOSBox-X executable directory if they cannot be found in the current working directory, which is done in the cross-platform manner thanks to the library. Also, DOSBox-X will search for default font path (/usr/share/fonts/truetype on Linux and /Library/Fonts in macOS) in addition to C:\WINDOWS\fonts on Windows. Tested in Windows, Linux, and macOS. There are also some smaller fixups.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -42,6 +42,10 @@
     to the change, ancient versions of the API dating
     back to the mid 2000s were used which no longer
     work on Big Sur.
+  - DOSBox-X will now look for the dosbox-x.conf (and
+    dosbox.conf) file in the directory containing the
+    DOSBox-X executable too if the config file cannot
+    be found in DOSBox-X working directory. (Wengier)
   - The system menu in Windows SDL1 builds is now also
     available for Windows SDL2 builds, and menu items
     "Reset font size", "Increase TTF font size" and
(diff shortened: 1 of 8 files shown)


rderooy:
Just want to point out there are a few potential locations for fonts on Linux.

According to /etc/fonts/fonts.conf on my fedora system, these are the locations to check:

- /usr/share/fonts
- /usr/share/X11/fonts/Type1 (not relevant here)
- /usr/share/X11/fonts/TTF (I think you can skip this also, since it is just standard X11 fonts)
- /usr/local/share/fonts
- ~/.local/share/fonts
- ~/.fonts (deprecated)

rderooy:
Also, perhaps this same code should be used to find the printer fonts.

Wengier:
@rderooy There is a /usr/share/fonts directory in my Linux. but no TrueType font files can be found in this folder.

rderooy:
@Wengier correct, that directory is meant to "locally installed files by sysadmin".

Also, for ``~/.local/share/fonts`` you should first check if the XDG_DATA_HOME environment variable is set, and use ``XDG_DATA_HOME/fonts`` if it is. And only if the environment variable is not set (or blank), you should use ``~/.local/share/fonts``

Issue #2046, opened 2020-11-30
HX-DOS target build issues


joncampbell123:
**Describe the bug**
Build failure for HX-DOS build using MinGW

One error is caused by the use of std::thread in the CHD image support.

The other is caused by an undefined datatype used in FLAC support.

The quickest way to restore HX-DOS build capability is to conditionally comment out CHD image and FLAC audio support in HX-DOS builds.

Anyone in addition to @yksoft1 interested in this project continuing to make HX-DOS builds?

joncampbell123:
[hxdosbuildfail.txt](https://github.com/joncampbell123/dosbox-x/files/5618390/hxdosbuildfail.txt)

Wengier:
As far as I know it is the only official DOSBox-X build that will work in DOS (and Windows 9x/Me/2000), so I think it is something nice to have. I will check the errors in the log, and in fact I think it is better for me to have an environment to periodically check for the HX-DOS build to ensure it compiles.

joncampbell123:
@Wengier I am using a Windows XP VM with the original MSYS and MinGW compiler, because the original MSYS has occasional concurrency problems on Windows 10, and it compiles FASTER in the VM than on the host Windows 10 system.

It needs to be the original MinGW which can target all the way down to Windows 98, and even Windows 95 as long as you ship MSVCRT.DLL with your executable. The compiled binary is quite compatible with HX-DOS.

The MinGW64 project only seems to be able to target Windows XP/2000 or higher.

Wengier:
@joncampbell123 I have already made some fixes to the code for the HX-DOS target. The CHD image support is still enabled but prefetch (which requires thread) will be disabled for it.

Tested with the original MinGW. I think it is pretty slow here in both Windows 10 and Windows XP VM, thus it takes a lot of time to do a full compile.

PR #2048, opened 2020-12-01
Fix for HX-DOS build (Issue #2046)

Wengier:
Per issue #2046, I have made some fixes to the current code so that the HX-DOS target will hopefully build.


joncampbell123:
It compiles now :+1:

Issue #2083, opened 2020-12-12
Add LPT4-9?


emendelson:
Many years ago, I tried adding LPT4 through LPT9 to DOSBox-X, without success. I hope someone who knows something about coding might consider this now.

Under vDos, I can use LP1 through 9, and use the output files to each port in different ways (print to different printers, pop up a select-printer dialog, create PDF files in different locations, etc. DOSBox-X only allows the standard LPT1-3.

I hoped I could use COM1-4 for this purpose, but programs like WPDOS probe the COM ports if they exist, and produce undesirable results when the ports are emulated.

DOSBox-X already includes the CLIP device. Is there any hope that it might also add LPT4-9 as devices?

Wengier:
@emendelson I have already added the option to optionally add LPT4-9 devices as you suggested, similar to LPT1-3. Since they don't have default BIOS base addresses, you will need to set one for them by yourself if you enable these ports, e.g. 

```
parallel4 = base:2bc file file:output4.prn timeout:1000 openpcl:pcl6 ...
```

You can download the updated Windows binary from:

* [dosbox-x-update.zip](http://cndos.fam.cx/dosware/dosbox-x-update.zip)

emendelson:
Thank you, Wengier! Are there any guidelines of what base addresses will work? I don't want to specify an address that will interfere with anything else.

Wengier:
@emendelson You can try the following updated DOSBox-X Windows binary where the base addresses for LPT4-9 are now optional (but you can still specify for them; for LPT1-3 you are never required to specify base addresses for them since they have default base addresses):

* [dosbox-x-update.zip](http://cndos.fam.cx/dosware/dosbox-x-update.zip)

Hope this helps.

emendelson:
Thank you, Wengier. I just tried it out with LPT4 and it worked perfectly.

Issue #2082, opened 2020-12-11
Emulator real time clock stopped when menu dialog box is open


MX9000:
**Describe the bug**
When I typed the DOS command TIME after opening a dialog box for a few minutes (e.g. "Configuration Tool" or "About DOSBox-X"), I noticed that the emulator's internal clock lagged behind the host clock, as if it stopped while the dialog box was open.
I tried again the TIME command after a few minutes, but the emulator's clock never regained the lost time.
To synchronize the emulator's clock with host PC clock I have to reset the virtual machine.
Thank you.

**To Reproduce**
1. Leave a dialog box open for some time.
2. Close the dialog box
3. Type TIME command at DOS prompt
4. Compare emulator time with host computer time

**Expected behavior**
I think that the emulator clock and the host computer clock should always be synchronized.

**Environment (please complete the following information):**
 - Windows 10 Enterprise 64-bit version 1809
 - DOSBox-X 0.83.8 SDL2 64-bit
 - dosbox-x.conf file attached: [dosbox-x.conf.zip](https://github.com/joncampbell123/dosbox-x/files/5680642/dosbox-x.conf.zip)

Wengier:
@MX9000 Thanks for reporting this! You don't need to reset the virtual machine to synchronize the internal clock with host PC clock though. Just type ```TIME /H``` should do the job. But I agree that there can be perhaps the default option to automatically synchronize the internal clock with the host PC clock.

Wengier:
I have now added the ability to automatically synchronize the date/time with the host system by default. It can be turned off by setting the new config option"synchronize time" to "false" in [dosbox] section, or if you manually change the date/time. Hope this helps.

MX9000:
@Wengier Thank you!
I'll definitely try it in the next release.

joncampbell123:
@Wengier I had to adjust the BIOS_TIMER synchronize code due to some conflicts with demoscene stuff that relies on BIOS_TIMER while setting a higher tick rate, and I had to make "synchronize time" false by default.

The fix retains BIOS_TIMER synchronization with local time, but only if the timer is at the proper 18.2Hz rate. The fix temporarily skips synchronization if anything still uses the BIOS IRQ0 handler but at a faster tick rate (like the aformentioned demoscene stuff).

Wengier:
[...]

Issue #2081, opened 2020-12-11
Emulator real time clock too fast when software is running


MX9000:
**Describe the bug**
When I typed the DOS command TIME after playing a game, I noticed that the emulator's internal clock was far ahead compared to the host PC clock.
With some games the issue is more pronounced (hours with Crystal Caves), less pronounced with others (minutes with Xargon).
To synchronize the emulator's clock with host PC clock I had to reset the virtual machine.
Thank you.

**To Reproduce**
1. Play a game for some time.
2. Exit from the game
3. Type TIME command at DOS prompt
4. Compare emulator time with host computer time

**Expected behavior**
I think that the emulator clock and the host computer clock should always be synchronized.

**Environment (please complete the following information):**
 - Windows 10 Enterprise 64-bit version 1809
 - DOSBox-X 0.83.8 SDL2 64-bit
 - dosbox-x.conf file attached: [dosbox-x.conf.zip](https://github.com/joncampbell123/dosbox-x/files/5680642/dosbox-x.conf.zip)

Wengier:
As mentioned in Issue #2082, you don't need to reset the virtual machine to synchronize the internal clock with host PC clock though - just type ```TIME /H``` should do the job. But I agree that there can be perhaps the default option to automatically synchronize the internal clock with the host PC clock.

Wengier:
My PR hopefully solves this by doing automatically synchronizing the host date/time. See Issue #2082 for more information.

MX9000:
@Wengier Thank you!
I'll definitely try it in the next release.

joncampbell123:
@Wengier I had to adjust the BIOS_TIMER synchronize code due to some conflicts with demoscene stuff that relies on BIOS_TIMER while setting a higher tick rate, and I had to make "synchronize time" false by default.

The fix retains BIOS_TIMER synchronization with local time, but only if the timer is at the proper 18.2Hz rate. The fix temporarily skips synchronization if anything still uses the BIOS IRQ0 handler but at a faster tick rate (like the aformentioned demoscene stuff).

Wengier:
@joncampbell123 I did see the changelog and could understand there are some concerns for the time synchronization. Apparently the time synchronization function is not yet perfect. For example, I was aware that it has no effect for the PC-98 mode. If there turns out to be more concerns I would support it to make it disabled by default, but users who desire it can always enable it if they want, either from the config file or from the “DOS” menu.

joncampbell123:
[...]

Issue #2078, opened 2020-12-09
Really Bizarre Filesystem Issue


sduensin:
**Describe the bug**
Using OpenWatcom to compile a program, the results are written in totally the wrong place:
- Everything is a subdirectory under `C:\fltkpico`
- Source code is in `src`
- Compiled object code should be written to `obj`

But...

- A zero-byte file is written to `obj`
- The actual obj output is written into the source file!

**To Reproduce**
Unpack the attached tarball.   (Rename, please, GitHub didn't like my bz2 file.)  My configuration and everything are included.  The `B.BAT` in the root begins the build.

**Expected behavior**
`fltk.lib` generated in the `lib` folder.  Confirmed this works under actual DOS (Windows 98).

**Screenshots**
No screenshot.  DOSBox-X eventually blows up.

**Environment (please complete the following information):**
DOSBox-X 0.83.8 on Linux Mint 19

**Additional context**
Never seen anything like this before.  :-)

Not really a zip:
[dosbox-x.tar.bz2.zip](https://github.com/joncampbell123/dosbox-x/files/5669357/dosbox-x.tar.bz2.zip)

Wengier:
@sduensin I have tried your package and made minor modifications to the B.BAT batch file to let it run on my system. What I got is the following screen:

![image](https://user-images.githubusercontent.com/8216923/101718002-57df5600-3a6e-11eb-9237-5e45f6e95c0b.png)

Clearly wpp386 complains about the source file src\fl.cxx. And by looking at the file fl.cxx in your tarball, it is apparently not a C source file (similar for FL_Bitmap.cxx). Perhaps you did not include the original source file in the tarball?

Also, can you post the DOSBox-X configuration file you are using? Thanks!

sduensin:
Ah crap!  Yep, that's the object file that overwrote the source.  Apparently I tarred the wrong thing.  New `src` attached.  And this time, it IS a zip.  :-)


[src.zip](https://github.com/joncampbell123/dosbox-x/files/5674590/src.zip)

grapeli:
@sduensin 
Looks like a serious bug with long filenames. Certainly source files should never be overwritten with relocatable files.
From FreeDOS (doslfn) running under DOSBox-X, the compilation is done correctly and successfully, and the whole thing is really fast.

For comparison (on a really slow processor) how fast is the compilation process from the included dosbox-x.tar.bz2.zip file:
72s - dosemu2 (host filesystem)
[...]

PR #2084, opened 2020-12-13
Menu option to create blank disk images and solve issues #2078, #2081, #2082, #2083

Wengier:
@rderooy had previously suggested the ability to create blank disk images from menu in Issue #1988. So this adds a "Create blank disk image" menu option in "DOS" menu to create blank floppy or hard disk images of a common disk size, including 1.2MB,1.44MB, 2.88MB for floppy images as well as 250MB, 520MB, 1GB, 2GB, 4GB and 8GB for hard disk images.

Also solved the file system problem for Issue #2078, the time auto-synchronization ability for Issue #2081 and #2082, and custom LPT4-9 ports for Issue #2083.


joncampbell123:
@Wengier Unrelated question: Two parts of DOSBox-X now no longer compile with GCC 4.8.5. One is the use of the PRIx64 defines in inttypes.h for some reason, the other is a call to std::istringstream::swap in render.cpp which GCC 4.8.5 does not recognize apparently, despite supporting C++11 otherwise.

Does anyone here still care to compile with GCC 4.8.5, or should the project just require a later version of GCC that provides both of those C++ features?

joncampbell123:
render.cpp:965:24: error: ‘std::stringstream’ has no member named ‘swap’
             if (empty) buf.swap(tmp);

    whereami.c: In function ‘int wai_getModulePath(char*, int, int*)’:
    whereami.c:268:30: error: expected ‘)’ before ‘PRIx64’
           if (sscanf(buffer, "%" PRIx64 "-%" PRIx64 " %s %" PRIx64 " %x:%x %u %s\n", &low, &high, perms, &offset, &major, &minor, &inode, path) == 8)
    whereami.c:268:139: warning: spurious trailing ‘%’ in format [-Wformat=]
           if (sscanf(buffer, "%" PRIx64 "-%" PRIx64 " %s %" PRIx64 " %x:%x %u %s\n", &low, &high, perms, &offset, &major, &minor, &inode, path) == 8)
                                                                                                                                           ^

Wengier:
@joncampbell123 I don't have GCC 4.8.5 for testing here, but if these are the only issues for compilation with GCC 4.8.5, I don't think they are very difficult to be fixed though.

Wengier:
@joncampbell123 I hope the code in my new PR will fix the said issues with GCC 4.8.5. I don't have GCC 4.8.5 to verify myself though.

joncampbell123:
@Wengier Nope. GCC 4.8.5 considers the constructor deleted.

render.cpp: In function ‘bool RENDER_GetShader(std::string&, char*)’:
render.cpp:922:23: error: use of deleted function ‘std::basic_stringstream<char>& std::basic_stringstream<char>::operator=(const std::basic_stringstream<char>&)’
         if (empty) buf=std::stringstream();
                       ^
In file included from render.cpp:24:0:
/usr/include/c++/4.8.5/sstream:502:11: note: ‘std::basic_stringstream<char>& std::basic_stringstream<char>::operator=(const std::basic_stringstream<char>&)’ is implicitly deleted because the default definition would be ill-formed:
     class basic_stringstream : public basic_iostream<_CharT, _Traits>
           ^
/usr/include/c++/4.8.5/sstream:502:11: error: use of deleted function ‘std::basic_iostream<char>& std::basic_iostream<char>::operator=(const std::basic_iostream<char>&)’
In file included from /usr/include/c++/4.8.5/fstream:38:0,
                 from render.cpp:23:
/usr/include/c++/4.8.5/istream:795:11: note: ‘std::basic_iostream<char>& std::basic_iostream<char>::operator=(const std::basic_iostream<char>&)’ is implicitly deleted because the default definition would be ill-formed:
     class basic_iostream
           ^
/usr/include/c++/4.8.5/istream:795:11: error: use of deleted function ‘std::basic_istream<char>& std::basic_istream<char>::operator=(const std::basic_istream<char>&)’
/usr/include/c++/4.8.5/istream:58:11: note: ‘std::basic_istream<char>& std::basic_istream<char>::operator=(const std::basic_istream<char>&)’ is implicitly deleted because the default definition would be ill-formed:
     class basic_istream : virtual public basic_ios<_CharT, _Traits>
           ^

Just wondering: If buf.str().empty() then why do anything to buf to reinitialize it, which empties it anyway?

If it does need to be cleared to be certain, why not just buf.str("") to clear it (since buf.str() returns a const reference)? https://www.cplusplus.com/reference/sstream/istringstream/str/

Wengier:
@joncampbell123 I used buf.swap(tmp) to solve a problem with loading the shader, but I don't quite remember what exactly it was. I looked at the following page, which suggested a few alternatives for swap including xx.str("") and xx=std::stringstream(), and a user mentioned that "std::stringstream(); should be the most reliable way regardless of the compiler". But unfortunately it still did not work for GCC 4.8.5 as you said. I think I will use "xx.clear(); xx.str("");" instead then because another user mentions that it will solve the  stringstream stuck issue with just xx.str(""):

https://stackoverflow.com/questions/20731/how-do-you-clear-a-stringstream-variable

joncampbell123:
I'm not familiar with stringstream, what makes it get stuck?

If it's empty (because empty==true according to the test two lines up), why empty it again?

Does it just need some error bits cleared?

Wengier:
@joncampbell123 I think I should have inserted a comment in the code so that I will remember the exact reason I added that part of code at that time. Right now I only remember that I tried to reset it to the default in such case, which did solved the problem I discovered back then, and the problem itself was kind of mysterious indeed.

Issue #2106, opened 2020-12-20
Printer Font Location — Linux.


ThePillenwerfer:
Just a quick point to those in charge of documentation.

The Wiki currently states that on Linux the recommended fonts are in `/usr/share/fonts/liberation-mono/`.  However on my Debian system they are in `/usr/share/fonts/truetype/liberation`.  Maybe a note could be added saying what distro the instructions refer to and advising people with different ones to check and make appropriate links to them in `~/.config/dosbox-x/FONTS'.

I make no claims to be a programmer but looking at the code of `printer.ccp` it looks as though at one time it could read font_path from the `.conf` file but it has been commented out.  If that could be restored without messing anything up the line `basedir = "/usr/share/fonts/";` could be changed to `basedir = font_path` and the individual fonts changed from the likes of `fontName = basedir + "liberation-serif/LiberationSerif-Regular.ttf";` to `fontName = basedir + LiberationSerif-Regular.ttf";`

joncampbell123:
DOSBox-X could automatically use whatever the host system uses if it linked to and used the fontconfig libraries to enumerate them. That way it's entirely up to the font configuration file everybody else on the same system uses.

Wengier:
The current printer code will read /usr/share/fonts/liberation in addition to /usr/share/fonts/liberation-<whatever>. I added /usr/share/fonts/liberation earlier in the code as an alternative path (not yet mentioned in the Wiki) but it should be changed to /usr/share/fonts/truetype/liberation. (if font paths are more complicated than this then a library like font config may be used instead).

The printer code will also try to read fonts from the FONTS directory, and I am not sure why the fontpath option was commented out in the code. I think this option can be reactivated to allow the user to specify a custom path.

ThePillenwerfer:
#2132 has made me aware that there's a TTF in the source code.  Would it be possible to use this as a fallback of last resort, ie the times DOSBox-X currently displays an error saying it can't find FONTS/courier.ttf?  The install script would need amending to copy it over as well.

It certainly wouldn't be my choice of printing font, though it could be argued that it has a certain retro dot-matrix appeal, but would be better than nothing.  

I think this change would be within my limited capabilities but I don't know how to do Pull Requests nor if such use is permitted by the font's licence.

Wengier:
[...]

PR #2108, opened 2020-12-21
Printer font and PhysFS code cleanups

Wengier:
This makes some cleanups to the printer font code, as well as PhysFS support code. The printer font support per issue #2106, including a "fontpath" config option in [printer] section to specify the font path instead of the hard-coded FONTS directory. For PhysFS support, there was partial PhysFS code in the source tree for some time but it was very incomplete and did not work, and this will make it actually work, so that you can mount archives as drives and run DOS applications from it directly, although only in read-only mode at this time. Also fixed the Visual Studio filter not working properly.


ThePillenwerfer:
Great stuff.

PR #2143, opened 2021-01-01
Update Ubuntu build instructions

rderooy:


rderooy:
The current instructions don't work on 20.10. There are missing and broken dependencies that this fixes. With this you can compile dosbox-x SDL1 or SDL2 on a clean Ubuntu 20.10 installation.

There is one strange thing I noticed. With the SDL2 build, FluidSynth works, but not with the SDL1 version. 

When I compile for SDL2, I get the following:
```
checking fluidsynth.h usability... yes
checking fluidsynth.h presence... yes
checking for fluidsynth.h... yes
checking for fluid_synth_sysex in -lfluidsynth... yes
```

But compiling as SDL1:
```
checking fluidsynth.h usability... yes
checking fluidsynth.h presence... yes
checking for fluidsynth.h... yes
checking for fluid_synth_sysex in -lfluidsynth... no
```

PR #2164, opened 2021-01-06
Improved OPL3 Duo support

DhrBaksteen:
# Description

* This change adds a buffer thread to get rid of slowdowns and breakups in audio playback when using the OPL3 Duo board
* Fixed a bug where only the first bank of registers were cleared during reset (Second OPL3 chip is still ignored!)
* Reset the OPL chip when Dosbox is closed, so if a game is running when closing the program the OPL3 Duo is not left playing any sounds

No

No

No

**Additional information**
Tested with games using both OPL2 and OPL3 audio. 

Stunts will still overwhelm the buffer due to it writing loads of duplicate register values.

### DIFF
--- a/src/hardware/opl3duoboard/opl3duoboard.cpp
+++ b/src/hardware/opl3duoboard/opl3duoboard.cpp
@@ -1,6 +1,7 @@
 #include "../serialport/libserial.h"
 #include "setup.h"
 #include "opl3duoboard.h"
+#include <thread>
 
 Opl3DuoBoard::Opl3DuoBoard() {
 }
@@ -12,6 +13,10 @@ void Opl3DuoBoard::connect(const char* port) {
 	if (SERIAL_open(port, &comport)) {
 		SERIAL_setCommParameters(comport, 115200, 'n', SERIAL_1STOP, 8);
 		printf("OK\n");
+
+        resetBuffer();
+        stopOPL3DuoThread = false;
+        thread = std::thread(&Opl3DuoBoard::writeBuffer, this);
 	} else {
 		printf("FAIL\n");
 	}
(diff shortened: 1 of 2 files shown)


Wengier:
Thanks for the updates. I think <thread> probably needs to be disabled for the HX-DOS environment, since original MinGW does not support thread and won’t build with it.

DhrBaksteen:
Thank I will make that change. If you can let me know what would be the preferred way to define a thread so it also compiles with MinGW I can have a look at that also.

Wengier:
@DhrBaksteen Original MinGW does not support thread, so they can just be disabled. Original MinGW is only used for HX-DOS environment which does really need thread anyway.

Wengier:
@DhrBaksteen The fix for MinGW is already included in #2171. Just let you know so that you will be aware of this in the future.

DhrBaksteen:
@Wengier clear, I will take it into account. Thank you for merging.

josephillips85:
@DhrBaksteen 
i test the macos version today and is not working just run the initialization at beginning but no sound at all 
Works well on Windows version?

Thanks

DhrBaksteen:
@josephillips85 I saw that the directives added in #2171 mean there is no buffer thread when MinGW is used and thus no data will ever be sent to the board. Assuming this is the problem it's a simple fix for which I provided #2258

josephillips85:
@DhrBaksteen  i check last night the problem is not caused by MinGW directive also im fixing this for this platform.

the weird thing is adding a printf or a delay inside of the do{ on the thread make it work.. look like a race condition or something wrong with the compiler

can you confirm in Windows or Linux computer work?

DhrBaksteen:
I can check it on Windows. I don't have a Linux box anymore to try it on

Issue #2168, opened 2021-01-07
Ctrl+C makes a heart symbol


JGCYASYS:
I'm using the latest release of dosbox-x. The program I'm using is the GNU ed port to DOS via DJGPP and there is a problem with sending a break signal with Ctrl+C. It makes a heart symbol instead and I can't figure out a way to be able to exit editing mode in GNU ed for DOS. Can I have some help please.

joncampbell123:
Does CTRL+C work with anything else?

Perhaps DOSBox-X is unable to send a CTRL+C signal to protected mode applications.

JGCYASYS:
> 
> 
> Does CTRL+C work with anything else?
> 
> Perhaps DOSBox-X is unable to send a CTRL+C signal to protected mode applications.
I tested it in EDIT and outside of programs at the command line. Nothing happened in EDIT and command line gave me a ^C. You can test out the program yourself because I got it from here:
ftp://ftp.delorie.com/pub/djgpp/current/v2gnu/ed-1142b.zip
Also the source code if you need it:
ftp://ftp.delorie.com/pub/djgpp/current/v2gnu/ed-1142s.zip

grapeli:
@JGCYASYS 
ed.info
_"An interrupt (typically <Control-C>) has the effect of **aborting the current command** and returning the editor to command mode"._
You leave ed-a by command:
q

or Ctrl-D (sends an EOF to close the terminal). Doesn't work under dos.

JGCYASYS:
> 
> 
> @JGCYASYS
> ed.info
> _"An interrupt (typically ) has the effect of **aborting the current command** and returning the editor to command mode"._
> You leave `ed` by command:
> q
> 
> or Ctrl-D (sends an EOF to close the terminal). Doesn't work under dos.

I'm testing GNU ed for DOS right now in a VM for freedos 13 livecd. It works as intended like on linux. I can type "." during edit mode to return to command mode and typing Ctrl+C and then enter or Ctrl+D and enter will return to me command mode. In Dosbox-X I'm stuck in edit mode if I type . or Ctrl+C or Ctrl+D. I'm convinced now this is a bug because it works just like linux under freedos but not dosbox-x.

grapeli:
@JGCYASYS 
`ed` does not work properly under dosbox-x.
Neither command is functioning.
`a` [append]
`s` [substitute]
`w` [write]
`q` [quit]

[...]

Issue #2159, opened 2021-01-05
Is it possible to "press" two keys at once with ADDKEY or AUTOTYPE?


Banjo-Oz:
I was looking for a keyboard stuffer when I discovered DOSBox-X already includes the utilities ADDKEY and AUTOTYPE.

Is there a way to press two keys at once (or hold down one key then press the other) using those programs?

For example, to send "F11+f" (toggle fullscreen) via ADDKEY or AUTOTYPE as part of a batch file.

I can't seem to figure out how to do it, if it is possible.

Wengier:
I think this is certainly an interesting question. I don't really think it is possible to press two keys at once with either command at this time, but I guess allowing holding down one key then pressing another key is a good suggestion. Alternatively, perhaps we can allow these commands to activate mapper functions, such as the toggle fullscreen function. Good suggestion!

Banjo-Oz:
Thanks for looking at this and considering it! I actually had tried SCANCODE, KEY-FAKE and a couple of other old DOS-era keystuffers first and couldn't get it to work, then I discovered the built-in DOSBox-X tools (ADDKEY and AUTOTYPE) but as I said, couldn't get those to work either.

To be honest, the F11+f combo is a bit wonky on my machine anyway... I need to be fairly "deliberate" when I press and hold F11, then press f. It feels a bit more sensitive than the usual ALT+ENTER for some reason.

Regardless, just having a "FULLSCREEN" and "WINDOW" command would be really, really helpful.

I can set my other config settings right from a batch file (e.g. aspect from false to true, fullscreen to true) but actually making the window move to fullscreen can't be done without F11+f right now, it seems.

rderooy:
There is a config command for this, but it has no effect once dosbox-x is started.
```
config -set fullscreen=true
```

Wengier:
@Banjo-Oz With updated code in #2171 you can now turn on full-screen mode with a command like ```config -set fullscreen=true``` and ```config -set fullscreen=false```. Hopefully this will do exactly what you want.

By the way, the pixel-perfect scaling as you desired is already working. You probably want to check out #2167 for more information.

Banjo-Oz:
Wow, that's awesome news! So setting it via config will actually toggle my window to fullscreen (and back) and not just set the config setting? Previously, it didn't work (set the setting, but needed a reboot or F11+f). Is this in a build I can test for you?

Wengier:
[...]

PR #2171, opened 2021-01-08
Update version number and fix Issue #2168 etc

Wengier:
This updates the version number from 0.83.9 to 0.83.10, and made some fixes such as issue #2168 (Ctrl+C makes a heart symbol) and HX-DOS build. The commands "config -set fullscreen=true" and "config -set fullscreen=false" now turn on and off the full-screen mode respectively, per suggestion in Issue #2159.


Jookia:
Hard object to the pcaptimeout option. Users shouldn't ever need to configure this. Just set it to 10ms or something on Windows. -1 is an invalid value.

Wengier:
@Jookia A value of -1 seems to mean “non-blocking mode” for Windows, but perhaps only valid for Windows. A value of 3000 may cause freeze when DOSBox-X starts on the Windows system, but it seems to work otherwise. In any case, I think different values are probably needed for different OSes. The pcaptimeout option was added as an advanced config option, not a common option. There is a distinction between them.

Jookia:
Can you provide documentation or an API guarantee of what -1 does?

Wengier:
@Jookia I don't think a timeout value of -1 is part of the official cross-platform APIs, but it appears to be part of the de-facto implementations of Windows-side pcap, so it likely only works for Windows and will cause problems with platforms like macOS. On Windows however, I found that even setting to 10ms may cause noticeably slowdown in the DOS shell (easily checked with  pcaptimeout option), but setting to -1 will work. So I believe -1 may be considered a special case only for the Windows platform.

Jookia:
Hmm. That makes sense, but it seems like a bug in WinPcap. The timeout isn't how long it waits to use pcap, it's how long it waits for a device. I guess it's fine now that it's merged :(

Wengier:
@Jookia For the timeout, I think the description was intended to say how long it waits for a device *for the pcap feature/backend* (in contrast to the coming slirp backend), not how long it waits to use pcap itself. But I agree that the description can be further expanded to make it more clear, and it will definitely come.

Jookia:
Oh yes, I understand that. But -1 still gives you a device instantly, so the device is ready. It just sits around doing nothing for a bit if you specify a timeout?

Wengier:
@Jookia It appears that (despite the cross-platform API) WinPcap interprets -1 as "non-blocking mode", so it will use any data available but will never block. But if I specify a positive timeout it may actually block and cause slowdown. I believe the original NE2000 pcap code was written for Windows, which was why it worked well for Windows since the beginning, but may not necessarily be so for platforms like macOS.

Jookia:
Okay, yeah that makes sense. Could you document that in the code since nowhere in the Internet is this explained? Given that it's undefined behaviour basically, having a pcaptimeout option is good is this thing bugs out.

Thanks for that change now that you've explained it. :)

Wengier:
@Jookia Sure! It can be documented in the code to explain why these default values are choosen. Thanks for the suggestion.

Issue #2174, opened 2021-01-08
Adding function to translate drop-down menus


Dalekamistoso:
Hi :)

It would be a very good idea adding a way to translate the top bar without having the modify the code and recompile the whole thing.

In the same way as using "messages.txt" for dos translation

I would suggest a similar option that reads a text file as "menus.txt" or something similar

Thanks again for your time :)

Wengier:
Yes, I agree this is something nice to have. Thanks for the suggestion!

Wengier:
@Dalekamistoso I have already added the ability to translate drop-down menus. All menu texts will still be written to the standard messages.txt file, instead of a separate file. If you use Windows, the updated Windows binary package is available from:

* [dosbox-x-update.zip](http://cn-dos.fam.cx/dosware/dosbox-x-update.zip)

Just write the message file again and do the translation:-)

Dalekamistoso:
Thank you very much :)

Can you update the link to updated build? The link you include is offline.

Also, I have a question.

I would like to contribute to the project uploading the translated files as soon as I check that are properly functional.

should I add those files in a separate repository by my own or ask for a pull request on web interface or simply send it to you?

Thanks again in advance :)

> @Dalekamistoso I have already added the ability to translate drop-down menus. All menu texts will still be written to the standard messages.txt file, instead of a separate file. If you use Windows, the updated Windows binary package is available from:
> 
> * [dosbox-x-update.zip](http://cn-dos.fam.cx/dosware/dosbox-x-update.zip)
> 
> Just write the message file again and do the translation:-)

rderooy:
You can use the 'fork' button in the top right, to create a personal fork of the project. You can then modify that fork and once your ready you can submit your changes back.

But we don't have a location in the tree right now for translations, so I would propose we create directories as such:
```
/contrib/translations/de
/contrib/translations/es
/contrib/translations/fr
/contrib/translations/nl
```
etc...

And then inside the language directory create files like:

```
fr_BE.lng
fr_CH.lng
fr_FR.lng
```
Similar to how DOSBox staging does it.

Dalekamistoso:
Ok, I will create those as soon I have a working build in which I can translate full menus to start the work :)

Wengier:
@Dalekamistoso The download link should be here:

* [dosbox-x-update.zip](http://cndos.fam.cx/dosware/dosbox-x-update.zip)

Please try it again. Thanks!

[...]

Issue #2173, opened 2021-01-08
"messages.txt" Broken translation system


Dalekamistoso:
Hi, the problem is the following:

I am working on a spanish translations for DosBox-X and I found that some texts are not replaced with the ones in "messages.txt".

All tests are being made under Windows 10 LTC x64 version (spanish language)

I generated a clean language file using the gui, following the procedure described  in the documentation:

Main->Configuration Tool->Configuration->Save language file->(I introduced default name "messages.txt")

I am editing the file "messages.txt" using OEM-US text codification plus two more in order to check mistakes. 

But the text chains contained inside the section

":SHELL_STARTUP_BEGIN2" 

Are ignored and no matter what I write, DosBox-X ignores it and  force its internal english names breaking the texts.

I attach you a screenshot for better explanation, I use a default dosbox.conf config file with the only modification of:

language=message.txt

![DOXBOX-X BUG001](https://user-images.githubusercontent.com/67012746/104067299-30151680-5203-11eb-9aec-59dbbcf0abfb.png)

I found this behaviour in the version 0.83.9 in all x86/x64 VSBuilds (both SDL1 & SDL2 )

Also I found that mingw builds have an additional bug that adds an extra CR/LF plus the mentioned one:

![DOXBOX-X BUG002](https://user-images.githubusercontent.com/67012746/104068043-88004d00-5204-11eb-9c48-b8ad063af249.png)

Thanks in advance for your patience and for your hard work!

Wengier:
@Dalekamistoso Thanks for reporting this! Yes, there is a bug there which I just fixed. Please check out the updated Windows binary below:

* [dosbox-x-update.zip](http://cndos.fam.cx/dosware/dosbox-x-update.zip)

Thanks for your support!

Dalekamistoso:
Solved by Wengier, thank you very much!

PR #2185, opened 2021-01-12
Translation files for DOSBox-X

Dalekamistoso:
These are the files needed for spanish language support.

Also I include original "messages.txt" of a clean installation inside with an appropriate name for default as "en_US.lang" in order to be used as basis for translating DOSBox-X to multiple languages (currently the special characters are not supported in dropdown menus but inside DOS are OK).

In order to use it only needs to change the appropiate value in dosbox.conf (language= .\translations\es\es_ES.lng in windows and ./translations/es/es_ES.lng in Linux)

This enhancement is the result of the fixed issues on #2174 and #2173


rderooy:
Thanks. Just want to mention that English files should be under 'en' and not 'us'.
e.g. en/en_US.lng

That way there can be an en/en_UK.lng file with colour spelled correctly ;-)

Dalekamistoso:
Yes Richard, you are right.

I was in a hurry and I totally forgot to rename before uploading the commit, thanks for noticing it.

I already fixed it :)

> Thanks. Just want to mention that English files should be under 'en' and not 'us'.
> e.g. en/en_US.lng
> 
> That way there can be an en/en_UK.lng file with colour spelled correctly ;-)

Wengier:
@Dalekamistoso This is definitely helpful.

Why you closed it?

Dalekamistoso:
One of my atomated scripts closed it, thanks for your notice!!

I will reopen now

PR #2192, opened 2021-01-16
SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/4387 (Skipped due to conflicts)
https://sourceforge.net/p/dosbox/code-0/4388 (Skipped, see comments below)
https://sourceforge.net/p/dosbox/code-0/4389 (Skipped due to conflicts. DOSBox-X already has a fix for the MayDay demo)
https://sourceforge.net/p/dosbox/code-0/4390 (Skipped due to conflicts)
https://sourceforge.net/p/dosbox/code-0/4391 (Skipped, see comments below)
https://sourceforge.net/p/dosbox/code-0/4392
https://sourceforge.net/p/dosbox/code-0/4393
https://sourceforge.net/p/dosbox/code-0/4394 (Undone by 4395)
https://sourceforge.net/p/dosbox/code-0/4395

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -52,6 +52,16 @@
   - Improved OPL3Duo support, such as adding a buffer
     thread to get rid of slowdowns & breakups in audio
     playback when using the board. (DhrBaksteen)
+  - Integrated SVN commits (Allofich)
+    - r4392: Improve capturing of Zeliard and fix a
+    few percussion-related capturing problems as well.
+    - r4393: Remove video page check when drawing
+    internal mouse pointer, as the BIOS current page
+    is not updated in some cases. Fixes QQP games: Lost
+    Admiral, Conquered Kingdoms, and Solitaire's
+    Journey. Also be a bit smarter about enabling
+    autolock, supporting more cases where the mouse
+    is used for input without resetting it.
 
 0.83.9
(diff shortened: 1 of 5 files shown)


joncampbell123:
Regarding SVN r4388, DOSBox-X defines LINUX=1 when compiling for Linux.

If the #if statement is about Linux adding "8" to scancodes (really?) then it needs to be an #ifdef LINUX, not #if !this && !that && !other

joncampbell123:
Also regarding SVN r4391, masking the Gravis Ultrasound peek/poke address to 20 bits on write is in dispute, because we don't yet know whether the card holds all 24 bits and masks on memory access, or holds only the low 20 bits.

Unfortunately that register appears to be write-only, so I can't just do a read/write test to determine that. Playing audio to test it at the hardware level only confirms that the low 20 bits matter at playback.

Masking at write may also make it more difficult to implement future Gravis Ultrasound PnP (AMD InterWave) emulation which supports more than 20 bits of addressing.

Allofich:
OK, I will remove those commits from the PR.

Allofich:
Shoot, I accidentally added an unneeded commit. Hold on.

Allofich:
OK, I removed those two commits, adding notes about why they were skipped to the skipped commits file.

Issue #2197, opened 2021-01-17
Monochrome/VGA Multi-Screen Emulation


oitofelix:
Are there plans to implement [this](https://www.vogons.org/viewtopic.php?f=32&t=26110)?  It seems there were working patches for DOSBox [here](https://www.vogons.org/viewtopic.php?p=269328#p269328).

joncampbell123:
There are. At least my idea is to use SDL2 multi-window capabilities to do so.

Wengier:
I have already implemented this by adapting the patch. It should work on both SDL1 and SDL2 builds, but later on SDL2 multi-window capabilities may be added for further enhancement of the feature on SDL2 build. Right now it is enabled on debug builds only since it requires the pdcurse library.

Wengier:
This is already supported in DOSBox-X version 0.83.10. Further information may be found in Issue #1075.

PR #2200, opened 2021-01-18
Support for standard/monochrome dual-screen mode

Wengier:
Issue #2197 has requested monochrome/VGA Multi-Screen Emulation. So this ports the patch to implement this feature. The feature is confirmed to work on both SDL1 and SDL2 builds, but later on SDL2 multi-window capabilities may be added for further enhancement of the feature on SDL2 build. For now it is enabled on debug builds only since it requires pdcurses. Tested on Windows, Linux and macOS and seems to work.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -46,6 +46,8 @@
     "windowheight" so that commmands like "CONFIG -GET
     screenwidth" and "CONFIG -GET screenheight" will
     get the current screen width and height. (Wengier)
+  - BOOT command without a parameter will now try to
+    boot the current drive if possible. (Wengier)
   - You can now change properties such as "fullscreen",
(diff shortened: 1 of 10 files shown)


joncampbell123:
@Wengier What happens then if you bring up the debugger interface? Does it switch between secondary display and debugger interface?

Wengier:
@joncampbell123 Since there is only one console window, the logging console, the debugger, and the secondary display could never be enabled at the same time. Even before this patch if you enabled the debugger then the logging console will be disabled. Similarly, if you enable the secondary display then the logging console and debugger will be disabled (the debugger will not show up in such case, and the menu item for debugger is automatically disabled as well if this is enabled).

PR #2211, opened 2021-01-23
SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/4286 - In this PR
https://sourceforge.net/p/dosbox/code-0/4287 - In this PR
https://sourceforge.net/p/dosbox/code-0/4288 - In this PR
https://sourceforge.net/p/dosbox/code-0/4289 - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4290 - In this PR
https://sourceforge.net/p/dosbox/code-0/4291 - Skipped since DOSBox-X is probably phasing out `Bitu` and `Bits` anyway.
https://sourceforge.net/p/dosbox/code-0/4292 - Same as above
https://sourceforge.net/p/dosbox/code-0/4293 - DOSBox-X uses `(void)type;` for this.
https://sourceforge.net/p/dosbox/code-0/4294 - Same as above
https://sourceforge.net/p/dosbox/code-0/4295 - Already in DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/4296 - Mostly already in or irrelevant to DOSBox-X. Just `support.cpp` changes added, except for the part about leaving newline handling to the exception catcher, since as far as I could see DOSBox-X doesn't call `throw` there like SVN does.
https://sourceforge.net/p/dosbox/code-0/4297 - Already in DOSBox-X. Updated comment that said SVN doesn't have this fix.
https://sourceforge.net/p/dosbox/code-0/4298 - Timing change implemented in this PR. `vga.draw.cursor.count++` relocation part is already in DOSBox-X.
https://sourceforge.net/p/dosbox/code-0/4299 - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4300 - Irrelevant to DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4301 - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4302 - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4303 - Skipped because DOSBox-X has its own solution (the "dpi aware" .conf setting)
https://sourceforge.net/p/dosbox/code-0/4304 - Already in DOSBox-X (for 7 years!)
https://sourceforge.net/p/dosbox/code-0/4305 - Already in DOSBox-X (although log message is commented out in DOSBox-X)
https://sourceforge.net/p/dosbox/code-0/4306 - Skipped because of code being significantly different.

To summarize, mostly just skipped a bunch of commits that were either already implemented in DOSBox-X or irrelevant to it. Changes are updates to documentation, and a timing change in https://sourceforge.net/p/dosbox/code-0/4298. @joncampbell123, you may want to look at that timing change.

Also I updated the skipped commits file. I removed the "Website" entry I made for one commit earlier. @Wengier, I think you wanted me to put in the reason for _all_ skipped commits, which is why I added that "Website" entry, but I'd rather only include commits in this file that might possibly be relevant to DOSBox-X in the future. Lots of commits have already been separately implemented in DOSBox-X or are irrelevant, such as the website stuff, so I don't want to bother with those ones.

### DIFF
--- a/README.source-code-description
+++ b/README.source-code-description
@@ -170,42 +170,42 @@ SDL 1.2.x or SDL 2.0.x
     been modified from the original. You can find the changes in the
     source package of DOSBox-X (src/platform/sdl-win32.diff). If you
     want the patched source tree send us an email. (see README)
-    Licensed under LGPL
     Note that only version 1.2.x (SDL1 version) and version 2.0.x
     (SDL2 version) are currently supported.
+    License: LGPLv2+
 
 Curses (optional)
     If you want to enable the debugger you need a curses library.
     ncurses should be installed on just about every unix distro.
-    For win32 get pdcurses at https://pdcurses.sourceforge.net
-    License: Open source
+    For win32 get pdcurses at https://pdcurses.org/
+    License: Public Domain
 
 Libpng (optional)
     Needed for the screenshots.
     For win32 get libpng from https://gnuwin32.sourceforge.net/packages.html
     See http://www.libpng.org/pub/png/ for more details.
-    License: Open Source
+    License: zlib/libpng
 
 Zlib (optional)
     Needed by libpng.
     For win32 get libz (rename to zlib) from https://gnuwin32.sourceforge.net/packages.html
-    See http://www.zlib.net for more details.
-    License: Open Source
+    See https://www.zlib.net/ for more details.
+    License: zlib
 
 SDL_Net (optional)
-    For modem/ipx support. Get it from https://www.libsdl.org/projects/SDL_net/
-    Licensed under LGPL
+    For modem/ipx support.
+    Get it from https://www.libsdl.org/projects/SDL_net/release-1.2.html
+    License: LGPLv2+
 
-SDL_Sound 
-    For compressed audio on diskimages. (optional)
+SDL_Sound (optional)
+    For compressed audio on diskimages (cue sheets) support.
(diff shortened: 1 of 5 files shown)


Wengier:
@Allofich Thanks for these. I agree that the website stuff appears to be completely irrelevant, and there is no need to bother with those ones.

joncampbell123:
Regarding commit 4298, the new VGA blink rate is too slow. Real hardware blinks as it does in DOSBox-X now. See this YouTube video to see what I mean: https://youtu.be/03o2oa9j75I?t=347. I've already gone over this with DOSBox SVN but apparently to no avail.

Allofich:
I've removed commit 4298 and updated the skipped commits file. I agree with you that the cursor blink speed with that change is clearly slower than in that video footage.

> I've already gone over this with DOSBox SVN but apparently to no avail.
I don't know when you talked with them but the commit is from 2019-12-10. Sometimes things get reversed later but I just checked a build of the latest SVN code and it looks like they are still using the slower blink speed.

joncampbell123:
@Allofich When I saw that commit appear in DOSBox SVN I linked to a similar YouTube video pointing out that the VGA blink rate was already correct, and that the new blink rate was too slow. Someone on their team does not seem to have real DOS hardware for reference. SVN seems to have a problem making changes that actually *reduce* emulation accuracy, including that one commit some years ago to GUS emulation that disables the GUS sample interpolation if the channel is playing at > 1x speed.

PR #2213, opened 2021-01-24
Cleanup RPM spec file

rderooy:
I am working to submit dosbox-x to Fedora for possible inclusion,
but for that I need to ensure I had a suitable spec file. The
one included in the source tree had to be cleaned up.

This is still not 100% the same as the spec file I will use to
submit dosbox-x for a packaging review, but it is close.

One difference is, that this spec file generates a SDL1 binary,
while I will be submitting a SDL2 version upstream for 2 reasons
- SDL2 should be the future
- Fedora is unlikely to accept the SDL1 version with customized SDL1

I can change the in-tree spec file to also generate the SDL2 version,
if there is no opposition.

### DIFF
--- a/contrib/linux/dosbox-x.spec.in
+++ b/contrib/linux/dosbox-x.spec.in
@@ -1,33 +1,38 @@
-BuildRequires: libX11 libX11-devel libXext libXext-devel libpng libpng-devel
-BuildRequires: alsa-lib alsa-lib-devel ncurses ncurses-devel zlib zlib-devel
-BuildRequires: mesa-libGL mesa-libGL-devel pulseaudio-libs pulseaudio-libs-devel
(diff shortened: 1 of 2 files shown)


rderooy:
FYI. You can see a Koji build I did for Fedora here: https://koji.fedoraproject.org/koji/taskinfo?taskID=60322491

joncampbell123:
That's fine. It also means if I boot into CentOS to make the RPM releases they too will use SDL2 instead of SDL1.

rderooy:
If we get Fedora included, we should also be able to generate packages for CentOS I think (EPEL)

FYI, here is a newer build with the missing headers added. https://koji.fedoraproject.org/koji/taskinfo?taskID=60355367

rderooy:
I have submitted a packaging review request: https://bugzilla.redhat.com/show_bug.cgi?id=1919639

PR #2216, opened 2021-01-25
Allow TTF cursor blinking rate customization and update comments

Wengier:
The current cursor blink rate is fixed according to the platform (PC-98 or elsewhere). This will make it possible to customize the cursor blink rate, but the current value will be kept as the default value. Also updated some other comments in the config file.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,12 +1,19 @@
 0.83.10
+  - Cursor blinking rate for TrueType font (TTF) output
+    can now be customized with the ttf.blinkc option.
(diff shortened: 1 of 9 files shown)


joncampbell123:
I see something in the code that uses pow(2.0, ...). Can that be replaced with an integer arithmetic shift i.e. 1 << bits? If all you're doing is computing 2^x you can just do 1<<x instead of asking the CPU and C runtime math library to do all that work.

Wengier:
@joncampbell123 Sure! This is definitely a very useful hint, and I have updated the code as you suggested. Thanks!

PR #2219, opened 2021-01-26
Add option to specify "auto" for "use dynamic core with paging on"

koolkdev:
Which turns on this option when 386 paging is used. Change default to "auto".
As discussed here:
https://github.com/joncampbell123/dosbox-x/pull/2182#issuecomment-767393316
The risk that I can think of is if a code would be generated without the ability to handle exceptions when paging is off, but will run from the cache with exceptions enabled after enabling paging. Should I reset the dynamic cache on change? I don't think that it should happen since "MOV CR" isn't implemented in the dynamic core, and any code that ran before enabling paging shouldn't run after it since the addresses are different, no?


Wengier:
@koolkdev Thanks! I think the auto setting will hopefully work in standard usages. Strangely enough though, when I run the timedemo, the logs shows that the function PAGING_Enable() is constantly called with value true, false, true, false ...., so the end score will essentially be the same as "use dynamic core with paging on=true". On the other hand, the setting "use dynamic core with paging on=false" will result in a higher score for this case.

PR #2253, opened 2021-02-07
SVN r4330

Allofich:
https://sourceforge.net/p/dosbox/code-0/4330

This is a somewhat big commit that required adjustment to match with DOSBox-X in several areas. I tried to be careful but since there could easily be problems with it I'm only including this one commit. It does build successfully.

### DIFF
--- a/include/mem.h
+++ b/include/mem.h
@@ -63,14 +63,36 @@ MemHandle                   MEM_NextHandleAt(MemHandle handle,Bitu where);
     Working on big or little endian machines 
 */
 
-#if !defined(C_UNALIGNED_MEMORY)
-/* meaning: we're probably being compiled for a processor that doesn't like unaligned WORD access,
-            on such processors typecasting memory as uint16_t and higher can cause a fault if the
-        address is not aligned to that datatype when we read/write through it. */
-
 static INLINE uint8_t host_readb(ConstHostPt const off) {
     return *off;
 }
+
+static INLINE void host_writeb(HostPt const off,const uint8_t val) {
+    *off = val;
+}
+
+// use __builtin_bswap* for gcc >= 4.3
+#if defined(WORDS_BIGENDIAN) && defined(__GNUC__) && \
+    (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3))
+
+static INLINE uint16_t host_readw(ConstHostPt off) {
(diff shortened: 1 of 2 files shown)


joncampbell123:
This commit modifies the FAT driver. Can you confirm that you can mount both floppy and hard disk images and that this change does not break any file operations?

Wengier:
@Allofich There have been major improvements in DOSBox-X's FAT driver compared with that of vanilla DOSBox, such as FAT32 and LFN support. So I agree this particular SVN commit needs to be tested to make sure it works.

joncampbell123:
One bug that SVN's FAT driver had (that I don't want to creep back into DOSBox-X) was a case where if a file was written up to a multiple of the cluster size, the FAT driver would preemptively allocate a cluster even if not needed. This did not corrupt the file any but if you ran SCANDISK on it, it would pick up on it as a lost cluster and helpfully extend the size of the file to include whatever contents lie in the extra cluster (adding corruption to the end of the file).

Allofich:
I tested the following and they all seemed to work fine:
- Mounting a CD image
- Using "dir" to browse CD image
- Mounting a floppy image
- Running an installation program from the floppy image. It successfully installed to the hard drive.
- Using "xcopy" to copy a file from the hard drive to the same folder. A copy was successfully made (files were identical)
- Mounting a hard drive image
- Copying a file from hard drive image to hard drive
- Overwriting a file from hard drive image to hard drive
- Deleting a file from hard drive
- Copying multiple files from hard drive image to hard drive
- Copying a folder with subfolders from hard drive image to hard drive
- Booting from a hard drive image

Allofich:
I have just noticed something regarding hard disk images and file operations that is unrelated to this PR. Doing the following:

1) mount c (real hard drive folder)
2) imgmount d "(hard disk image).img" -size 512,63,16,507 -fs fat
3) d:
4) xcopy (folder on d) c:/(same folder name) /s
5) Close DOSBox-X

the .img file has its "date modified" changed. I'm not sure if its the time that DOSBox-X was closed or the time of the xcopy operation or something else. Doing a binary compare of the file before and after, it is identical, so only the date modified seems to have changed. But is this intended? It doesn't seem like it should change since the image file was not modified.

I also tried just mounting the .img file and just mounting it and switching to it and using "dir", but this didn't cause the date to change, so it seems to be related to the xcopy command.

Issue #2236, opened 2021-02-01
0.83.10 broken on 32bit arm


rderooy:
**Describe the bug**
When building the new dosbox-x for 32bit arm, I found that the build is broken. 0.83.9 builds fine.

error: #error DYN_X86 core not supported for this CPU target. And it errors out at that point.

The build_log is located here:
https://koji.fedoraproject.org/koji/taskinfo?taskID=61049688

Wengier:
For ARM targets, ```C_DYNAMIC_X86``` should be undefined in config.h. Can you check if this is the case? I don't think there was any actual change related to this in the file configure.ac between versions 0.83.9 and 0.83.10 though. In any case, you can try add the option --disable-dynamic-x86 to the configure script to force disable the dynamic_x86 core for the target(s).

rderooy:
@Wengier part of the issue may be that with the 0.83.10 build, I switched from simply running ``./configure`` to using the %configure script insistent upon. With 0.83.9 the %configure macro did not work at all due to https://github.com/joncampbell123/dosbox-x/issues/2229

In any case, I cannot really check config.h (I think), since this is a build machine and I cannot open individual files.

Here is an example of a build where 32bit arm did build (0.83.9 using ``./configure``:
https://koji.fedoraproject.org/koji/taskinfo?taskID=60490574

Here is an example of a build where ppc64 and s390x failed:
https://koji.fedoraproject.org/koji/taskinfo?taskID=60307465

Here you can see the flags set by the %configure macro on my local system:
```
$ rpm --eval '%configure'

  
  CFLAGS="${CFLAGS:--O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection}" ; export CFLAGS ; 
[...]

PR #2266, opened 2021-02-13
Fix building DOSBox-X RPMs for 32-bit ARM, PPC and s390x

rderooy:
# Description

As part of trying to get DOSBox-X approved for Fedora packaging, I ran into
some issues with 32-bit ARM, PPC and s390x. This PR fixes those issues
and now all the architectures that Fedora Koji builds on (above + i686,
x86_64 and Aarch64) are able to build successfully.

There is still the issue that DYNREC should apparently also compile for
MIPS32 (see src/cpu/core_dynrec/risc_mipsel32.h), but I'm not sure
how to add that to the CPU detect routine in configure.ac such that
it can be whitelisted for DYNREC. And I have no way of testing if it would
even work.

https://github.com/joncampbell123/dosbox-x/issues/2236


rderooy:
Example of a successful koji build with this patch included.
https://koji.fedoraproject.org/koji/taskinfo?taskID=61873522

PR #2268, opened 2021-02-15
Fix For OPL3Duo no Sound on MacOS

josephillips85:
# Description
-This patch handle the thread stop that is happening on MacOS 
making no sound for OPL3 Duo boards on MAC

-FIx MINGW no sound on OPL3 Duo board unbuffered because the lack of threading.
- Developer:Added support for debug on VSCODE


josephillips85:
@DhrBaksteen  in this way i can solve the issue of no sound on the board. on MacOS

DhrBaksteen:
@josephillips85 I tested it and it still works on Windows. Good catch!

Issue #2270, opened 2021-02-15
Record locking not working in DOSBox-X 0.83.10 under Linux and MacOS


dev4inno:
Hello,

I am using several DOS based multi-user applications developed in MS Cobol 4.5 (same as Microfocus Cobol).
However, until now I must use DOSEMU under Linux and vDOSPlus under Windows to get the record locking working.
And: there is no MacOS solution for me so far.

That's why I like DOSBox-X. If it works, I could use the same DOS Emulator everywhere :-)

Unfortunately although announced in the changelog for release 0.83.10 the record locking does not work in this version.

To demonstrate it, I have created a small test program which creates an indexed file and locks one record (if it is unlocked). The program can be started several times and every other instance should detect the locked record and send a message.

Here is the correct behavior in DOSEMU under Debian Linux:

![DOSEMU_Linux](https://user-images.githubusercontent.com/63194871/107964184-528c1380-6fa9-11eb-94ee-7283ba866690.png)

And here is the correct behavior in vDOSPlus under Windows 10:

![vDosPlus_Windows10](https://user-images.githubusercontent.com/63194871/107964272-63d52000-6fa9-11eb-8cf8-cef418acf80b.png)

But in the second DOSBox-X instance under Debian Linux the record lock is not detected:

![DOSBox-X_Linux](https://user-images.githubusercontent.com/63194871/107964327-72bbd280-6fa9-11eb-97f8-73618c2eb97f.png)

Same error in DOSBox-X under macOS:

![DOSBox-X_mac_OS](https://user-images.githubusercontent.com/63194871/107964355-79e2e080-6fa9-11eb-91d3-7312270ad013.png)

For your own tests and to replicate the error I have attached the MS Cobol 4.5 based DOS test program (lockrec.exe) via this link:

http://www.mediafire.com/file/xy088tm6j5dvwua/lockrec.exe/file

It would be perfect if you could implement the same locking function in DOSBox-X like it is available in DOSEMU and vDOSPlus. 

Kind regards from Northern Germany!

emendelson:
@dev4inno For macOS, you can try this Wine-packaged version of vDos (not vDosPlus); it runs under Big Sur and probably a few earlier versions, but it isn't notarized. You may need to clear extended attributes before it will run, but perhaps not.

http://www.columbia.edu/~em36/vDosMac64.zip

dev4inno:
[...]

PR #2273, opened 2021-02-16
Improving record locking for Linux and macOS

Wengier:
As mentioned in Issue #2270, while record-locking is confirmed to work on Windows, it may not work properly under Linux and macOS. This will address the issue so that it should work on all platforms. Verified with the test program posted there.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,7 @@
 0.83.11
+  - Improved handling for file- and record-locking
+    for non-Windows platforms. Portions of the code
+    are adopted from DOSEmu. (Wengier)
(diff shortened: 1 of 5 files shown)


joncampbell123:
@Wengier True to your interest in PCem-like emulation, the latest commits now allow you to load and execute a VGA BIOS image instead of DOSBox-X's internal INT 10h emulation. S3 Trio and Tseng ET4000 emulation generally works with the ROM images from the PCem collection, though S3 256-color modes have pallette issues and Windows 3.1 is unable to correctly set up 640x480 256-color mode with machine=svga_et4000 with those ROM images. But they do work!

Issue #2293, opened 2021-02-21
Errors when mounting certain floppy/CD images from the Drive menu


DSN9990:
**Describe the bug**
Serious issue when I try to mount some CD images from the Drive menu. However IMGMOUNTing the image from the DOS prompt works just fine.

**To Reproduce**
1. Go to the Drive menu
2. Click on D: and choose any ISO affected by this error
3. See error prompt "Drive D: failed to mount"

**Expected behavior**
Mounting such images from the Drive menu should work the same way as IMGMOUNT without errors.

**Screenshots**
From the Drive menu:

![mounterror](https://user-images.githubusercontent.com/69558812/108613082-4222d100-73cd-11eb-8fe8-dd4700826af8.PNG)

From the DOSBox-X prompt:

![command_000](https://user-images.githubusercontent.com/69558812/108613094-641c5380-73cd-11eb-8eec-b95df511cbaf.png)

**Environment (please complete the following information):**
 DOSBox-X 0.83.10

Wengier:
Thanks for reporting this. I think it only happens if the full path is long. It is hopefully fixed in #2294.

Issue #2291, opened 2021-02-20
VGA BIOS is only found in the current working directory


rderooy:
**Describe the bug**
I had a look at the new code for loading a VGA BIOS and quickly noticed an issue. It only finds the ROM if it is in the current working directory.

It should probably look in various places.

What I propose, is something similar to what dosbox-staging has done, and have one function to find additional files (MT32 ROMs, PC98 font, TTFs for screen and printer, BIOS ROMs).

https://github.com/dosbox-staging/dosbox-staging/blob/master/src/misc/cross.cpp
DetermineConfigPath()

This will not be perfect, for instance for a MIDI soundfont, you still will want to also check the system directories as the current code does.

rderooy:
Also look for XDG in:
- https://github.com/dosbox-staging/dosbox-staging/blob/master/src/src/midi/midi_fluidsynth.cpp
- https://github.com/dosbox-staging/dosbox-staging/blob/master/src/src/midi/midi_mt32.cpp

Wengier:
Thanks for the report. With #2294 it should now allow VGA ROMs to be found in DOSBox-X program directory and user config directory too. And the path will be resolved as with other config options that involve paths. As for XDG, I think it is already handled for the user config directory.

rderooy:
The XDG reference was partly meant to quickly locate the relevant piece of code. But also XDG is important on Linux, especially for things like Flatpak.

PR #2294, opened 2021-02-21
Improve VGA ROM loading

Wengier:
Recent code adds support for loading VGA ROM images, which is a great improvement. On the other hand, as mentioned by @rderooy in #2291 it currently only finds the VGA ROM if it is in the current working directory. This PR allows the ROM to be found in the DOSBox-X program directory and the user config directory too. And the path will be resolved (~ or environment variables) as with other path options in the config file.

Also fixed Issue #2293 for mounting images from the menu.


joncampbell123:
Will this patch still check the current directory as well? If the dosbox.conf specifies a full path for the VGA BIOS, will it only use that full path?

Wengier:
@joncampbell123 Sure, it will check the current directory first, then the DOSBox-X program directory, and finally the user config directory. If a VGA ROM image is found and read at any stage then it won't go further. The path will be resolved which means that tilde (Linux/macOS) / environment variables (Windows) will be expanded before searching.

PR #2295, opened 2021-02-21
SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/4337 - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4338 - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4339 - Skipped. Web site-related.
https://sourceforge.net/p/dosbox/code-0/4340 - In this PR.
https://sourceforge.net/p/dosbox/code-0/4341 - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4342 - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4343 - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4344 - In this PR
https://sourceforge.net/p/dosbox/code-0/4345 - Skipped. Not needed.
https://sourceforge.net/p/dosbox/code-0/4346 - In this PR

sBitfs(x) changes in the implemented commits were skipped since the commit implementing that function was already skipped, since DOSBox-X seems to be moving away from using Bitu and Bits. Also the LOG calls sBitfs(x) is being used with in SVN already have their parameters cast to the expected sizes in DOSBox-X.

### DIFF
--- a/src/hardware/memory.cpp
+++ b/src/hardware/memory.cpp
@@ -1825,7 +1825,7 @@ void Init_RAM() {
 
     /* Allocate the RAM. We alloc as a large unsigned char array. new[] does not initialize the array,
      * so we then must zero the buffer. */
-    MemBase = new uint8_t[memory.pages*4096];
+    MemBase = new(std::nothrow) uint8_t[memory.pages*4096];
     if (!MemBase) E_Exit("Can't allocate main memory of %d KB",(int)memsizekb);
     /* Clear the memory, as new doesn't always give zeroed memory
      * (Visual C debug mode). We want zeroed memory though. */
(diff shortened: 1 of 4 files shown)


src/hardware/memory.cpp
```diff
@@ -1825,7 +1825,7 @@ void Init_RAM() {
 
     /* Allocate the RAM. We alloc as a large unsigned char array. new[] does not initialize the array,
      * so we then must zero the buffer. */
-    MemBase = new uint8_t[memory.pages*4096];
+    MemBase = new(std::nothrow) uint8_t[memory.pages*4096];
```
joncampbell123, line 1828: Good catch!

(std::nothrow) makes sure that if the new operator cannot allocate memory, it returns NULL instead of throwing a std::bad_alloc exception.

PR #1863, opened 2020-09-17
WIP: Add support for networking using libslirp

Jookia:
# Description

This implements a networking backend for the NE2000 using libslirp on Unix systems. This mean DOSBox-X can provide networking to users without any host system configuration, only guest configuration.

To test this: compile this code (it will error if libslirp isn't found), enable ne2000 and load the ne2000 driver: 'ne2000.com 0x60 3 0x300' and browse the net using Lynx

**Does this PR address some issue(s) ?**

It's a pain to network DOSBox. I personally couldn't even get pcap working on my Linux system, the best I've done is using SLIP over serial. This is discussed a bit more in #320.

**Does this PR introduce new feature(s) ?**

Yes, it implements connectivity to the Internet using libslirp.

**Are there any breaking changes ?**

See below.

**Additional information**

This patch works, but there are multiple issues I need input on:
- How should this coexist with pcap support? Should it be a runtime or compile-time selection?
- Should I dynamically allocate memory for resources like pollfds and slirp timers or just error when out of resources?
- Should I use global variables like I do now?
- Should users be able to change their IPs and DHCP hostname?

Other stuff I need to do:
- slirp_input runs without checking slirp is set
- I need to fix up registering pollfds
- I have some logging code enabled I forgot to remove
- I have some code that should make pcap coexist with slirp in the code but I haven't tested it
- Check for slirp at configure-time

PR #2327, opened 2021-03-03
Implement support for slirp-based connections

Jookia:
# Description

This is no longer a WIP but a finished version of #1863 . Sorry for the duplicate PR.

See #1863

This adds support for slirp-based Internet connectivity.

None to my knowledge.

**Additional information**

I've tested this on Linux and Windows.


Jookia:
Force pushed to fix some whitespace issues.

Jookia:
Force pushed to change 'PCAP' to 'pcap' and 'SLIRP' to 'Slirp' in user-visible messages.

Wengier:
@Jookia Thanks for the efforts. I will test it further, but for now I would suggest to rename the config option "pcaptimeout" in [ethernet, pcap] section to "readtimeout" or so. It was previously named "pcaptimeout" in [ne2000] section only to make it clear that it was for pcap, but it is now under [ethernet, pcap] section any way so I think a name like "readtimeout" would be better.

Jookia:
Could you provide some example code on how to do that? I haven't seen
any other config migrations handle things that way.

Wengier:
@Jookia You are right that most existing config migrations do not involve renaming. But there is already an exception in the code which can serve as an example. You may take a look at the file sdlmain.cpp between line 11826 and 11857, which transfers the config option "dos in hma" in [dos] section to "dos=high" or "dos=low" option in [config] section (which emulates config.sys). Not only it renames the config option but actually completely change its form. Please check it out.

joncampbell123:
Let me know when you're ready for me to merge this.

Jookia:
I'll check out the migration code you referenced.

With that fixed, would this be good for merging?
Though you still need to test.

Jookia:
Okay, done that.

Jookia:
Force pushed to clean up ethernet_slirp error messages

Jookia:
Force pushed to add 'ETHERNET:' namespace in ethernet.cpp error output.

Most of these are just quick fixes I'm picking up as I'm working on
porting a condensed version of this patch set to dosbox-staging.

Wengier:
@Jookia More test is desired of course, but since DOSBox-X 0.83.11 was recently released, I think there should be time to clean up for the rest before the next version.

Jookia:
Is there anything you actually want me to do now? Is this getting merged?

Wengier:
@Jookia I think the main purpose of this PR is to get the feature work, assuming we already have the dependency, since the feature has already been worked on for while and also tested by multiple people. But it will be further cleaned up of course before the next release, e.g. I will see the dependency problem can be improved especially for Windows. Getting merged simply means that the feature is already in a production state, but does not mean that it cannot be further improved to make it work better.

Jookia:
I'm confused. Do you want this or not?

joncampbell123:
@Jookia I'll go ahead and merge this if it's good enough to start testing.

Jookia:
It's production ready, all the code is there and I've tested it fairly well.

Wengier:
I noticed that PKG_CHECK_MODULES is not be supported in some environments and will cause the configure script to fail in such cases, so I have been trying to replace it with AC_CHECK_HEADER and AC_CHECK_LIB as with other libs in #2347.

Jookia:
The change there looks good to me at a first glance.

Wengier:
DOSBox-X 0.83.12 that supports the slirp-based connections has been released. It is not yet enabled for the Visual Studio builds, but I hope more work will be done so that these builds will support it as well.

rderooy:
The new flatpak build is also out and has slirp support
https://flathub.org/apps/details/com.dosbox_x.DOSBox-X

PR #2339, opened 2021-03-06
Add Linux MAN page

rderooy:
# Description

This restores the Linux man page. With major cleanups.

This is part of an effort to also add bash auto-completion in
the (hopefully) near future.


rderooy:
Also includes the commit to fix the build issue on the s390x by @Wengier

Issue #2332, opened 2021-03-05
Mac OS X users need a way to turn on all debugging output, and show a terminal window of some kind for console output


joncampbell123:
**Is your feature request related to a problem? Please describe.**
Linux users can just run from the command line for console output, and Windows users can bring up a console window on demand. Mac OS X users, if they need the console output, are stuck with log.txt because there is no Terminal window DOSBox-X can just bring up on a whim.

**Describe the solution you'd like**
- A menu item to set all log output options to "debug"
- A "show console" window (just like Windows builds) that creates another NSWindow with console output the user can read log output with.

See also: https://discord.com/channels/231947749862539264/319281334054944778/817223246779908096

Jookia:
Would it be possible to log to an in-memory buffer that can be dumped later using a command?

joncampbell123:
Funny you mention that, the log is already held in memory, or at least the last 500 lines anyway, so that the debugger interface can let you scroll through it. Perhaps a "save log as..." menu command, which could then use the file browser and interface additions @Wengier added some time back?

Jookia:
Holding more lines and having a 'save log as' would be really helpful in
general since it would mean platforms wouldn't necessarily need the
command line at all. You could even call out to the OS to open the new
file in an editor.

Wengier:
Sure, I agree that a 'Save log as..' would be very helpful in general (independent of the OS) so that you don't need to enable the log console nor the log file in the configuration. I have added this under "Debugging options" in #2340.

josephillips85:
as a suggestion why not push the command or option key while the app is opening to pop up the debug window?

joncampbell123:
The debug window is implemented using ncurses, for the terminal. To actually get the debug interface, you would have to run DOSBox-X manually from the Terminal application. DOSBox-X does not have a console to implement the debug interface if it is started from the Finder.

Unlike Windows, Mac OS X does not have an API to spawn and manage a process console window.

Linux has the same problem, if you run it from the graphical desktop without instructing it to open a terminal window.

Therefore, the need to implement a separate NSWindow that displays the console output as text.

PR #2340, opened 2021-03-06
Add "save log as..." menu option

Wengier:
This adds a "Save log as..." menu option under "Debugging options" as requested in #2332.

### DIFF
--- a/src/debug/debug_gui.cpp
+++ b/src/debug/debug_gui.cpp
@@ -22,6 +22,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <errno.h>
+#include <fstream>
 
 #if defined(WIN32)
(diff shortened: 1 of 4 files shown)


src/debug/debug_gui.cpp
```diff
@@ -106,6 +107,14 @@ static list<string>::iterator logBuffPos = logBuff.end();
 
 extern int old_cursor_state;
 
+void savetologfile(const char *name) {
+    std::ofstream out(name);
```
joncampbell123, line 111: This is OK for now but in a future pull request please add a call to out.is_open() to confirm the ofstream instance was able to open the file without error.

https://www.cplusplus.com/reference/fstream/ofstream/is_open/

PR #2369, opened 2021-03-14
Update MMX code and some fixes

Wengier:
This updates the dynamic MMX support to the latest code by @kekko-dos, which according to @kekko-dos is more stable than the current implementation in DOSBox-X. Tested with MMXTEST.

Also fixed some smaller issues.


koolkdev:
I would recommend to replace `gen_call_function` (of gen_mmx_op) with `dyn_call_function_pagefault_check`, because there may be a pagefault in gen_mmx_op that will cause a crash if non recursive exception happens..

Wengier:
@koolkdev Thanks for the suggestion! I will replace gen_call_function with dyn_call_function_pagefault_check as you suggested.

PR #2370, opened 2021-03-14
SVN r4421

Allofich:
https://sourceforge.net/p/dosbox/code-0/4421

I tested running a game with the dynamic core, and it seemed to run without problems.

### DIFF
--- a/src/cpu/core_dyn_x86/decoder.h
+++ b/src/cpu/core_dyn_x86/decoder.h
@@ -2371,14 +2371,40 @@ static void dyn_interrupt(Bitu num) {
 	dyn_closeblock();
 }
 
-static void dyn_add_iocheck(Bitu access_size) {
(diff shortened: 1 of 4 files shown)


Allofich:
@joncampbell123, @Wengier, I'm currently working on
https://sourceforge.net/p/dosbox/code-0/4424
which will probably be mergeable (haven't come across any big conflicts so far).

It's a big commit so it's taking me quite a while to work through it. If you can take a quick look at URL above, and if you notice that it's something we don't want in DOSBox-X for whatever reason, it will save me from having to spend any more time on it.

joncampbell123:
Hm... I appreciate changing pointers to const when the data is not being modified, but... isn't it modifying the data in some cases anyway (typecasting back to non-const)? I know GCC and static analyzers don't like that and will warn over that.

There's also the issue brought up on Discord where dynamic core causes Touhou Project 1 boundary checks to fail, apparently x87 related (yes, TH01 uses floating point for game object position!)

PR #2403, opened 2021-03-31
SVN r4444

Allofich:
https://sourceforge.net/p/dosbox/code-0/4444/

Modified as needed to make it work with DOSBox-X.

The original SVN commit changes
`#define BIOS_DEFAULT_RESET_LOCATION		(RealMake(0xf000,0xe05b))`
to
`#define BIOS_DEFAULT_RESET_LOCATION		(RealMake(0xf000,(machine==MCH_PCJR)?0x0043:0xe05b))`

but DOSBox-X instead uses

`Bitu BIOS_DEFAULT_RESET_LOCATION = ~0u;`
and
`BIOS_DEFAULT_RESET_LOCATION = PhysToReal416(ROMBIOS_GetMemory(64/*several callbacks*/,"BIOS default reset location",/*align*/4));`

so this part of the SVN commit is omitted.

The original commit claims to fix the CGA versions of Robomaze III and Road Runner, and I can confirm that in SVN, before this commit these two programs in CGA mode would not work correctly, with the screen stuck showing the DOS prompt, but at r4444  Road Runner CGA mode works and Robomaze III stops showing the DOS prompt (although for me it has a black screen).

In DOSBox-X, though, Road Runner already works in CGA mode before this commit, and Robomaze III also already behaves like current SVN (gets out of the DOS prompt but to a black screen). These two games seem to work the same also after this commit. This means that I'm not aware of this PR fixing anything, it just brings in as much of the code updates as I could from r4444.

This PR could cause problems with PC-98 mode, since SVN doesn't have PC-98 and so it's not accounted for in the changes.

### DIFF
--- a/src/dos/dev_con.h
+++ b/src/dos/dev_con.h
@@ -130,7 +130,7 @@ private:
         col=CURSOR_POS_COL(page) ;
         row=CURSOR_POS_ROW(page) ;
         if (!IS_PC98_ARCH)
-            ansi.nrows = real_readb(BIOSMEM_SEG,BIOSMEM_NB_ROWS) + 1;
+            ansi.nrows = IS_EGAVGA_ARCH ? (real_readb(BIOSMEM_SEG,BIOSMEM_NB_ROWS) + 1) : 25;
         tempdata = (ansi.data[0]? ansi.data[0] : 1);
         if(tempdata + static_cast<Bitu>(row) >= ansi.nrows)
         { row = ansi.nrows - 1;}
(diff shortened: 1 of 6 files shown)


Allofich:
At dosbox-staging, SVN r4444 supposedly broke the `cls` and `dir` commands for CGA, Hercules, and PCjr.
https://github.com/dosbox-staging/dosbox-staging/commit/9d0073a2a72812cad3af014d412a918245ce67d1

But, testing with `machine=cga`, I found that this PR does not break  `cls`  or `dir` for DOSBox-X.

joncampbell123:
Good call.

If you look at IBM PC/XT BIOS source code, you'll notice the "number of rows" byte in the BIOS data area didn't exist until 1986 or so, so it is right to assume 25 unless EGA/VGA.

I'm pretty certain (but not 100% sure) that the field does not exist under MDA/Hercules/PCjr/Tandy either.

PR #2450, opened 2021-04-16
Add cross-platform support and more options for DOSBox-X working directory

Wengier:
With the recent code DOSBox-X shows a dialog box that asks the user to select a folder if current directory is the root directory and the operating system is macOS. It is definitely useful for macOS, but I think this function can be further extended - both to other platforms and with additional options for working directories (e.g. user can set the primary config file directory or the DOSBox-X program directory as its working directory in the config directly instead of perhaps having to select a folder every time). This PR adds two config options (namely "working directory option" and "working directory default" in [dosbox] section) for this.

For the setting "working directory option", the following options are available (it defaults to "promptifroot" for Linux and macOS):

prompt: DOSBox-X will (always) ask the user to select a working directory when it runs.
promptifroot: DOSBox-X will ask for a working directory if run from root (Linux/macOS).
custom: Specify a working directory via the "working directory default" option.
config: DOSBox-X will use the primary config file directory as the working directory.
program: DOSBox-X will use the DOSBox-X program directory as the working directory.
runtime: The directory DOSBox-X runs becomes the working directory (default for Windows).
force: Similar to "custom", while overriding -defaultdir command-line option if used.

The setting "working directory default" allows the user to specify a working directory directly if "working directory option" is set to "custom" or "force", and it will set the default directory for the folder selection dialog box if "working directory option" is set to "prompt" or "promptifroot".

Tested in Windows, Linux, and macOS platforms.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,11 @@
 0.83.13
+  - Added config options "working directory option" and
+    "working directory default" in [dosbox] section of
+    the config file to provide various options for the
+    users to control DOSBox-X's working directory. For
+    example, DOSBox-X can use the primary config file
+    directory or the DOSBox-X program directory as its
+    working directory. (Wengier)
   - Mac OS X builds will prompt the user to select a
     folder at startup if run from the Finder (or from
     the root directory). The folder selected will then
--- a/contrib/windows/installer/dosbox-x.reference.setup.conf
+++ b/contrib/windows/installer/dosbox-x.reference.setup.conf
@@ -185,16 +185,16 @@ logfile     =
 #                                                    If set to autofile, DOSBox-X will warn if there are open file handles or a guest system is currently running.
 #                                                    Possible values: true, false, 1, 0, auto, autofile.
 #                        working directory option: Select an option for DOSBox-X's working directory when it runs.
-#                                                    prompt: DOSBox-X will ask the user to select a working directory when it runs.
-#                                                    promptifroot: DOSBox-X will ask for a working directory if run from root (Linux/macOS).
-#                                                    custom: Specify a working directory via the "working directory default" option.
 #                                                    config: DOSBox-X will use the primary config file directory as the working directory.
-#                                                    program: DOSBox-X will use the DOSBox-X program directory as the working directory.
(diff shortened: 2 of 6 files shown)


joncampbell123:
@Wengier How can DOSBox-X read dosbox.conf know to prompt for a folder, if it first needs to browse for the folder to read dosbox.conf?

Remember, the whole reason for the Mac OS X code is that the current working directory is undefined or / if run from the Finder, and the Finder doesn't provide any control for the user to say *what* the current working directory is. If run from the Terminal however, the user has direct control over what the current working directory will be, and the prompt is not needed.

Wengier:
@joncampbell123 If you look at the code, you will notice that DOSBox-X not only reads config file (dosbox-x.conf/dosbox.conf) from the current directory, but also from the DOSBox-X program directory (the directory where dosbox-x itself is located), user config directory, as well as user-defined directory (-conf option) if specified. So even if DOSBox-X's run from the Finder with / as the current directory, DOSBox-X will still look for its config file (dosbox-x.conf/dosbox.conf) from directories such as the DOSBox-X program directory and the user config directory. So if the user puts such a config file in any of these directories, then DOSBox-X will read it when it runs even if the current directory is the root. I have confirmed this myself on macOS too (even if I run it from the Finder).

joncampbell123:
@Wengier I like some of the ideas here, but I would prefer to implement them differently.

My other concern is the general library you are using for other dialogs would not allow the additions and settings used for common dialogs needed to help convey to the user what they are selecting a folder for.

I am quite familiar with Windows GetOpenFileName() so I can add that easily in about half an hour.

Regarding Windows, I like the idea of DOSBox-X prompting for a folder if you run it from the desktop icon or Start menu icon put there by the installer, but only if there is no dosbox.conf already in the current working directory so that existing users can continue to use it normally who are accustomed to copying dosbox-x.exe or dosbox.exe into the same folder as their game, and if not run from the command prompt.

I do like the idea of a global dosbox.conf. The user could put their most common settings in there. It would first load the global one, then the local dosbox.conf overrides that.

I'll sit down and work on something equivalent to this, and then do some cleanup.

PR #2452, opened 2021-04-17
No S_ISDIR defined fix

lab313ru:
Here: https://github.com/joncampbell123/dosbox-x/blob/master/src/gui/sdlmain.cpp#L11716


joncampbell123:
I can confirm VS2019 for some reason does not have S_ISDIR. Interesting.

PR #2461, opened 2021-04-19
SVN r4447

Allofich:
https://sourceforge.net/p/dosbox/code-0/4447/

The original commit message is

"Attribute Controller port alias on EGA machine. Fixes EGA display of older Super Pac-Man release."

I can confirm that Super Pac-Man is fixed.

Before this PR (game runs but screen is black):

egaspac_000

After this PR:

egaspac_000

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -92,6 +92,10 @@
     version 2.5.0. (Wengier)
   - Updated FLAC decoder library to the latest version
     (0.12.29 by David Reid). (Wengier)
+  - Integrated SVN commits (Allofich)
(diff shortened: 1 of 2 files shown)


joncampbell123:
Interesting, but makes sense.

This reminds me of how Sound Blaster cards alias odd and even I/O ports as if bit 0 is not part of deciding how to handle the I/O, unless even/odd ports explicitly have different purposes (such as mixer registers). This is why on such cards you can get away with writing DSP commands to either I/O port 22Ch or 22Dh.

PR #2482, opened 2021-04-26
SVN r4424

Allofich:
https://sourceforge.net/p/dosbox/code-0/4424

"Make all references to cache const, use cache_adds instead of direct access."

Modified slightly to work with DOSBox-X. (I had to change `cache_rwtox` to take a const pointer because it is passed pointers that were made const by this commit.)

I'm OK with waiting until after the end-of-month release (that I assume is coming) before merging, if necessary.


joncampbell123:
I'm going to have to reject this one on the principal that you should never declare const pointers and then typecast them back to non-const on a whim. If you're going to dereference a pointer and modify memory, don't declare it a const pointer.

PR #2500, opened 2021-05-02
Added 132.com

tonypdmtr:
# Description

_Added 132.com for switching to 132 wide screen (useful for program listings, such as assembly language)._

_Not in the emulator but only the built-in utilities_


rderooy:
You know you can accomplish the same thing with ``mode con cols=132``?

tonypdmtr:
> You know you can accomplish the same thing with `mode con cols=132`?

OK but 1. Simpler to type, 2. complements the already present utilities (50.com, 28.com, 25.com)

rderooy:
You can just create a batch file named 132.bat and place it on the Z:\ drive that does just that.

Since you can do that, and ``mode con`` now works in recent DOSBox-X versions, I *personally* think that 50/28/25.COM are no longer really needed, and are just clutter on the Z:\ drive.

Wengier:
In the Z: drive there is a zip file named TEXTUTIL.ZIP, which contains the following tools:

```
80x25.COM
80x43.COM
80X50.COM
80X60.COM
132X25.COM
132X43.COM
132X50.COM
132X60.COM
...
```

(You can see the full list by entering the command ```UNZIP -l TEXTUTIL.ZIP``` on the Z drive)

I wonder exactly which function does this 132.COM do, i.e. 132X25.COM, 132X43.COM. 132X50.COM, or 132X60.COM?

Wengier:
@rderooy I agree that 50/28/25.COM are probably no longer really needed either, but for compatibility purpose perhaps they can be made hidden by default so that DIR alone will not see them (but DIR /A will do), and they can still be executed as usual. See the "drive z hide files" config option for more info.

Wengier:
I am sure @tonypdmtr did a great job on this, but I agree that putting all files like this may just be clutter as @rderooy said. This is also a reason that some similar tools are packaged inside a zip file.

joncampbell123:
I'm perfectly fine with 132.COM myself, but it's not that important to me. Also recent pull requests has caused a conflict with src/shell/shell.cpp.

Let me know if anyone really does want 132.COM in drive Z:\

Wengier:
Since latest code added support for directories on the Z drive, I think we can probably just put the tools included inside the TEXTUTIL.ZIP package to a directory named TEXTUTIL, which will include 80x25.COM, 80x43.COM, 80X50.COM, 80X60.COM, 132X25.COM, 132X43.COM, 132X50.COM, 132X60.COM, etc. It will no longer be clutter with this new feature.

Wengier:
See PR #2511 for the addition of such a dedicated directory on the Z drive.

Wengier:
I checked myself - this 132.COM program has the exact same functionality as the 132X43.COM program on the TEXTUTIL directory of Z drive.

Issue #1472, opened 2020-04-09
Microsoft DMF image support


rderooy:
**Is your feature request related to a problem? Please describe.**
In a similar vain to
https://github.com/joncampbell123/dosbox-x/issues/1466
https://github.com/joncampbell123/dosbox-x/issues/1465

It would be good if DOSBox-X gained support for Microsoft's DMF (Distribution Media Format), which was used for software distribution in the mid 90's. For instance it was used for the diskette versions of Windows 95, but also for some versions of Office.

https://en.wikipedia.org/wiki/Distribution_Media_Format
http://www.winimage.com/wimushlp/wini1a1y.htm

**Describe the solution you'd like**
Ability to imgmount/read/write DMF images in DOSBox-X.

**Describe alternatives you've considered**
Some software was distributed also on CD-ROM, for those you don't need DMF image support.

**Additional context**
USB disk drives apparently cannot read/write DMF disks either.

joncampbell123:
It may be possible to support DMF using some of the other disk image formats that were added for PC-98 games, because they can represent arbitrary sectors in a way that flat disk images cannot.

rderooy:
The patch to support DMF seems a bit simplistic according to what I read here: http://www.winimage.com/wimushlp/wini1a1y.htm

There seem to be 2 different cluster sizes (1024 and 2048), and 1680 and 1720 capacities. And there are also changes in the maximum number of files in the root of the filesystem. Perhaps if you limit the support to read-only...

Allofich:
As commented by @rderooy in https://github.com/joncampbell123/dosbox-x/pull/2508, there was a mainline DOSBox commit that appears to have added DMF support. In that PR (a commit to merge in SVN updates) I did not include that because of a difference in code between mainline and DOSBox-X, namely that DOSBox-X had extra parameters.

I recently tried adding the commit to DOSBox-X by filling in the missing parameters, searching what they should be online, but I had no luck; the resultant build still couldn't mount DMF images. I think I will try latest mainline DOSBox mainline code and see if it can mount them.

Allofich:
Yes, SVN DOSBox can mount the image.

Siddhant-Mashalkar:
> Yes, SVN DOSBox can mount the image (a Windows 95 installation disk).

DOSBox  was able to mount image, but DOSBox-X couldn't 
[...]

PR #2508, opened 2021-05-06
SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/4444/ - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4445/ - Code conflict
https://sourceforge.net/p/dosbox/code-0/4446/ - Code conflict
https://sourceforge.net/p/dosbox/code-0/4447/ - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4448/ - Code conflict. DOSBox-X uses additional parameters for defining disk types not defined in the SVN commit. This could be easily added to DOSBox-X if the additional parameter values are known. @joncampbell123 or @Wengier, want to take a look?
https://sourceforge.net/p/dosbox/code-0/4449/ - Skipped. Code duplication here only existed in SVN, not DOSBox-X, so making it into a function is unnecessary.
https://sourceforge.net/p/dosbox/code-0/4450/ - In this PR. I don't know what the regression mentioned is or if it is present in DOSBox-X, so I refrained from putting it in the changelog. Similar code to preserve the error code is also already in another function, `fatDrive::FileCreate`.
https://sourceforge.net/p/dosbox/code-0/4451/ - Partially in this PR. The deleted file is not in DOSBox-X so just the help message changes are in this PR.
https://sourceforge.net/p/dosbox/code-0/4452/ - Skipped. File is not used in DOSBox-X (it's in `unused-source-files.zip` in the `src/libs/zmbv` folder.)

### DIFF
--- a/patch-integration/Skipped SVN commits.txt
+++ b/patch-integration/Skipped SVN commits.txt
@@ -75,3 +75,6 @@ Commit#:	Reason for skipping:
 4418		No C_SRECORD in DOSBox-X
 4424		Declares const but then casts to non-const
 4429		Conflict
+4445		Conflict
+4446		Conflict
+4448		Additional parameters for DOSBox-X needed
--- a/src/dos/drive_fat.cpp
+++ b/src/dos/drive_fat.cpp
@@ -2180,8 +2180,10 @@ bool fatDrive::FileCreate(DOS_File **file, const char *name, uint16_t attributes
 bool fatDrive::FileExists(const char *name) {
     direntry fileEntry = {};
 	uint32_t dummy1, dummy2;
-	if(!getFileDirEntry(name, &fileEntry, &dummy1, &dummy2)) return false;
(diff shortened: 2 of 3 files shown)


rderooy:
Regarding https://sourceforge.net/p/dosbox/code-0/4448/
that seems to add support for the MS DMF (1.68MB) format for which I had added a feature request some time ago
https://github.com/joncampbell123/dosbox-x/issues/1472

rderooy:
The patch to support DMF seems a bit simplistic according to what I read here: http://www.winimage.com/wimushlp/wini1a1y.htm

There seem to be 2 different cluster sizes (1024 and 2048), and 1680 and 1720 capacities. And there are also changes in the maximum number of files in the root of the filesystem. Perhaps if you limit the support to read-only...

PR #2512, opened 2021-05-07
MSVC warning fixes

Allofich:
Fixes/silences all warnings in DOSBox-X files when building the 64-bit SDL1 solution. Warnings still are raised for the various library files used.

By the way, a warning is what allowed me to notice the issue fixed by my previous PR https://github.com/joncampbell123/dosbox-x/pull/2510.

Next I'm looking at fixing other warnings, such as those of GCC and Clang.


Allofich:
The fix for the "not all control paths return a value" warning is a bit of a guess. Based on the existing code I assumed it should return true if things went OK and false if it failed, so in the part with the warning I made it return true if it successfully closed the file and false if it failed to.

PR #2516, opened 2021-05-09
Add menu option to restart DOSBox-X with selected config file

Wengier:
This PR adds the menu option in the "Main" menu to automatically restart DOSBox-X with the selected config file so that DOSBox-X will re-load itself with a new configuration file that may be completely different from the current one, similar to the ```config -bc``` command-line option I added previously. Also added/updated help messages for commands like CGASNOW and VFRCRATE, and a TITLE.COM program in Z:\BIN to change the DOSBox-X window title, along with smaller fixes.

P.S. I noticed that the fix in PR #2513 by @Allofich seems to break the readTTF function for different TTF styles. That PR was intended to reformat the code style, but unfortunately it did not work as expected for this function. So I reverted this to the previous code to make it work again.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -30,6 +30,8 @@
   - Added menu options "Generate NMI interrupt" & "Hook
     INT 2Fh calls" ("Help" => "Debugging options") for
     more debugging options. (Wengier)
+  - Fixed issues related to screen dimensions in TTF
+    output in CGA/EGA modes since 0.83.12. (Wengier)
   - Integrated SVN commits (Allofich)
     - r4443: Improve detection of Paradise SVGA in some
     installers with additional signature.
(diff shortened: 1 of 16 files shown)


Allofich:
> P.S. I noticed that the fix in PR #2513 by @Allofich seems to break the readTTF function for different TTF styles. That PR was intended to reformat the code style, but unfortunately it did not work as expected for this function. So I reverted this to the previous code to make it work again.

Thanks for fixing it. I'm planning to revisit it again later (along with fixing other warnings) and will be sure to test that TTF still works afterwards.

Wengier:
@Allofich I found the reason that your code did not work is that the parameters in the readTTFStyle function needs to be passed by reference instead of by value. I have already fixed this as part of PR #2519.

PR #2520, opened 2021-05-11
SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/4453/ - In this PR
https://sourceforge.net/p/dosbox/code-0/4454/ - In this PR

@joncampbell123 (and maybe @Wengier), I imagine 4454 may be of interest to you, as it deals with the A20 gate that DOSBox-X has various options for. DOSBox-X's `INT15_Handler` function didn't have a case 0x24, which this commit adds, so it seems like a positive addition, but I don't know if it might interfere in some way with existing A20 related code or options in DOSBox-X.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -38,6 +38,9 @@
   - Integrated SVN commits (Allofich)
     - r4443: Improve detection of Paradise SVGA in some
     installers with additional signature.
+    - Improve bittest instructions to wrap more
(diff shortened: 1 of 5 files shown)


Allofich:
I forgot to put the SVN commit number in the CHANGELOG file for r4453. I'll fix that in another PR.

PR #2535, opened 2021-05-18
Remove casts that remove const qualification

Allofich:
Improves const-correctness by removing several casts of const values to non-const values (identified with a static analyzer).

The removed casts were of 2 cases.

Case 1: A const value was being cast to non-const to be used as a parameter in a function call. In some cases the function parameter was const, so I just removed the cast if I could, or if there was a type conversion happening with the cast but the const keyword was missing, I added the const keyword. In other cases the function parameter was non-const, but it could be changed to const without issue, so I did that and removed the cast to non-const.

Case 2: A string/const char pointer was being cast to non-const. In some cases the cast was just unnecessary, so I removed it. In the others, I used `strcpy` to copy the const char pointer contents to a non-const char array, which is then used in the following operations that need to modify it.

### DIFF
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -122,7 +122,7 @@ public:
 	virtual uint8_t Read_AbsoluteSector(uint32_t sectnum, void * data);
 	virtual uint8_t Write_AbsoluteSector(uint32_t sectnum, const void * data);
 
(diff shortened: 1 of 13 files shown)


Wengier:
@Allofich The use of strcpy to copy the const char pointer contents will cause crash if there is a buffer overflow (since the array is fixed-length). The previous version did not crash in such cases. Please check it. Thanks.

PR #2539, opened 2021-05-19
Use "emplace_back" and "emplace_front", and other cleanup (via static analysis)

Allofich:
Mainly this is the replacement of various calls to `push_back` and `push_front` with `emplace_back` and `emplace_front`, which appear as recommendations when using the static analyzer Sonic Lint. Here is their page on this recommendation: https://rules.sonarsource.com/cpp/RSPEC-6003

The page says it only creates the recommendations "when an insertion function on a supported container leads to the construction of a large temporary object that can be avoided by using the provided emplacement member function."

I checked a bare-bones example based on the replacement recommendation that appeared for `debug_gui.cpp` using this site:
https://godbolt.org/z/x5hMq8r9n

and replacing `push_back` with `emplace_back` did shorten the generated assembly (for each of GCC, Clang and MSVC).

I couldn't find it now, but I thought I remembered seeing somewhere that DOSBox-X aims to be buildable on C++11? Not sure if that's correct, but the emplace functions were introduced in C++11, so I think this should be OK in that regard.

https://en.cppreference.com/w/cpp/container/vector/emplace_back
https://en.cppreference.com/w/cpp/container/list/emplace_front

Also a few other minor cleanups like redundant casts, what looks like a comma typo in place of a semicolon, etc. as found with the static analyzer.

### DIFF
--- a/include/control.h
+++ b/include/control.h
@@ -67,7 +67,7 @@ public:
     std::vector<std::string> startup_params;
     std::vector<std::string> configfiles;
     Config(CommandLine * cmd):cmdline(cmd),secure_mode(false) {
-        startup_params.push_back(cmdline->GetFileName());
+        startup_params.emplace_back(cmdline->GetFileName());
         cmdline->FillVector(startup_params);
         opt_exit = false;
         opt_debug = false;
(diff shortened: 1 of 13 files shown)


joncampbell123:
DOSBox-X aims to require a minimum of C++11, yes.

Future work (in a year or two) may bump that to C++14 or C++17, but that's not definite right now.

PR #2546, opened 2021-05-23
Warning fixes

Allofich:
Fixes for unused result warnings in GCC.

Also removal of redundant casts and parentheses.


joncampbell123:
Another merge request caused a conflict in debug.cpp. Which one is the correct one to use?

Allofich:
Fixed.

Allofich:
The line with the merge conflict was

`if (ch<32 || !isprint(*reinterpret_cast<unsigned char*>(&ch))) ch='.';`

which is changed by this PR to remove the cast. This line is not actually touched by https://github.com/joncampbell123/dosbox-x/pull/2543, it's the line just _before_ the changed code there, but I guess something about the way merges is done caused it to become a conflict anyway.

Issue #100, opened 2015-07-14
DOS/V emulation


joncampbell123:
I think it might be a nice enhancement if DOSBox-X were to emulate the DOS/V extensions of Chinese/Japanese/Korean MS-DOS. Not sure who it would help, but perhaps there are games from that part of asia that might like benefit from that addition.

Jarvik7:
Definitely agree. Japanese emulators tend to be of sketchy quality with a poor record of open source.

See also the predecessor to DOS/V https://en.wikipedia.org/wiki/AX_architecture

sikthehedgehog:
Reminds me, does anybody know what Rusty uses? There's a DOSJP.COM program that gets executed before the game itself (and yes, it runs under DOSBox just fine).

MoochMcGee:
There are some versions of PCem that runs DOS/V just fine. Apparently, it abuses the VGA a bit.

joncampbell123:
Documentation is hard to find, but I managed to find a reference (in Japanese):

http://www.hackipedia.org/browse/Computer/Platform/PC,%20IBM%20compatible/MS-DOS/DOS%E2%88%95V

timofonic:
@joncampbell123  @sikthehedgehog @MoochMcGee @Jarvik7 
What about contacting with Japanese and Chinese people potentially interested on this?

 Think of it as networking, maybe they can reach developers and make them interested in Dosbox-X.

I think the language barrier is a problem, so "ambassadors" are required to make the project rearch the non-western world :)

I found some Japanese sites related to DOS/V:

http://www.hmsoft.co.jp/lepton/software/dosv/index.htm
http://museum.ipsj.or.jp/en/computer/personal/0048.html

WTF?
https://twitter.com/DOSVPOWERREPORT/
http://www.dosv.jp

landloafer:
It has been long time that the QuarterDeck's DEVICE.COM (now included in dosbox-x) can load DOS/V's (1) $font.sys within the dosbox environment. Some may prefer using a privately developed (2) $FONTX driver for later English DOS versions. These ways, dosbox can runs most DOS/V software that require FONT ROM.

Method (1) requires font library from actual DOS/V (if you have the DOS/V diskettes, you can also side load it as guest OS), method (2) requires the user to generate new font library under your own DOS/V, PC98 or J3100. I think for copyright reason, it does not come with any font library. Otherwise, can try DOSVSV, an early font driver with its own fonts.

[...]

PR #2576, opened 2021-06-04
Support for Japanese EGA mode

Wengier:
This adds supports for Japanese EGA mode (a variant of standard EGA mode with Japanese support), by adopting code from the DOSVax fork of @yksoft1. In this mode Japanese characters can be displayed normally with or without the use of TTF fonts. The original DOSVax fork does not provide font files so that jfontsbcs and jfontdbcs options are required to specify other fonts. On the other hand, this PR makes font files fully optional by building-in fonts generated with PC-98 mode's free fonts (though jfontsbcs and jfontdbcs options (now in [render] section) can still be used to specify different fonts). Most (>90%) additions of this PR come from the built-in fonts (coded in jega.h). This may also be another way for SDL drawn menu to support Japanese text.


roytam1:
> adopting code from the DOSVax fork of @yksoft1

what about incorporating code from https://github.com/nanshiki/DOSVAXJ3 so IBM JEGA(i.e. AX) and Toshiba J-3100 mode can also be used?

Wengier:
> > adopting code from the DOSVax fork of @yksoft1
> 
> what about incorporating code from https://github.com/nanshiki/DOSVAXJ3 so IBM JEGA(i.e. AX) and Toshiba J-3100 mode can also be used?

It definitely looks like an interesting branch. It even ported the LFN and clipboard features from my older DOSBox SVN-lfn branch (although DOSBox-X apparently contains more recent code for these features). I think DOSBox-X can certainly incorporate other code as you mentioned too.

joncampbell123:
Is Japanese EGA mode for IBM PCs or PC-98?
I'm not aware of any EGA hardware for PC-98, though PC-98 has it's own digital RGB TTL video output connection as I understand it.

roytam1:
> Is Japanese EGA mode for IBM PCs or PC-98?

for IBM PCs. see also: https://en.wikipedia.org/wiki/AX_architecture

Wengier:
It is a variant of the regular EGA mode rather than PC-98 mode. The base code of both EGA and JEGA are very similar.

Wengier:
I have added support for Japanese DOS/V mode too, by setting `dosv=jp` in the config file. I think it is definitely desired by DOSBox-X (#100). I saw no earlier discussion about Toshiba J-3100 mode though. I did not add the IME feature because it apparently requires the modification of the SDL library, and the DOSVAXJ3's modification was for SDL1 build only.

joncampbell123:
Did you forget to commit jega.h?

    In file included from dos_devices.cpp:32:
    dev_con.h:25:10: fatal error: jega.h: No such file or directory
       25 | #include "jega.h"

Wengier:
@joncampbell123 The file jega.h was renamed to jega.h. Fixed accordingly.

PR #2582, opened 2021-06-09
Includes cleanup

Allofich:
- Removal of unnecessary #include statements
- Localization of #include statements (meaning, move #include statements out of header files that don't actually need them into the inheriting .cpp files that do use them, so that they are not needlessly included in all files that include those header files)
- Minimize #includes (for example, if file A includes a standard library and file B includes file A but only uses it for those standard library functions, then instead of file B including file A have it just include the standard library functions directly)

### DIFF
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -22,6 +22,7 @@
 #ifndef DOSBOX_DOS_INC_H
 #include "dos_inc.h"
 #endif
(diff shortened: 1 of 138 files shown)


Allofich:
Building on macOS with this fails currently. I'm looking at it now.

Allofich:
Should build on macOS now.

PR #2654, opened 2021-07-08
DBCS vector table fix.

nanshiki:
# Description

Fixed a bug related to DBCS vector table.

Fix incorrect return value for int 21h ax=6300h, ax=6507h.

**Additional information**

It is a DBCS vector table, but correctly, the table size is placed at the beginning, and in Japanese, it is 0006h,819fh,0e0fch in words.
The int 21h ax=6300h will return the address pointing to 819fh, while the int 21h ax=6507h will return the address pointing to 0006h.
See below for more information.
http://www.ctyme.com/intr/rb-3142.htm#Table1746
http://www.ctyme.com/intr/rb-3163.htm#Table1756

### DIFF
--- a/include/dos_inc.h
+++ b/include/dos_inc.h
@@ -40,8 +40,8 @@ struct CommandTail{
 #pragma pack ()
 #endif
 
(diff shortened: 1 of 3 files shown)


Wengier:
Thanks a lot for your fix and your work on DOSVAXJ3! Features such as DOS/V and JEGA were ported and extended in DOSBox-X, and these features make the emulator a lot more useful. You are credited in the CREDIT page too. Thanks again!

PR #2657, opened 2021-07-09
IME support in videodriver=windib

nanshiki:
# Description

IME support in videodriver=windib

Fixed the problem that IME could not be used to input text with videodriver=windib. #2572
Other fixes include adjusting the character size during conversion, handling the key buffer when the IME final string is long, and fixing the buffer size problem when getting the IME string.


Wengier:
I can confirm that Japanese and Korean IMEs now work with videodriver=windib (in addition to videodriver=directx). Thanks for the fix.

PR #2671, opened 2021-07-12
Corrected the method of judging whether DBCS characters have been dra…

nanshiki:
# Description

Fixes related to DBCS font acquisition on Linux

Corrected the method of judging whether DBCS characters have been drawn or not.
Fixed that some symbols in DBCS characters are not displayed correctly.

**Additional information**
setlocale() only needs to be called once at the beginning, so it has been moved to JFont_Init().

### DIFF
--- a/src/ints/int_dosv.cpp
+++ b/src/ints/int_dosv.cpp
@@ -58,6 +58,7 @@ static GC font_gc;
 static XFontSet font_set16;
 static XFontSet font_set14;
 static XFontSet font_set24;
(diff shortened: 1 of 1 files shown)


Wengier:
Thanks @nanshiki. The updated code is certainly an improvement over the previous code.

Wengier:
Thanks @nanshiki! I can confirm that both Simplified and Traditional Chinese DBCS characters work great (the latter with WenQuanYi font installed) with the updated code.

Issue #2554, opened 2021-05-27
[Japanese language] Menu translations


maron2000:
I'm trying to prepare a Japanese translation, however, Dosbox-x menu seems that it cannot handle Japanese characters.
I tried file-encoding both UTF-8 and Shift-JIS but it doesn't display properly. 
The screenshots below shows something, but not Japanese.
Are there any other settings that I have to change?
Thanks in advance.

Language file encoded in Shift-JIS
![translations2](https://user-images.githubusercontent.com/68574602/119784509-3d5c4480-bf09-11eb-97cb-ce5cf00d2af6.png)

Language file encoded in UTF-8 (PC-98 mode)
![translations](https://user-images.githubusercontent.com/68574602/119784011-c1fa9300-bf08-11eb-8122-d42af4f86349.png)

Wengier:
First of all, thanks for preparing a Japanese translation.

There are 3 types of menus in various DOSBox-X builds, namely Windows SDL1 menu, macOS NSMenu, and SDL-drawn menu. The first two types of menus should support Japanese characters just fine (with the latest code). The last one (used by Windows SDL2 and Linux builds) unfortunately not at this time. Also, please make sure to use the latest DOSBox-X 0.83.14 testing builds for the Japanese translations, not DOSBox-X 0.83.13 release version, as version 0.83.14 (to be officially released by the end of month) improves CJK support significantly, especially for TrueType font (TTF) output (set `output=ttf` and a Japanese TTF font e.g. `ttf.font=UnifontFullMediumMono.ttf`), and set `country=81,932` so that Japanese support should work for all machine types, not just PC-98 mode.

* Windows SDL1 build: [dosbox-x-update.zip](http://cndos.fam.cx/dosware/dosbox-x-update.zip)

Wengier:
@maron2000 Below is a screenshot showing latest DOSBox-X 0.83.14 Windows SDL1 build with Simplified Chinese translation:

![image](https://user-images.githubusercontent.com/8216923/119788295-056ce700-bea0-11eb-9cdb-265b976b77f3.png)

maron2000:
Tested the DOSBox-X 0.83.14 Windows SDL1 build, and confirmed the menu shown in Japanese.
It's a huge improvement than the present version!

There are some glitches when using half-width Kana, as shown in the first line of the screenshot below.
[...]

PR #2690, opened 2021-07-19
Update zh_TW.lng

1abcd:
corrections for zh_TW

#2554


Wengier:
@1abcd Thanks for the corrections! I could only make automatic Simplified => Traditional conversions plus some manual fixes myself, but this will make the zh_TW language file much better. I appreciate it.

Issue #2558, opened 2021-05-28
Cannot load a couple tsr's high


blindphoneman:
Hi:

Trying to load a couple tsr's high.  They are screen readers, and, under MS-DOS, they do load high with no problem.

I use the lh command with default options.

Do I need to install a memory manager for this to work?

Thanks!

Wengier:
I tried the built-in LH command and I don't think it works with UMB memory with the default setting. However, some DOS programs can auto-load into UMB memory by themselves and can indeed load into UMB in DOSBox-X. Two examples are CTMOUSE and COMMAND.COM from MS-DOS 7.x:

CTMOUSE.EXE: [ctmouse.zip](https://github.com/joncampbell123/dosbox-x/files/6561905/ctmouse.zip)

COMMAND.COM (MS-DOS 7.1): [commmand.zip](https://github.com/joncampbell123/dosbox-x/files/6561909/commmand.zip)
(Enter the command `VER 7.1` before running COMMAND.COM from MS-DOS 7.1).

![image](https://user-images.githubusercontent.com/8216923/120009923-87e1cd80-bfaa-11eb-873b-96af254736a4.png)

I guess the LH command may need some work.

PR #2705, opened 2021-07-22
Fix with how DOS_Execute allocates memory to comply with the memory a…

cimarronm:
…llocation strategy for UMB...now, it will allocate and execute in UMB if there is enough memory for the minimum allocation of the executable

# Description

Currently, `DOS_Execute` will search for the largest block of memory irregardless of the memory allocation strategy set and as a result will almost never execute anything in upper memory. The environment block will be allocated in UMB but none of the executable. This patch aligns with the behavior of `loadhigh` in DOS which will allocate in the upper memory if there is enough memory for the size of the executable.

Should address PR #2558 

Here is an example with snarf (dos screen capture TSR)
_Before Patch_
```
C:\>lh SNARF.EXE

Snarf 2.0 is now installed in memory

C:\>mem /c

Modules using memory below 1 MB:

  Name            Total          Conventional       Upper Memory
  --------  ----------------   ----------------   ----------------
  SYSTEM      28,720   (28K)     28,720   (28K)          0    (0K)
  COMMAND      3,216    (3K)      3,216    (3K)          0    (0K)
  SNARF        3,984    (4K)      3,984    (4K)          0    (0K)
  Free       700,224  (684K)    618,304  (604K)     81,920   (80K)

Memory Type         Total      Used       Free
----------------  --------   --------   --------
Conventional          640K        36K       604K
Upper                  80K         0K        80K
Reserved              304K       304K         0K
Extended (XMS)     15,360K       448K    14,912K
----------------  --------   --------   --------
Total memory       16,384K       788K    15,596K

Total under 1 MB      720K        36K       684K

Total Expanded (EMS)                   15M (15,269,888 bytes)
Free Expanded (EMS)                    15M (15,269,888 bytes)
...
```

_After Patch_
```
C:\>lh SNARF.EXE

Snarf 2.0 is now installed in memory

C:\>mem /c

Modules using memory below 1 MB:

  Name            Total          Conventional       Upper Memory
  --------  ----------------   ----------------   ----------------
  SYSTEM      28,720   (28K)     28,720   (28K)          0    (0K)
  COMMAND      3,216    (3K)      3,216    (3K)          0    (0K)
  SNARF        3,984    (4K)          0    (0K)      3,984    (4K)
  Free       700,224  (684K)    622,288  (608K)     77,936   (76K)

Memory Type         Total      Used       Free
----------------  --------   --------   --------
Conventional          640K        32K       608K
Upper                  80K         4K        76K
Reserved              304K       304K         0K
Extended (XMS)     15,360K       448K    14,912K
----------------  --------   --------   --------
Total memory       16,384K       788K    15,596K

Total under 1 MB      720K        36K       684K

Total Expanded (EMS)                   15M (15,269,888 bytes)
Free Expanded (EMS)                    15M (15,269,888 bytes)
...
```

### DIFF
--- a/include/dos_inc.h
+++ b/include/dos_inc.h
@@ -233,6 +233,7 @@ void DOS_Terminate(uint16_t pspseg,bool tsr,uint8_t exitcode);
 
 /* Memory Handling Routines */
 void DOS_SetupMemory(void);
+uint16_t DOS_GetMaximumFreeSize(uint16_t minBlocks);
 bool DOS_AllocateMemory(uint16_t * segment,uint16_t * blocks);
 bool DOS_ResizeMemory(uint16_t segment,uint16_t * blocks);
 bool DOS_FreeMemory(uint16_t segment);
@@ -670,6 +671,11 @@ public:
 	uint8_t GetType(void) { return (uint8_t)sGet(sMCB,type);}
 	uint16_t GetSize(void) { return (uint16_t)sGet(sMCB,size);}
 	uint16_t GetPSPSeg(void) { return (uint16_t)sGet(sMCB,psp_segment);}
+	enum class MCBType : uint8_t
+	{
+		ValidBlock = 'M',
+		LastBlock = 'Z'
+	};
 private:
 	#ifdef _MSC_VER
 	#pragma pack (1)
--- a/src/dos/dos_execute.cpp
+++ b/src/dos/dos_execute.cpp
@@ -366,6 +366,10 @@ bool DOS_Execute(const char* name, PhysPt block_pt, uint8_t flags) {
 		DOS_AllocateMemory(&pspseg,&maxfree);
 		if (iscom) {
 			minsize=0x1000;maxsize=0xffff;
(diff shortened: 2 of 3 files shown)


Wengier:
@cimarronm Thanks a lot for the patch! I can confirm that LH command can now load programs high with it. For Visual Studio builds though, "std::max" in line 164 of dos_memory.cpp needs to be changed to "max" in order to compile correctly.

Meanwhile, I think it is probably a good idea to add an option to load COMMAND high too, default to "auto" which means "true" if DOS version is at least 7, or "false" otherwise, similar to the "lfn" option for long filename support, because MS-DOS 7.x automatically loads COMMAND high by default, unlike earlier versions of DOS.

cimarronm:
@Wengier can you test the latest with Visual Studio? That's a good idea for COMMAND...I'll add that as well

Wengier:
@cimarronm Thanks for the fix. Yes, the updated code builds on Visual Studio.

Wengier:
@cimarronm Please feel free to add the said option for COMMAND in another PR. Thanks.

PR #2709, opened 2021-07-23
Update zh_TW.lng

1abcd:
change some local terms


Wengier:
Thanks for the update!

PR #2710, opened 2021-07-23
Fixed the display position of IME conversion candidates on Linux.

nanshiki:
# Description
Fixed the display position of IME conversion candidates on Linux.

Fixed the problem that the IME conversion candidate position was far down in Linux SDL2.


Wengier:
@nanshiki Thanks for making it work better. I could test IME on my Linux VM myself, so this will certainly help other people.

PR #2727, opened 2021-07-27
Loading command shell in upper memory

cimarronm:
# Description
Add config option under dos for loading the command shell in upper memory

The the option "loadhigh shell" under the dos section to support loading the command shell in upper memory. A value of `auto` will attempt to load in upper memory if the dos version is set to 7.0 or greater

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -4048,6 +4048,12 @@ void DOSBOX_SetupConfigSections(void) {
 			"Similarly, FAT32 disk images will be supported with a reported DOS version of 7.1 or higher.\n");
     Pstring->SetBasic(true);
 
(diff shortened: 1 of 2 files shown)


cimarronm:
@Wengier Please review and let me know of any changes to make

Wengier:
@cimarronm Thanks! I have tried your code, and with the default command shell (``SHELL=``) it does appear to work fine, but not if you specify a command shell, e.g. ``SHELL=4DOS.COM``. Actually, both COMMAND.COM and 4DOS.COM are loaded in memory in such case for some time, which needs to be fixed too.

I suggest to name the option to e.g. ``shellhigh`` (instead of ``loadhigh shell``) to avoid the space in the name for this.

PR #2772, opened 2021-08-07
Fpu improvements

cimarronm:
# Description
Refactored the RegBit class a bit, fixed typo bug in the fpu_longdouble code, and now accout for the reserved '1' bits in the FPU control word.


joncampbell123:
Have you tested this code in Visual Studio 2019 community edition?

cimarronm:
not yet

joncampbell123:
I'll merge this change, but I expect the next merge request to address whatever about it VS2019 doesn't like. Linux and GCC are happy with it though.

PR #2791, opened 2021-08-13
Fixes related to DOS/V settings and V-Text XGA and 24 pixel fonts.

nanshiki:
# Description
Fixes related to DOS/V settings and V-Text XGA and 24 pixel fonts

Changed the condition of judging whether DOS/V can be used or not.
Scrolling fix in XGA.
Screen clearing bug fix in vtext2.

Addition of 24-pixel font drawing processing.
Added option to substitute 20 pixel font for 24 pixel font.

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -1044,8 +1044,8 @@ void DOSBOX_RealInit() {
     if (!strcasecmp(dosvstr, "ko")) dos.set_kdosv_enabled = true;
     if (!strcasecmp(dosvstr, "chs")||!strcasecmp(dosvstr, "cn")) dos.set_pdosv_enabled = true;
     if (!strcasecmp(dosvstr, "cht")||!strcasecmp(dosvstr, "tw")) dos.set_cdosv_enabled = true;
-    if (svgaCard != SVGA_TsengET4K && svgaCard != SVGA_S3Trio) {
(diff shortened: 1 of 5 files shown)


Wengier:
@nanshiki I see you added 24-pixel font drawing processing for DOS/V. May I ask how to enable 24-pixel font (instead of 16-pixel font) mode for DOS/V mode? Thanks. I tried setting ``dosv=jp/cn/tw`` and ``vtext=xga24/sxga24`` (and/or ``machine=svga_et4000``), but it did not appear to make a difference in DOS/V's font size. It is probably not well-documented in the config file at the time.

nanshiki:
@Wengier The 24-pixel font is supported by V-Text.
It is DOS/V V-Text, but the video mode must be set to 70h.
Initially, DOS/V had only 80x25 mode of video mode 03h, and later V-Text with a variable screen size was added, which is specified to use video mode 70h.
[dosv] The mode specified in vtext is used in video mode 70h.
The specified mode in vtext2 is used in video mode 78h, but after the screen change, 70h is written to 0040h:0049h.
In DOSVAXJ3, the built-in commands CHEV VT and CHEV VT2 can be used to switch between them.

Wengier:
@nanshiki I see. Thanks. I got it working with the addition of VTEXT command to change the V-text mode.

PR #2798, opened 2021-08-15
Update AUR description and mention SDL2 built

rderooy:


FredBezies:
This is closer to the Archlinux reality. Thanks for the pull request!

PR #2823, opened 2021-08-21
SVN commits

Allofich:
https://sourceforge.net/p/dosbox/code-0/4462/ - Skipped because it is unnecessary for DOSBox-X. (I tested and the debugger window can be resized without issue on Windows 10)
https://sourceforge.net/p/dosbox/code-0/4463/ - In this PR
https://sourceforge.net/p/dosbox/code-0/4464/ - Already in DOSBox-X
https://sourceforge.net/p/dosbox/code-0/4465/ - In this PR

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -28,6 +28,9 @@
     done in video BIOS.
     - Implement mouse driver function 0x27, fixes
     Alive - Behind the Moon.
+    - Add log message for unusual cases where an
+    executable unsafely places its initial stack outside
+    the memory block.
 0.83.16
   - Release notes for this version:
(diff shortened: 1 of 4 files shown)


joncampbell123:
Ha, I love it: https://sourceforge.net/p/dosbox/code-0/4462/

DOSBox-X way of dealing with terminal resize: Use the width and height, decide where debug windows go, call newwin() appropriately.

DOSBox SVN way of dealing with terminal resize: Uh, well, if the terminal window is too small, newwin() returns NULL, causes a segfault, and besides it messes up the console. Let's hack the console window to make it non-resizeable :facepalm:

PR #2824, opened 2021-08-21
Fixed EMS set handle name function.

nanshiki:
# Description
Fixed EMS set handle name function.

Use ds:si instead of es:di to specify the handle name.
[int 67h, ax=5301h](http://www.ctyme.com/intr/rb-7435.htm)

### DIFF
--- a/src/ints/ems.cpp
+++ b/src/ints/ems.cpp
@@ -716,7 +716,7 @@ static uint8_t GetSetHandleName(void) {
 		break;
 	case 0x01:	/* Set Handle Name */
 		if (handle>=EMM_MAX_HANDLES || emm_handles[handle].pages==NULL_HANDLE) return EMM_INVALID_HANDLE;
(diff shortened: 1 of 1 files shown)


joncampbell123:
Name pointer is DS:SI? Are you sure about that?

http://hackipedia.org/browse.cgi/Computer/Operating%20Systems/MS%2dDOS/EMS%2c%20Expanded%20Memory/Expanded%20Memory%20Specification%20%28Lotus%2c%20Intel%2c%20Microsoft%29%20v4%2e0%20%281987%2d10%29%2etxt

Firefox_Screenshot_2021-08-21T18-26-30 868Z

joncampbell123:
Wait a minute, you're right. *GET* handle name uses ES:DI, *SET* handle name uses DS:SI. Good catch.

Firefox_Screenshot_2021-08-21T18-29-15 144Z

PR #2872, opened 2021-09-02
Fix more GCC warnings

Allofich:
1. Fixes all `-Wignored-qualifiers` warnings.
These warnings appear on ignored const qualifiers in casts. Most of these were caused by me in https://github.com/joncampbell123/dosbox-x/commit/a4bb7d8221f5ec958e2298fce0e31819009ba78f by using `const` when casting values.

2. Fixes some `-Wunused-parameter` warnings.
There are still a ton of these, but this is a start.


Allofich:
Fixed more -Wunused-parameter warnings.

Allofich:
Fixed remaining -Wunused-parameter warnings.

PR #2885, opened 2021-09-06
PC-98: TRAM: Correctly emulate and document JIS trail word rendering

nmlgc:
# Description

Extensively researched this issue on real hardware, and found that box-drawing characters, gaiji, and NEC's extended chars can in fact be switched mid-char. Also, attributes are always applied to every half individually, as you would expect from looking at VRAM.

(Interestingly, not a single one of the well-known PC-98 emulators had implemented all of this accurately…)


nmlgc:
Whoops, it's still not quite right…

joncampbell123:
That's OK, keep trying. It's a pretty strange corner case in the hardware's design that some games probably rely on.

As you work, please indicate what model of PC-98 you are testing this on. Perhaps it differs between revisions of hardware and PC-9801 vs PC-9821.

That code as I wrote it is based on some observations with a PC-9821 486 laptop and some test code in DOSLIB that I wrote: https://github.com/joncampbell123/doslib/tree/master/hw/necpc98

Issue #2421, opened 2021-04-05
Imcompatibility in filers (FD & FILMTN) PC98 


maron2000:
**Describe the bug**
A poplular 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - Operating system
 - DOSBox-X release version or commit SHA1 this issue occurred on
 - Used configuration, e.g. ```dosbox-x.conf```

**Additional context**

Wengier:
I con confirm this issue. Likely an issue in the implementation of the PC-98 subsystem.

maron2000:
@Wengier I understand that PC-98 emulation is yet to be developed, and highly appreciate your efforts to do so.

joncampbell123:
Download a PC-98 ROM package, extract FONT.ROM and put it in the same directory as your dosbox.conf, try again.

joncampbell123:
FD.COM seems to be reading STDIN, and waiting for the letter 'R'.

If you type SHIFT+R it continues into the program properly.

It also calls INT DCh once before that loop.

ANSI detect?

joncampbell123:
From a real disk image,

INT DCh AX=0000h BX=FF00h CX=000Ch DX=04C3h seems to trigger an ANSI response indicating the cursor.

FD.COM reads back from INT 21h:

bash-5.0# echo -en \\x24\\x1B\\x5B\\x32\\x34\\x3B\\x30\\x31\\x52 | hexdump -C
00000000  24 1b 5b 32 34 3b 30 31  52                       |$.[24;01R|
00000009

joncampbell123:
Alright, actually INT DC CL=0Ch AX=0000h isn't related to the readback.

FD.COM appears to write this at startup, INT 21h AH=9h stops at the $

0910:00000887 1B 5B 3E 35 6C 1B 5B 33 32 3B 31 48 1B 5B 36 6E  .[>5l.[32;1H.[6n
0910:00000897 24 05 5A 3A 5C 00 00 00 00 00 00 00 00 00 00 00  $.Z:\...........

So thats:

    .[>5l             (?)
    .[32;1H           (cursor position)
    .[6n              (?)

joncampbell123:
The correct response is ESC [ *row* ; *col* R

joncampbell123:
[...]

PR #2889, opened 2021-09-08
Fixed function key related functions of int DCh on PC-9801.

nanshiki:
# Description
The character string that can be registered as a function key was limited to 14 bytes, but 15 bytes is correct.
In addition, the acquisition of ax=0,15h-1fh had an extra 1 byte of 0 in it.
dch
#2421 FILMTN issues will be resolved.


joncampbell123:
Sounds good. My implementation was based entirely on taking notes and tracing through the INT DCh handler of a PC-98 MS-DOS 6.22 boot disk.

PR #2892, opened 2021-09-08
Fix MSVC builds failing to mount image files

### LINKED ISSUE #2871: Boot Image not mounting (image geometry auto detection issue?)
Just installed Dosbox-X 0.83.17 (SDL2, Windows 10) and my previous Windows 95 installation config file is not working anymore:

image

The previous version worked without any issue with the same config file which has the following:

[autoexec]
IMGMOUNT C hdd.img
IMGMOUNT D cd.iso -ide 2m
BOOT C:

### LINKED ISSUE #2876: Can't mount ram drive, or image file
when i try to mount .IMG File or Ram Drive , it just says "Cannot Create Drive From File" , why?
image

### LINKED ISSUE #2887: Cannot mount HDD image file anymore in version 0.83.17
I have a HDD image file and used to "imgmount x: x:\virtual\doslife.img" for mounting.

In version 0.83.17, mounting doesn't work anymore:

>Image geometry auto detection: -size 512,63,255,261
>Cannot create drive from file.

In the release notes, there is a short comment about "-bd" option added which I assume is the problem here. Unfortunately, this option is not documented (neither in "imgmount /?" nor in the wiki page) so I don't know how to mount this image now. I have already tried to add "-o partidx=1", but this doesn't work either.

Maybe @joncampbell123 or someone else can give me some hint. Thanks for helping!

Allofich:
Closes #2871. Closes #2876. Closes #2887. (appear to all be the same issue)

I removed the `GCC_ATTRIBUTE(packed);` as I believe it is redundant (it's absent in the `_PC98RawPartition` struct that is just above the `partTable` struct), but if it is still needed I can put it back.

### DIFF
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -471,7 +471,6 @@ struct _PC98RawPartition {
 	uint16_t	end_cyl;	/* end cylinder */
 	char		name[16];
 };
-#pragma pack(pop)
 
 struct partTable {
 	uint8_t booter[446];
@@ -491,7 +490,8 @@ struct partTable {
 #if SECTOR_SIZE_MAX > 512
     uint8_t  extra[SECTOR_SIZE_MAX - 512];
 #endif
-} GCC_ATTRIBUTE(packed);
+};
+#pragma pack(pop)
 
 void updateDPT(void);
 void incrementFDD(void);


Allofich:
I went ahead and searched the whole codebase (only the files that are part of the Visual Studio project) for `GCC_ATTRIBUTE(packed);` for any other cases where the MSVC build would be missing the packed attribute, but in all other cases `GCC_ATTRIBUTE(packed);` is accompanied by an opening
```
#ifdef _MSC_VER
#pragma pack (1)
#endif
```
and a closing
```
#ifdef _MSC_VER
#pragma pack ()
#endif
```
that covers the whole range.

I did find one place, in `includes/ipx.h`, where there is a struct that will be affected by `#pragma pack (1)` but not by  `GCC_ATTRIBUTE(packed);`. I don't know if that matters or not. It's the following struct

```
struct fragmentDescriptor {
	uint16_t offset;
	uint16_t segment;
	uint16_t size;
};
```

I confirmed that the `#pragma pack(push,1)` in `bios_disk.h` works for MSVC, but if
#ifdef _MSC_VER
#pragma pack (1)
#endif
should be used for style consistency or something, just let me know.

PR #2902, opened 2021-09-11
GitHub: Use issue forms

Dobby233Liu:
## Description

Issue forms is a new type of issue template in GitHub, currently in beta. Since DOSBox Staging moved from using Markdown issue templates to use issue forms, I think that change also makes sense to DOSBox-X. So, this is a proof-of-concept to make the repo use issue forms. The templates are based on what DOSBox Staging uses. I'd like some feedback from the maintainers on this, that would be as useful as issue forms itself is.

This PR also changes the pull request template.

~~I still haven't figured out how to squash commits~~


Dobby233Liu:
> If you use GitHub Desktop, you can supposedly squash commits as follows: https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/managing-commits/squashing-commits
> 
> Also, from the side of the side of the DOSBox-X repository, supposedly it can be configured to allow contributors to squash commits: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

I forgot I'm on mobile so I can't do that, but thanks anyway

Dobby233Liu:
I'll setup some environment to test this change now. Wait

Allofich:
Seems like this still needs work. Someone was able to create an issue without filling out the information that is supposed to be required (OS, reproduction steps, etc.). At the issue linked above.

Also, this "question" issue (https://github.com/joncampbell123/dosbox-x/issues/2919) shows as having 2 "tasks" completed. I guess this is caused by the two checkboxes or something.

Dobby233Liu:
> Also, this "question" issue (#2919) shows as having 2 "tasks" completed. I wonder if this is caused by the two checkboxes (the code of conduct/contributing guidelines one and the "I've looked for similar issues" one).

Yes. It's GitHub

Allofich:
> It already exists before issue forms. It can be turned off with configuration if you want

If we have a problem with people frequently choosing that option and then creating issues that lack useful information, maybe we should do that. For now I think we can just wait and see.

> Yes. It's just GitHub being GitHub

They aren't "tasks", so I think it would be good to fix this if possible. I would actually recommend removing those two checkboxes, because many if not most people will probably just check them without doing what they say, in which case they are just pointless extra clicks required for all issue creators. But I do think it is a good idea to have a link to the code of conduct and to ask that existing issues be checked first, so we could leave those just as sentences, without the checkboxes.

Dobby233Liu:
> They aren't "tasks", so I think it would be good to fix this if possible.

Ask GitHub maybe. I can't fix it

> I would actually recommend removing those two checkboxes, because many if not most people will probably just check them without doing what they say, in which case they will mostly just be pointless extra clicks required for all issue creators.
> ...
> so we could leave those just as sentences, without the checkboxes.

Other projects also use checkboxes for these stuff, so I think it is fine, and since i think the notices are good to the issue tracker, I let go of the required-to-tick checkboxes to make the notices easier to notice. But what you have said is completely vaild though

Anyway you can do what you have suggested any time with a PR, you can study this issue forms thingy yourself

(Also the version section isn't really descriptive. Should have fixed that)

Allofich:
> Ask GitHub maybe. I can't fix it

Looks like dosbox-staging also is having "_ tasks done" put next to all the new issues created by the issue forms. Well issue forms is still a new feature so it may change in the future.

Issue #2605, opened 2021-06-19
Doesn't build on arm64 FreeBSD 13.0-RELEASE


clausecker:
**Describe the bug**
I am trying to add a port for dosbox-x to the ports tree.  Besides some minor issues (poorly documented dependencies, some minor patches needed for `include/byteorder.h`), there is one major issue preventing the build right now:

The code does not build on arm64 FreeBSD 13.0-RELEASE due to dosbox-x assuming `long double` is an 80 bit type and supports x86-like `_FPU_SETCW` calls.

Here's what the errors look like:

```
In file included from core_dynrec.cpp:173:
In file included from ./core_dynrec/decoder.h:25:
In file included from ./core_dynrec/dyn_fpu.h:52:
./core_dynrec/../../fpu/fpu_instructions_longdouble.h:29:12: fatal error: 'fpu_control.h' file not found
#  include <fpu_control.h>
           ^~~~~~~~~~~~~~~
```

The byte order issue is patched by applying NetBSD behaviour to FreeBSD, too:

```
--- include/byteorder.h.orig	2021-06-19 13:37:28 UTC
+++ include/byteorder.h
@@ -115,11 +115,11 @@
 #define _BSD_SOURCE
 #include <endian.h>
 
-#elif defined(__NetBSD__) || defined(__OpenBSD__)
+#elif defined(__NetBSD__) || defined(__OpenBSD__) || defined(__FreeBSD__)
 
 #include <sys/endian.h>
 
-#elif defined(__FreeBSD__) || defined(__DragonFly__)
+#elif defined(__DragonFly__)
 
 #include <sys/endian.h>
```

**To Reproduce**
Build dosbox-x on arm64 FreeBSD 13.0-RELEASE (I built with sdl2, but I don't think this makes a difference).

**Expected behavior**
A successful build.

**Environment (please complete the following information):**
 - Operating system: FreeBSD 13.0-RELEASE on arm64
 - dosbox-x-v0.83.14

Wengier:
I don't have a FreeBSD environment to test myself. If you know how to fix the error(s) to make it build successfully on your FreeBSD platform, please feel free to send pull request to fix the issue. Thanks for your support.

P.S. There is now a dedicated [BUILD.md](https://github.com/joncampbell123/dosbox-x/blob/master/BUILD.md) page for building the DOSBox-X source code. If the source code can be built successfully on FreeBSD then such information may be added on the page as well.

clausecker:
[...]

PR #2911, opened 2021-09-13
Doesn't build on arm64 FreeBSD 13.0-RELEASE

Slesa:
# Description

Code compiles on FreeBSD, but SDL code has to be reconsidered (I have just removed the error when using v1)

Fixes ```#2605```

**Additional information**

When starting, it runs into a flickering screen with 
```
LOG: Aspect ratio: 640 x 480  xToY=1.333 yToX=0.750
LOG: menuScale=1
LOG: Screen report: Method 'None' (-1.000 x -1.000 pixels) at (0.000 x 0.000) (-1.000 x -1.000 mm) (-0.039 x -0.039 in) (-1.000 x -1.000 DPI)
LOG: font texture id=2 will make 128 x 256
```
I could need a little hint what to search for to fix it

### DIFF
--- a/include/byteorder.h
+++ b/include/byteorder.h
@@ -115,11 +115,11 @@
 #define _BSD_SOURCE
 #include <endian.h>
 
-#elif defined(__NetBSD__) || defined(__OpenBSD__)
+#elif defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)
 
 #include <sys/endian.h>
 
-#elif defined(__FreeBSD__) || defined(__DragonFly__)
+#elif defined(__DragonFly__)
 
 #include <sys/endian.h>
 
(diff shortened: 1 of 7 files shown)


grapeli:
>When starting, it runs into a flickering screen with
>
Try to run like this:
`dosbox-x -set startbanner=false -set fastbioslogo=true`

Slesa:
Runs now. Maybe those flags should be default then for FreeBSD?

grapeli:
> Runs now. Maybe those flags should be default then for FreeBSD?

I don't think this is just a temporary workaround. It will reappear when dosbox-x resizes its window (this happens quite often).
The same effect will be given by setting `-set char9=false`.

**Try on other window managers first.**
Run dosbox-x window maximized (not floating).
If you are using SDL2 then `-set maximize=true`.

Slesa:
Stupid little me... with './configure --enable-sdl2' it runs much better.

PR #2920, opened 2021-09-17
Japanese Windows 3.1 related fixes

nanshiki:
# Description
Fixed a bug that caused setup.exe in Japanese Windows 3.1 to exit with an error, and a bug that prevented the Microsoft version from starting in enhanced mode.

Fixed setup.exe not being able to start.
Fixed failure to start Microsoft Japanese Windows 3.1 in enhanced mode.
Fixed a problem with the Windows 3.1 splash screen.

**Additional information**
The reason why setup.exe could not start was because it could not find $IBMADSP by tracing the NUL device header in the int 21h ah=52h variable area.
Microsoft Japanese Windows could not start in enhanced mode because it needed to return the address of the Windows Startup Information structure with instance data in int 2fh ax=1605h.

### DIFF
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -1714,7 +1714,7 @@ static Bitu DOS_21Handler(void) {
                             break;
                         }
                     }
-                    if(!strncmp(name_start, "$IBMAFNT", 8) || !strncmp(name_start, "$IBMADSP", 8)) {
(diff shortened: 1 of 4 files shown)


maron2000:
I confirmed that I can setup and start Windows 3.1 Japanese (Microsoft) version with `dosv=jp` option.

PR #2926, opened 2021-09-19
Fix warnings and a little more

Allofich:
Fixes many -Wsign-compare warnings, for comparisons between signed and unsigned variables.
Adds gui_tk.h to the Visual Studio project as it was missing.
Fixes redefinition warnings for MIN and MAX (which were being redefined to functionally equivalent statements) by removing the second definitions. Fixes redefinition for MOVE_POS, which was redefined to a functionally different statement, by renaming the redefinition to MOVE_POS2 (both MOVE_POS definitions were only used within the same file they were declared in, so there should be no problem with the rename affecting use in other files, as there isn't any).


Allofich:
These warnings (sign/unsign compares and redefinitions) are also generated in Visual Studio builds, so this also reduces the number of warnings in those builds.

Allofich:
Fixed all the remaining -Wsign-compare warnings. This involved changing a number of variables from signed to unsigned (to avoid using casts all over the place). Most are just iterators that were only ever incremented from 0 upwards and would clearly never go negative, but a few were x, y values for screen coordinates. I don't think these would ever go negative, but if they do then these changes will need to be undone.

PR #2939, opened 2021-09-23
Slightly reword bug report template

Dobby233Liu:
To make the wording a little more clear.
I still didn't thought for a good way to get rid of that two checkboxes.


rderooy:
In addition to asking if the user checked for similar issues, it may also be good to mention the wiki. It happens regularly that the users question is already answered by the wiki.

Dobby233Liu:
In the question or bug report or feature request or all of that template you mean?

PR #2944, opened 2021-09-24
Fix virtual file names being logged as (null)

Allofich:
Summary of changes brought by this PR:

Fixes the file names of virtual DOSBox-X files like "AUTOEXEC.BAT" and "LOADFIX.COM", etc. from being logged as "(null)" when they are read or closed.

For example, enable file IO logging output to a file and simply open and close DOSBox-X. Check the log and you will see after AUTOEXEC.BAT is opened there are entries of reading from "(null)" and closing "(null)" in the current code. With this PR they will correctly show "AUTOEXEC.BAT".

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.83.18
+  - Fixed the names of virtual files being logged in
+    some messages as "(null)". (Allofich)
   - Fixed packed structure alignment problem with MSVC
     (Microsoft C++) builds regarding some disk image
     formats. These problems prevented VS2019 builds from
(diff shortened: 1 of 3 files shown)


Allofich:
Looks like files that are read from a mounted image file are also logged as (null). I haven't fixed that.

Allofich:
Applied the same fix for mounted image files.

The issue is already handled in, for example, `drive_iso.cpp` by calling `SetName` within the `isoFile` constructor, but the static analyzer SonarLint doesn't like that because it means calling an overridable method from a constructor/destructor. https://rules.sonarsource.com/cpp/RSPEC-1699 It also would probably require adjusting the constructor in `drive_virtual.cpp` to include the file name as a parameter, so I just copied the way that the flags were being set outside the constructor and set the name similarly, making it a one-line fix for each issue.

PR #2951, opened 2021-09-30
CMOS log cleanup

Allofich:
Cleanup and rewording of some CMOS logging.

I noticed while installing MS-DOS 6.22 in DOSBox-X that the following message was output to the log

`ERROR BIOS:CMOS:WRite to unhandled register 17`

According to http://www.bioscentral.com/misc/cmosmap.htm, register 17 is for "System Configuration Settings" and has these settings:
Bit 7 = Mouse support disable/enable
Bit 6 = Memory test above 1MB disable/enable
Bit 5 = Memory test tick sound disable/enable
Bit 4 = Memory parity error check disable/enable
Bit 3 = Setup utility trigger display disable/enable
Bit 2 = Hard disk type 47 RAM area (0:300h or upper 1KB of DOS area)
Bit 1 = Wait for <F1> if any error message disable/enable
Bit 0 = System boot up with Numlock (off/on)

I looked at the code in `hardware/cmos.cpp` and found that any write to a CMOS register past 0x0F (except for 0x32) will create this message, but the write is still performed. Reading from the registers is similar, but logs its message for such cases as `LOG_NORMAL`, not an error. Since the value the calling program tries to write is in fact written (almost, see below) and the read is not an error, I don't see why this is being called an error, so this PR changes it to LOG_NORMAL. Also I noticed messiness such as typos and lack of clarity in log messages, and redundant comparisons to `date_host_forced` when it is known to be false, so I fixed those as well.

Now something is I noticed that for some reason writes to some CMOS registers are ANDed with 0x7f, so bit 7 can never be written as 1. According to the site I linked above bit 7 mean something in these registers, such as in register 17 as shown above, or in register 10, where it means "Update in progress". Register 10 is also prevented from being written in DOSBox-X by being ANDed with 0x7f. Is it correct to prevent programs from writing bit 7?

The ANDing with 0x7f looks to have been pulled from SVN and SVN-Daum. @joncampbell123, do you know why they are there?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.83.18
+  - Cleaned up and more accurately worded CMOS-related
+    log messages. (Allofich)
   - Fixed the names of virtual files and files on mounted
     FAT drive images being logged as "(null)". (Allofich)
   - Fixed packed structure alignment problem with MSVC
--- a/src/hardware/cmos.cpp
+++ b/src/hardware/cmos.cpp
@@ -203,7 +203,7 @@ static void cmos_writereg(Bitu port,Bitu val,Bitu iolen) {
         case 0x01:      /* Seconds Alarm */
         case 0x03:      /* Minutes Alarm */
         case 0x05:      /* Hours Alarm */
-            LOG(LOG_BIOS,LOG_NORMAL)("CMOS:Trying to set alarm");
+            LOG(LOG_BIOS,LOG_NORMAL)("CMOS:Writing to an alarm register");
             cmos.regs[cmos.reg] = (uint8_t)val;
             return;     // done
         }
@@ -241,13 +241,13 @@ static void cmos_writereg(Bitu port,Bitu val,Bitu iolen) {
     case 0x03:      /* Minutes Alarm */
     case 0x05:      /* Hours Alarm */
         if(!date_host_forced) {
-            LOG(LOG_BIOS,LOG_NORMAL)("CMOS:Trying to set alarm");
+            LOG(LOG_BIOS,LOG_NORMAL)("CMOS:Writing to an alarm register");
             cmos.regs[cmos.reg]=(uint8_t)val;
             break;
         }
     case 0x0a:      /* Status reg A */
         cmos.regs[cmos.reg]=val & 0x7f;
-        if ((val & 0x70)!=0x20) LOG(LOG_BIOS,LOG_ERROR)("CMOS Illegal 22 stage divider value");
+        if ((val & 0x70)!=0x20) LOG(LOG_BIOS,LOG_ERROR)("CMOS:Illegal 22 stage divider value");
         cmos.timer.div=(val & 0xf);
         cmos_checktimer();
         break;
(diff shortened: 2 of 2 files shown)


joncampbell123:
Looking at that list... there seems to be a common theme where bits 6-0 are writeable and bit 7 is read only status.

For example, what good would it do for the system to write the "CMOS battery dead" bit? Or the IRQ flag pending bit? Or the "Do not write because the RTC is updating the clock" bit?

There's no reason though for register 11h to mask by 0x7F, that might have come from any mistakes or misconceptions about the RTC variables inherited from other forks, or just in general.

joncampbell123:
I can't think of any legitimate reason to mask registers 0-9 by 0x7F, other than perhaps there's no such thing as a clock that reads as "it's 244-o-clock and 192 minutes past the top of the hour". Perhaps there's an undocumented quirk in that the registers internally are smaller than 8 bits because you can fit hours (0-23) in 5 bits, minutes and seconds (0-59) in 6 bits, months (0-11 or 1-12) in 4 bits, day of month (1-31) in 5 bits. It's entirely possible older RTC chips might hold the year in 5, 6 or 7 bits if they weren't designed to run past 2020 or so. Later RTCs might have just made them a full 8 bits, or not. Just a guess.

joncampbell123:
This reference says that you can read the value as BCD if you set the right bit: http://hackipedia.org/browse.cgi/Computer/Platform/PC%2c%20IBM%20compatible/RTC%2c%20CMOS/CMOS%2dMemory%20Map%20by%20Ralph%20Brown%20%282000%2d07%2d16%29%2etxt

That doesn't affect the valid range of values then.

However notice the "alarm time" registers. 0xC0-0xFF seems to be a "don't care" (i.e. there is no alarm), which suggests that maybe the registers are 8 bits wide, at least for the alarm clock values.

joncampbell123:
Finally, I think this page [http://www.bioscentral.com/misc/cmosmap.htm] seems to have made a mistake for Status Register B. One bit got missed and the bit count is off by one. Notice no mention of the binary/BCD selection bit?

joncampbell123:
If you want to know the exact register map, the data sheet itself is your best bet: http://hackipedia.org/browse.cgi/Computer/Platform/PC%2c%20IBM%20compatible/RTC%2c%20CMOS/MC146818A%20CMOS%20REAL%2dTIME%20CLOCK%20PLUS%20RAM%20%281984%29%2epdf

joncampbell123:
Hold on a second, bit 7 of Register B is writeable. It's not a status on whether the RTC is updating the clock, it's a bit the program sets to stop the RTC when changing the date and time.

image
image

Allofich:
Another reference, btw: https://stanislavs.org/helppc/cmos_ram.html
This matches what you said about Status Register B being incorrect (missing BCD toggle and then off-by-one) on the other site.

Allofich:
> I can't think of any legitimate reason to mask registers 0-9 by 0x7F

Those one's aren't being masked in the DOSBox-X code. 0xA, 0xB, 0xF, and everything after 0xF except 0x32 are.

Allofich:
Also case 0xc is falling through to 0xd if `date_host_forced` is false, which seems wrong.

```
    case 0x0c:      /* Status reg C */
        if(date_host_forced) break;
    case 0x0d:      /* Status reg D */
        if(!date_host_forced) {
            cmos.regs[cmos.reg]=val & 0x80; /*Bit 7=1:RTC Power on*/
        }
        break;
```

According to https://stanislavs.org/helppc/cmos_ram.html, both 0xc and 0xd are read-only.

Allofich:
A minor note, I made a mistake in my first post. The "register 17" of the error message referred to 17 in hex, or register 23 in decimal. I updated the post.

Allofich:
This may be useful for emulating the CMOS:
https://www.tme.eu/Document/1990376f4a283f591c5baf9f5154462a/ds12885-887a.pdf

I found it when looking up the IRQF bit in register 0xC (bit 7, so the one bit that can be set currently because of the fallthrough to case 0xD). It has detailed information such as when the IRQF bit is set or cleared.

PR #2977, opened 2021-10-16
Correction of 0x00 to 0x1f characters in AX

nanshiki:
# Description
Correction of 0x00 to 0x1f characters in AX.

I found the font design document for AX and added the built-in font for AX.

**Additional information**
The characters 0x00 to 0x1f are different between DOS/V and AX. If you run a program for DOS/V in AX mode with this fix, the ruled line will be corrupted, but it is correct.
We have confirmed that they are displayed correctly in the Vz Editor for AX.


Wengier:
@nanshiki I noticed that the box-drawing characters in the JEGA/AX welcome screen looks strange in 0.83.19 with the setting `output=ttf`, and doing a `git bisect` shows it has something to do with this change. You can see the difference between 0.83.18 and 0.83.19 below:

DOSBox-X 0.83.18 (with machine=jega, output=ttf):
image

DOSBox-X 0.83.19 (with machine=jega, output=ttf):
image

Do you know how to fix this? Thanks!

P.S. There appears to be currently no cursor in JEGA/AX mode at all (any output). I wonder if this is expected for this mode?

nanshiki:
@Wengier #3054 fixed it.
In this environment, the cursor is blinking in the AX setting.

Issue #2971, opened 2021-10-10
a game cannot run


280696516:
### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

### Have you checked that no other similar bug report(s) already exists?

- [X] I have searched and didn't find any similar issues.

### What operating system(s) this bug have occurred on?

win10

### What version(s) of DOSBox-X have this bug?

0.83.18 SDL

### Describe the bug

[liao.zip](https://github.com/joncampbell123/dosbox-x/files/7318341/liao.zip)
execute play.bat, then black screen.

### Steps to reproduce the behaviour

dosbox.conf
[autoexec]
# Lines in this section will be run at startup.
# You can put your MOUNT lines here.

mount c ./liao
c:
PLAY.BAT

grapeli:
Does not work with intermittent settings. I was curious because even with dosbox-0.65 it works fine.

The (very) old versions of dosbox-x work fine.
Has stopped working since this [commit](https://github.com/joncampbell123/dosbox-x/commit/ec382ea9ffa2ee9784b05765ef9c0642d909dded).

f986c6cfb4e60afeb7f37015f5883fad35d662cf  bad #v0.801
f1d10dbea0294b4fdf50f4cb18ab544934f80592  bad #mount command now accepts -q option to not print anything unless error.
45ab8926c7aac8712df3af2fab5a27916c851fb0  bad #-debug switch at startup
ec382ea9ffa2ee9784b05765ef9c0642d909dded  bad #add date, update intro text. add code to auto-patch VS2008 config.h.
fea484d43a50655f4f6154dc6ae44060da3784f8  good #new code to generate version string
d682534b9cd80c7ddb2324f47a4dbf08a782c4af  good #if misc at debug loglevel, print SDL video state
6ba2abd942920bc343f48dfb6f874971e15c20c4  good #<dosbox-x-windows-6ba2abd942920bc343f48dfb6f874971e15c20c4.zip>

grapeli:
@Allofich 
I used this configuration file. It works fine under it.
[good.conf.txt](https://github.com/joncampbell123/dosbox-x/files/7357617/good.conf.txt)

grapeli:
The x86_64 sdl1 and sdl2 versions work under linux with this configuration.
The key option is:
```
[dosbox]
mainline compatible bios mapping = true
```
with `mainline compatible bios mapping=false` does not start.

[...]

PR #2980, opened 2021-10-17
Set some fixed BIOS addresses.

Allofich:

Sets some fixed BIOS addresses.

The values are from DOSBox SVN (see pre-change values in https://github.com/joncampbell123/dosbox-x/commit/d548d1eb222d9702a69a9dfc575f86556fe40d3b) and the INT5 value is from the table seen at https://www.vcfed.org/forum/forum/genres/pcs-and-clones/1215993-fixed-addresses-in-the-bios-what-program-still-uses-them. That table also agrees with the DOSBox SVN values where they overlap, and I added labels from there as source comments.

## What issues does this PR address?

https://github.com/joncampbell123/dosbox-x/issues/2971

The game still has at least one remaining issue: missing music when run from the DOSBox-X shell that is present when run from a booted MS-DOS image.

## Additional information

Two values caused problems and are commented out.

1) The BIOS_DEFAULT_RESET_LOCATION (for non-PCJr) causes DOSBox-X to crash on startup. (The PCJr one works.)
2) The BIOS_DEFAULT_IRQ1_LOCATION causes keyboard input at the DOS prompt to not work. Trying to type causes a freeze.

@joncampbell123, any ideas? `ROMBIOS_GetMemory` looks like it protects against itself overwriting used memory on subsequent calls, but maybe those two locations get overwritten somewhere.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.83.19
+  - Set some parts of the emulated BIOS to addresses
+    that they are fixed to in real BIOSes. (Allofich)
   - Fix SELINFO command not showing output correctly
     in the debugger window. (Allofich)
   - Better CMOS register B and C emulation. (Allofich)
--- a/src/ints/bios.cpp
+++ b/src/ints/bios.cpp
@@ -9405,13 +9405,30 @@ public:
         }
 
         /* pick locations */
-        BIOS_DEFAULT_RESET_LOCATION = PhysToReal416(ROMBIOS_GetMemory(64/*several callbacks*/,"BIOS default reset location",/*align*/4));
-        BIOS_DEFAULT_HANDLER_LOCATION = PhysToReal416(ROMBIOS_GetMemory(1/*IRET*/,"BIOS default handler location",/*align*/4));
-        BIOS_DEFAULT_INT5_LOCATION = PhysToReal416(ROMBIOS_GetMemory(1/*IRET*/, "BIOS default INT5 location",/*align*/4));
-        BIOS_DEFAULT_IRQ0_LOCATION = PhysToReal416(ROMBIOS_GetMemory(0x13/*see callback.cpp for IRQ0*/,"BIOS default IRQ0 location",/*align*/4));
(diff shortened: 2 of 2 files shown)


Allofich:
PC-98 continues to use the behavior of the current code.

Allofich:
Looking into BIOS_DEFAULT_IRQ1_LOCATION, it fails because the address it tries to get, F000:E987, is already claimed by "DOSBox callback area", so nothing gets allocated.

joncampbell123:
ROMBIOS_GetMemory() has optional parameters to specify alignment and a fixed memory address to allocate at.

One good example of this is in src/ints/bios.cpp line 10072 where the fixed table of the first 128 characters of the 8x8 font are allocated.

See also include/bios.h which defines:

    Bitu ROMBIOS_GetMemory(Bitu bytes,const char *who=NULL,Bitu alignment=1,Bitu must_be_at=0);

Allofich:
I am using ROMBIOS_GetMemory, but looks like "DOSBox callback area" allocates its memory first.

BIOS_DEFAULT_RESET_LOCATION (for non-PCJr) also fails to be allocated. I want to give it 1040475-1040539 (64 bytes). But the nearby area at the time of the call to ROMBIOS_GetMemory is:

1041516 1047149 DOSBox callback area
1040501 1041515 free
1040481 1040500 BIOS version string

so there is no room for it.

joncampbell123:
Perhaps the memory allocator needs a method that can be called to reserve memory above a certain point, then the limit can be removed when the BIOS sets itself up. That way the callback array can be prevented from allocating itself anywhere above, say, F000:8000.

joncampbell123:
I have some free time today, I could add that, incorporate your changes, and commit them myself to improve things.

Allofich:
Sure, please feel free to go ahead.

Allofich:
The clash between the supposed fixed address for BIOS_DEFAULT_RESET_LOCATION at 0xFE05B and the fixed value DOSBox-X is using for the BIOS string at 0xFE061 will need to be resolved somehow, as that only gives 6 bytes for BIOS_DEFAULT_RESET_LOCATION, but the current code wants to give it 64 bytes.

joncampbell123:
@Allofich I fixed a few bugs and updated BIOS allocation based on your pull request. That seems to make Liao happy, though still no MIDI music.

Issue #1274, opened 2019-10-04
Unable to type input to start game


Allofich:
I found this recent bug report for DOSBox SVN:
https://sourceforge.net/p/dosbox/bugs/514/

The game (game is in Chinese, called 爆笑三国志, "Bàoxiào sānguózhì") is attached to the bug report. Unzip to a folder, open the folder and run `SANGO.EXE` to start it.

After pressing a key to get past the title screen, another screen will appear. In DOSBox SVN you can type here, then advance through the menus to start the game, but in DOSBox-X you can't type anything and you can't, as far as I can tell, advance into the game. No errors in the log.

![sango_000](https://user-images.githubusercontent.com/19624336/66208775-e29d2600-e6f0-11e9-8660-49790655e803.png)

In SVN, you can input text into the rectangle in the center of the screen, and advance to the next screens.

By the way, I checked out the gameplay in SVN and found out that what the bug report means. The player and computer players take turns moving a number of tiles across the game board. It is supposed to be a random number, but both the player and computer roll a 5 every time and always move 5 tiles. Since I can't start the game in DOSBox-X I don't know if it happens in DOSBox-X.

Allofich:
Here's a video of the game (it's in fast-forward for some reason).
https://www.youtube.com/watch?v=1O_NUH03lDE
It's working correctly here, the game starts and the movements are not always 5.

joncampbell123:
I see a loop polling port 61h, not to check if keyboard input is there but to wait on bit 4 as a delay loop of some kind.

    033B:0000D890 50                  push ax
    033B:0000D891 51                  push cx
    033B:0000D892 B9D007              mov  cx,07D0
    033B:0000D895 E461                in   al,61
    033B:0000D897 2410                and  al,10
    033B:0000D899 3AC4                cmp  al,ah
    033B:0000D89B 74F8                je   0000D895 ($-8)         (up)
    033B:0000D89D 8AE0                mov  ah,al
    033B:0000D89F E2F4                loop 0000D895 ($-c)
    033B:0000D8A1 59                  pop  cx
    033B:0000D8A2 58                  pop  ax
    033B:0000D8A3 C3                  ret

joncampbell123:
There is code to count using the timer counter in the BIOS data area.

    033B:0000D916 B84000              mov  ax,0040
    033B:0000D919 8EC0                mov  es,ax
[...]

PR #2981, opened 2021-10-18
Use Chinese titles of games

Allofich:
Puts the Chinese titles of two games in the changelog.

聊齋誌異之幽谷傳奇 is the game referred to in https://github.com/joncampbell123/dosbox-x/issues/2971.

爆笑三国志 is the game referred to in https://github.com/joncampbell123/dosbox-x/issues/1274.

The reason is that these are the actual titles, and the transliterations "Liáozhāi zhì yì zhī yōugǔ chuánqí" and "Bàoxiào sānguózhì" don't bring up search results for these games, while the Chinese character titles do.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,9 +9,10 @@
     and force dynamic allocation below that address to
     prevent further conflicts should others submit
     pull requests for additional fixed locations. This
-    fixes some games like "Liao" which appear to assume
-    some of these fixed BIOS addresses and will crash
-    if they are wrong.
+    fixes some games like "聊齋誌異之幽谷傳奇" (Liáozhāi
+    zhì yì zhī yōugǔ chuánqí) which appear to assume
+    some of these fixed BIOS addresses and will
+    crash if they are wrong.
   - Fixed bugs in the region memory allocator C++ class
     that claimed the entire free block even if the
     size wanted was smaller. Added field to remember if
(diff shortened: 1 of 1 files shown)


Allofich:
From the little I can find on that game, 爆笑三国志 seems to be the title used online. I found a YouTube video and a site selling a copy that both use that title. From looking at the video the actual in-game title looks to be 三国誌, but I can't find anything on the game searching with that, so I guess I'll keep it as it is.

Allofich:
@Dobby233Liu, does 爆笑三国誌 (Bàoxiào sānguózhì) seem OK to you for the second game's name?

Dobby233Liu:
> @Dobby233Liu, does 爆笑三国誌 (Bàoxiào sānguózhì) seem OK to you for the second game's name?

Looks fine

joncampbell123:
@1abcd It might be the best policy to match whichever writing the game uses for it's own title, even if it mixes the two. I assume the game developer may have artistic reasons for whichever they chose for their game. That's what I'd do, at least.

1abcd:
@joncampbell123
未命名

国 is not contained in code page 950.
And this page https://www.ruten.com.tw/item/show?21819995322528
The text on 5.25 inch floppy is 爆笑三國志.
For example, Seven-Eleven uses one lower case n in their logo.

Allofich:
@1abcd, thank you for your input.

> The text on 5.25 inch floppy is 爆笑三國志.

This seems like the name to go with. I will update it again.

PR #2982, opened 2021-10-18
Set AX to 0000 when INT 21 AX=4B03 is called

Allofich:

Sets AX to 0000 when INT 21 AX=4B03 (LOAD AND/OR EXECUTE PROGRAM, overlay flag) is called, as verified to happen in MS-DOS. This is checked for by the game "聊齋誌異之幽谷傳奇" (Liáozhāi zhì yì zhī yōugǔ chuánqí), which won't run it's music code if AX is not 0 on return from the call.

If there is any related issues, add their numbers there. Ex. #1234

Closes 2971.


joncampbell123:
I can confirm UK.EXE now plays MIDI music.

Issue #3012, opened 2021-10-26
WordStar macros that access the file system fail under DOSBox-X but not vDosPlus or TameDOS


RobertJSawyer:
### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

### Have you checked that no other similar bug report(s) already exists?

- [X] I have searched and didn't find any similar issues.

### What operating system(s) this bug have occurred on?

Windows 7 32-bit SP1

### What version(s) of DOSBox-X have this bug?

0.83.18 both SDL1 and SDL2 32-bit

### Describe the bug

DOSBox-X has trouble with WordStar for DOS 7.0 macros that access the file system to overwrite or delete existing files.

The attached macro SAMPLE.WSM copies DOSBox-X's README.TXT from DOSBox-X's drive Z: to a file named TESTFILE.TXT in the current folder, and then copies TESTFILE.TXT five times to files named COPY1.WS, COPY2.WS, COPY3.WS, COPY4.WS, and COPY5.WS.

I've provided error handling in the macro to deal with prompts to confirm overwriting or for files not found, so that it should run successfully regardless of whether TESTFILE.TXT or COPY?.WS already exist.

Furthermore, the macro sets WordStar to help level 0 while it is running, which turns off all confirmations, so the error handling is actually redundant.  The macro should simply run to completion with no prompting for confirmations by the user, no matter how many times the macro is invoked.

If you run the attached macro SAMPLE.WSM from WordStar's OPENING MENU (by selecting MP, for Macro Play), it succeeds the FIRST time, when none of the files it tries to overwrite already exist.

But if you run it a SECOND (or subsequent time), it will crap out, first by asking the user for confirmations that should not be required, and, second, by dropping parts of commands.

The macro runs flawlessly under vDosPlus by @Wengier  as well as TameDOS, but fails under DOSBox-X.

### Expected behavior

WordStar should be able to overwrite or delete existing files using its own file-copy or file-delete commands from within a macro.

### Steps to reproduce the behaviour

1. Unzip the attached WordStar for DOS 7.0 macro named SAMPLE.WSM to the folder that contains your WordStar macros (commonly either C:\WS or C:\WS\MACROS).
2. Run WordStar for DOS 7.0
3. From WordStar's opening menu, run the macro SAMPLE by selecting M then P (for Macro Play) or Alt-U,M,P if using the pulldown menus)
[...]

PR #3015, opened 2021-10-27
Fix WordStar macro issues

nanshiki:
Fixed a bug in #3012. 

**Additional information**
Inside DOS_FindDevice(), I'm checking if I can open a device registered at device.com.
In order to determine if the device is a registered device, GetInformation() is called, but the GetInfomation() of the built-in CON device calls int 16h, which seems to be the cause of this phenomenon.
It has been modified so that DOS_FindDevice () does not call GetInformation () of the built-in CON device.

### DIFF
--- a/src/dos/dos_devices.cpp
+++ b/src/dos/dos_devices.cpp
@@ -799,6 +799,8 @@ uint8_t DOS_FindDevice(char const * name) {
 						break;
 					}
 				}
+			} else {
+				break;
 			}
 		}
 	}


joncampbell123:
FYI the original reason the CON driver calls INT 16h is that Microsoft SCANDISK.EXE hooks INT 16h and expects INT 21h to call down to INT 16h, or else certain keyboard "hooks" do not work properly.

PR #3018, opened 2021-10-28
Set DS to 0x40 before calling INT 9 in PCjr

### LINKED ISSUE #2453: The SHAMUS game freezes when setting machine = pcjr
**Describe the bug**
In DOSBox-X v0.83.12 x64 MinGW SDL1 or in DOSBox-X v0.83.13 x64 MinGW SDL1, the SHAMUS game freezes after about 10 seconds of play with machine = pcjr and standard dosbox-x.conf. In DOSBox-ECE r4444, DOSBox-Staging v0.76, when setting machine = pcjr, the SHAMUS game works fine, without freezing. Also, this game works fine without freezes in 86Box on the emulated PCjr machine.

**To Reproduce**
1. Setting in dosbox-x.conf:
    machine = pcjr
2. Launch the SHAMUS game and start playing by moving your character by pressing and holding the arrow keys or the arrow 
    keys on the numeric keypad 

**Expected behavior**
SHAMUS should work fine without freezing when setting machine = pcjr 

**Screenshots**
shamus_000

 - Windows 10 x64 v1909
 - DOSBox-X MinGW SDL1 x64 v0.83.12 or v0.83.13
 - Used configuration, e.g. ```dosbox-x.conf``
   machine = pcjr

**Additional context**
SHAMUS game to reproduce the bug:
[SHAMUS.ZIP](https://github.com/joncampbell123/dosbox-x/files/6329742/SHAMUS.ZIP)

Allofich:

## What issues does this PR address?

Closes #2453.

Essentially the same fix as https://sourceforge.net/p/dosbox/code-0/3884/ (incorporated into DOSBox-X as https://github.com/joncampbell123/dosbox-x/commit/3b6aba9d6e4bfdee08d9b460d2bbfa14a373bdb6). Currently when running the game SHAMUS, that code is not run, but rather the "PCjr NMI Keyboard" routine defined in `bios_keyboard.cpp` is run.

The game runs without issue in MAME, so I traced the execution there and found it was setting DS to 0x40 before calling INT 9. I tried adjusting "PCjr NMI Keyboard" to set DS to 0x40 before the call to INT 9, and the crashes stopped.

### DIFF
--- a/src/ints/bios_keyboard.cpp
+++ b/src/ints/bios_keyboard.cpp
@@ -1469,9 +1469,13 @@ void BIOS_SetupKeyboard(void) {
          * a+4 = iret (1 bytes) */
         phys_writeb(a+5,0x50);          /* push ax */
         phys_writew(a+6,0x60E4);        /* in al,60h */
-        phys_writew(a+8,0x09CD);        /* int 9h */
-        phys_writeb(a+10,0x58);         /* pop ax */
-        phys_writew(a+11,0x00EB + ((256-13)<<8));    /* jmp a+0 */
+        phys_writeb(a+8,0x1e);          /* push ds */
+        phys_writew(a+9,0x406a);        /* push 0x0040 */
+        phys_writeb(a+11,0x1f);         /* pop ds */
+        phys_writew(a+12,0x09CD);       /* int 9h */
+        phys_writeb(a+14,0x1f);         /* pop ds */
+        phys_writeb(a+15,0x58);         /* pop ax */
+        phys_writew(a+16,0x00EB + ((256-18)<<8));    /* jmp a+0 */
     }
 
     if (IS_PC98_ARCH) {


joncampbell123:
I will accept this for now, but know that opcode 0x6A does not exist on the 8086/8088 and the ISR will fail to execute properly if cputype=8086.

It will need to use a PUSH <imm word> opcode if cputype=8086 to run correctly.

http://mlsite.net/8086/

joncampbell123:
I just fixed the NMI handler to use 8086-level instructions as I had described.

Allofich:
OK, but I copied that part from the code of the SVN and DOSBox-X commits I referenced above. That code still exists under `case CB_IRQ6_PCJR:	// pcjr keyboard interrupt` in `callback.cpp.` SHAMUS wasn't using that code as far as I could see, so it may be that it is not ever used. I would think it should either be removed if not used, or if it is used, it may need the same fix to use 8086-level instructions.

The actual PCjr BIOS code that SHAMUS called while running on MAME can be seen in http://bitsavers.trailing-edge.com/pdf/ibm/pc/pc_jr/PCjr_Technical_Reference_Nov83.pdf in the BIOS code at line 138B.

```
; ---------------------------------------------------------------
; THIS SUBROUTINE SETS DS TO POINT TO THE BIOS DATA AREA
; INPUT: NONE
; OUTPUT: DS IS SET
; ---------------------------------------------------------------
DDS      PROC   NEAR
             PUSH   AX
             MOV    AX, 40H
             MOV    DS, AX
             POP    AX
             RET
DDS      ENDP
```

PR #3036, opened 2021-11-04
Fix 0x0F opcode being accepted on 80186

Allofich:
According to https://www.pcjs.org/documents/manuals/intel/8086/

> POP CS (0x0F)
> 
> This single-byte opcode existed only on the 8086/8088, and was of limited use, since it altered CS without a corresponding IP.
> 
> There is no POP CS instruction on later x86 CPUs. The opcode was explicitly made invalid on the 80186/80188, but was reused on later CPUs (starting with the 80286) as the first byte in a series of two-byte opcodes.

Under current code, the 80186 accepts 0x0F as the first byte of a two-byte opcode like the 286+, but this PR makes it an illegal opcode instead to match what the above site says.

Note that the current code misleadingly looks like it would make 80186 wrongly treat 0F like the 8086 does, but the 80186 shares the 80286 code and CPU_CORE is set to CPU_ARCHTYPE_286 for it, so `CPU_CORE < CPU_ARCHTYPE_286` won't be entered for it.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 0.83.20
+  - Fixed 0x0F opcode being valid on 80186 core when
+    it should be invalid. (Allofich)
   - Added stub IBM ROM BASIC points in the BIOS area
     so that when MS-DOS 1.x and 2.x BASIC.COM and BASICA.COM
     are run, a polite message is displayed instead of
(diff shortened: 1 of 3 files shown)


Allofich:
I accidentally added a log file to this commit. Sorry, I will remove it in my next PR.

PR #3037, opened 2021-11-04
Make 8086 treat 0x60-0x6F as 0x70-0x7F

Allofich:
Again based on https://www.pcjs.org/documents/manuals/intel/8086/.

> Duplicate Conditional Jumps (0x60-0x6F)
> Opcodes 0x60 through 0x6F decode identically to the conditional jump opcodes at 0x70 through 0x7F, respectively. This 
> is not true for any other x86 CPU.

Currently, 8086 treats these as illegal opcodes.

Also tried to speed up emulation by only running checks in these opcodes for relevant cputypes.

Also removed an extraneous file from my last PR.


joncampbell123:
Hey, yeah, I read about those on a blog about old DOS stuff who's name I can't remember right now.

I think he also found a few other simple aliases as well.

Allofich:
https://www.pcjs.org/documents/manuals/intel/8086/ says it got some of its information from there.

Allofich:
Just realized I forgot to update the CHANGELOG. It's so easy to let something slip. Well, I'll put that in my next PR.

joncampbell123:
Certainly. I added comments to the aliases to better explain why they are there for the 8086 cputype.

Allofich:
I just noticed I made a mistake in opcode 0x63. The 286 cputype will not have any code for it. The way it was before was 

`(CPU_ArchitectureType<CPU_ARCHTYPE_286) goto illegal_opcode;`

joncampbell123:
The 286 normal and prefetch cores are ALSO used for 80186 emulation.

8086 normal and prefetch cores are exclusively 8086.

see also src/cpu/cpu.cpp line 3558.

Allofich:
Right, I made a mistake in this PR with opcode 0x63, though. I'll try to fix it.

PR #3054, opened 2021-11-08
Fixed a problem with  box-drawing characters in JEGA/AX and TTF.

nanshiki:
Respond to bug reports in the comments of #2977.

**Additional information**
The code for box-drawing characters in AX is different from DOS/V, so it has been modified to get them from the code table for AX.


Wengier:
@nanshiki It fixed the issue with box-drawing characters. Thanks.

image

Issue #2835, opened 2021-08-22
Floppy and CD-ROM drives I/O too fast


MX9000:
When I install a software from a floppy/CD-ROM image or type DIR or COPY commands at the DOS prompt, the I/O on these drives seems too fast to me and affected by the cycles of the emulated CPU.
In the dosbox-x.conf file I have the "hard drive data rate limit" setting set to -1, but I think it would be useful to add a separate setting for the floppy disk controller that is more faithful to the original speeds (https://en.wikipedia.org/wiki/Floppy-disk_controller).
Another setting could simulate the transfer rates of CD-ROM drives 1x and higher (https://en.wikipedia.org/wiki/CD-ROM#Original_speed).
Thank you.

Allofich:
There is also an issue about hard drive I/O speed. Might be useful to have the two issues linked so I'll do that here.
https://github.com/joncampbell123/dosbox-x/issues/2720

MX9000:
@Allofich Thank you.
I was able to find issue #2039 but I missed the one you mentioned.

Siddhant-Mashalkar:
if i disable Hard Drive Data Limit, dosbox-x freezes some times when there are too many files to copy, see #2751

MX9000:
The installer of the game MegaRace has a CD-ROM speed detection routine because the game requires at least a 2x drive.
If I start the installation with the option "hard drive data rate limit = 0" the routine fails to measure the speed of the drive:

![install_001](https://user-images.githubusercontent.com/58113425/150108873-b80376a8-e77e-4531-b330-34f40759562d.png)

However, if I start the installation with the option "hard drive data rate limit = -1" the routine complains that the transfer rate is too low.

![install_000](https://user-images.githubusercontent.com/58113425/150108955-d9e838f6-41c5-491b-ad60-d81fa5a3dbe9.png)

In both cases it is still possible to install the game ignoring the warning.

rderooy:
0 and -1 are not the only values you can specify. You should be able to set the data transfer rate you want in bytes/sec.

So for a double-speed CD-ROM it should be 307,200 bits/sec. So divide by 8 and the value should be 38400 bytes/sec.
https://www.computerhope.com/issues/ch000218.htm

MX9000:
@rderooy Thank you very much!
I didn't realize there was the opportunity of such a precise configuration: maybe it could be useful to add this explanation to the comments in the configuration file.

grapeli:
[...]

PR #3060, opened 2021-11-09
Make floppy disk I/O configurable

Allofich:

https://github.com/joncampbell123/dosbox-x/issues/2835 (partially)

I've been testing old PC booter games lately, and I've found that some of them have their title screens fly by very quickly due to the floppy drive emulation being too fast.

Probably the most extreme example I've seen of this is "Tapper" (RGB version).

https://user-images.githubusercontent.com/19624336/140979897-3080e22b-381f-45c2-9f57-51a922d9af17.mp4

This PR is a simple hack to lower the floppy drive speed to something more reasonable so that title screens like these can be seen. It's only applied when the floppy is accessed through Int 13 and the new speed is not based on a real drive's speed. Floppy drive speeds cited by people online make for a (by modern standards) very slow load that is probably more accurate but also probably much longer than most of us want to wait. Instead the speed I chose here is a rough equivalent of the speed I saw when loading Tapper in MAME. It's still about 420x slower than before.

https://user-images.githubusercontent.com/19624336/140981697-7b6b28b1-0109-477e-9685-9287a67119fb.mp4

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -68,6 +68,10 @@
   - Make memory B0000-B7FFF unmapped for the CGA
     emulation. Fixes "Backgammon 5.0" detecting that
     an MDA is also present when using CGA. (Allofich)
+  - Make floppy I/O delay separataly configurable from
+    hard drive delay, and make it slower than before
+    by default.(Allofich)
+  - Fixed possible crash with printing. (jamesbond3142)
   - Video emulation for PC-98 mode (for 400-line modes)
     is now 16:10 instead of 4:3. 480-line PC-98 modes
     are still 4:3. (joncampbell123)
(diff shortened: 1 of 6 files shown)


Allofich:
I checked again and it's actually slower in this PR than in MAME.

Allofich:
I just realized this PR could be a nuisance to people installing MS-DOS from floppy images, so I'll close it for now. It probably needs to be configurable before imposing any slowdowns on the floppy access speed.

Allofich:
Updated so that all the existing calls to `diskio_delay` are covered.
Added a `floppy drive data rate limit` setting in the .conf file.

This PR assumes `DOS_GetDefaultDrive() < 2` and `drivenum < 2` are reliable for detecting that a floppy drive is being used. From a little testing these conditions seem to work.

If we add separate I/O delay for CD-ROM emulation (not covered by this PR), we'll need some way to detect that calls are being made on a CD-ROM drive.

PR #3066, opened 2021-11-11
Don't try to read sector for INT 13 with cl=0

Allofich:
The PC booter game "Backgammon 5.0" creates this message in the log when run:

`Attempt to read invalid sector in Read_AbsoluteSector for sector 4294967295`

I think this message is misleading. What happens is Backgammon 5.0 calls INT 13 AH=2 (read sectors into memory) with CL=0. CL is the sector number, but valid values are 1 through 63.
http://www.ctyme.com/intr/rb-0607.htm
(Or, may only be 1 through 17, based on this site: https://stanislavs.org/helppc/int_13-2.html)

DOSBox-X gets the sector as
`( (cylinder * heads + head) * sectors ) + sector - 1L;` which for this call evaluates to 0 - 1, wrapping around to 4294967295. It tries to read this sector, which is beyond the size of the floppy and makes the error occur. The invalid sector called by the program, though, was 0, not 4294967295. This PR just catches this situation a moment earlier so the log message is more accurate.

Checking real BIOS behavior in MAME, it also returns an error of 4 in AH on this call and sets the carry flag, as DOSBox-X does.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -52,6 +52,8 @@
     of 8, rather than reading it from address 485h.
     Fixes corrupt graphics in the PC Booter version
     of Apple Panic. (Allofich)
+  - Fixed underflow when INT 13 AH=2 called for
+    sector 0. (Allofich)
   - Fixed possible crash with printing. (jamesbond3142)
   - Fixed possible freeze when shutting down Windows 9x
     after changing a CD image from the menu. (Wengier)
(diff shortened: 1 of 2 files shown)


Allofich:
I found another minor issue. Some games will call INT 13 with an invalid drive specified in DL. In such cases `GetDosDriveNumber` in bios_disk.cpp would always translate them to 0x7f, or 127. This would then lead to the misleading error message

`ERROR BIOS:Disk 127 non-existant`

This PR changes the error to use the value that actual value that was called for. In the game I am testing the value is 20, so now that will be given by the error message instead of 127.

PR #3081, opened 2021-11-16
Re-apply core_dyn_x86 commits lost when that core was removed and restored

Allofich:
I went through the history of the files in the core_dyn_x86 folder and found the remaining commits that were missing from the current code following its removal and restoration. This recreates the changes of those missing commits.


Wengier:
@Allofich The code did not compile with MinGW due to removed dyn_interrupt() function in the dynamic_x86 code. I tried to add it back so that it can compile again with MinGW.

Issue #2617, opened 2021-06-25
Irregular SJIS code handling in PC98


tomyun:
Just got curious how irregular (undefined) SJIS code should be handled.

[ir.txt](https://github.com/joncampbell123/dosbox-x/files/6713674/ir.txt)

```
$ xxd ir.txt
00000000: 837e 837f 8380                           .~....
```

This text file has `837f` which is not defined (shouldn't be used because of del 7f) in SJIS code map.

However, in Anex86, it seems to be mapped to a character next to it, i.e. `8380` (ム).

![anex86](https://user-images.githubusercontent.com/906626/123372851-2ffe9e00-d539-11eb-999e-3a9108358c00.png)

In DOSBox-X, it doesn't.

<img width="752" alt="dosbox-x" src="https://user-images.githubusercontent.com/906626/123372830-25440900-d539-11eb-8e16-b56d579628a5.png">

I guess DOSBox-X is right here, but just want to double check if this is what actual hardware would behave. I'm asking this since there are some translated games somehow relying on such irregular SJIS codes for custom font mapping and they just run fine on Anex86. Then I'm afraid they would run into a problem on DOSBox-X due to this subtle difference.

Wengier:
I also assume the DOSBox-X behavior is correct, since it does not sound right to map a undefined character to the character next to it. But you (or anyone else) please feel free to check the behavior on actual hardware so we will be sure.

maron2000:
Hi, I think it is a bug of MS-DOS. 
The result of guest NEC MS-DOS 6.2 on DOSBox-X is shown below.
Also got the same result using NP21/w.

![shiftjis_bug](https://user-images.githubusercontent.com/68574602/123409011-ab6b5880-d5e8-11eb-895e-1e801226f212.png)
[ir.txt](https://github.com/joncampbell123/dosbox-x/files/6715439/ir.txt)

tomyun:
Thanks for the info! Now I'm curious how many games/programs took advantage of this "undefined" behavior as most, if not all, of them were probably running on MS-DOS back in the days.

Wengier:
Thanks @maron2000 for confirming the actual behavior on Japanese MS-DOS! I have added a config option `del` in [dosv] section (defaults to `true`) which will map the said symbol (0x7f) to the next character (0x80) for all Japanese modes (Japanese DOS/V, TTF, JEGA, PC-98) supported by DOSBox-X. Please check out the updated Windows builds:

[...]

PR #3087, opened 2021-11-18
DCGA (640x400 2colors) and J-3100 are now supported.

nanshiki:
# Description
DCGA (640x400 2colors) is now supported.

To switch to DCGA video mode, set video mode 0x40 or 0x74. This can be used for VGA settings.
J-3100 emulation has also been added. [dosv] dosv=jp and j3100=true will boot in J-3100 mode, and Japanese will be available on the DCGA screen.

**Additional information**
DCGA is the video mode used by the Olivetti M24, AT&T 6300 and Toshiba T3100.
The J-3100 is the Japanese version of the Toshiba T3100, which was popular as a Japanese-capable AT-compatible machine before DOS/V came out.


Wengier:
@nanshiki Glad to see DCGA and J-3100 support added. I have some questions though:

1. Should a machine type like "dcga" be added for DCGA mode so that it will be entered automatically at startup?
2. Exactly when will the Japanese text be available on the DCGA screen? Only when dosv=jp *and* j3100=true? I added a DCGA command in Z:\TEXTUTIL so that you can enter video mode 0x74 from e.g. a VGA mode, but no Japanese text is displayed even with dosv=jp. Is this expected?
3. I see there are now two new functions for ShiftJIS conversions, namely jis2shift and shift2jis in int_dosv.cpp. I wonder if they correctly handle the 0x7f character with the "del" config option according to the "del_flag" variable (it was added due to #2617)?

Thanks for adding the support.

nanshiki:
@Wengier 
1. If machine = dcga is added, it will be CGA hardware + DCGA, but it was difficult to support DCGA screen display based on CGA.
Therefore, DCGA rendering based on CGA emulation of VGA used in DOSVAXJ3 is adopted.

2. If dosv = jp and j3100 = true, set the video mode to 0x74 and it will be displayed in Japanese.
I tried to run DCGA command after running VGA command, but it seems that Japanese can be displayed.

3. shift2jis and jis2shift of int_dosv.cpp are used only by J-3100's original function int 60h.
I actually converted 0x837f of ShiftJIS to JIS code using int 60h, but it became the same value as 0x8380. Similarly, 0x8380 data can be obtained for the acquisition of Kanji fonts.
These are the same operations on the actual machine, so I think that you can leave it as it is regardless of the del_flag setting.

Wengier:
@nanshiki It is certainly right that from the DOSVAXJ3 code `machine= dcga` is based on CGA emulation of VGA. As you said it is difficult to support DCGA directly from CGA, I think perhaps something like `machine=dcga_vga` can be made for the time being.

There is no doubt Japanese text can be displayed with both `dosv=jp` and `j3100=true`, even after DCGA command. However, it is not the case with only `dosv=jp` (but not `j3100=true`). I have attached a screenshot showing running DCGA command with `dosv=jp` and `j3100=false`:

image

I am thinking that with the addition of `machine=dcga_vga` the option "j3100" may not be really needed any more. That is, by setting both  `machine=dcga_vga` and `dosv=jp`, DOSBox-X will automatically enter into DCGA's J-3100 mode with Japanese support. On the other hand, if there is no `dosv=jp`, then with `machine=dcga_vga` only DOSBox-X will enter DCGA mode without Japanese support. Perhaps this approach is easier?

nanshiki:
@Wengier 
When dosv=jp and j3100=false, Japanese cannot be displayed in video mode 0x74.
When j3100=false, J-3100 specific functions, EMS frame address, etc. are not set, and even if Japanese can be displayed in 0x74, applications for the J-3100 will not work.
This can cause a lot of confusion, so we have chosen to display in English.

If I set machine=dcga_vga, does it set dos.set_j3100_enabled=true?
If so, I don't think the j3100 option is necessary.
dos.set_j3100_enabled is referenced by the IS_J3100 macro, and if IS_J3100=true, various J-3100-related settings are also made.
On the DCGA screen, Japanese is displayed only if IS_J3100=true and 0040:0049 is set to 0x74.
Also, if dos.set_j3100_enabled=true, then set_jdosv_enabled=true is required.

nanshiki:
@Wengier I consider the following.
If machine=dcga_vga and dosv=jp, set dos.set_j3100_enabled=true and set_jdosv_enabled=true.
The settings for the J-3100 will be made, and the application for the J-3100 will run in video mode 0x74 and display Japanese. Video mode 0x40 will display in English.
If machine=dcga_vga and dosv!=jp, video mode 0x40 and 0x74 will be displayed in English, and the J-3100 settings will not be made and the J-3100 application will not run.
In case of machine!=dcga_vga, video mode 0x40,0x74 cannot be set.

Wengier:
@nanshiki I believe it is a good proposal, although I think that video mode 0x40/0x74 may be set as long as machine is set to a VGA machine type (as it is now), so that you can manually enter it with DCGA command (the setting `machine=dcga_vga` just causes video mode 0x74 to be entered when DOSBox-X starts) . For the "j3100" option in [dosv] section, I think it can be changed from a true/false value to a value that specifies the J-3100 machine type (similar to the option in DOSVAXJ3), and unless it is set to "off" you can change to DCGA mode with J-3100 Japanese support with video mode 0x74 as long as you have `dosv=jp` and machine is set to a VGA machine type. If either `dosv=jp` is not set or "j3100" is set to "off", then DCGA command or `machine=dcga_vga` will enter DCGA mode without J-3100 Japanese support. In summary:

1. `machine=dcga_vga` and `dosv=jp` and `j3100!=off`: Start DOSBox-X with J-3100 DCGA mode
2. `machine=dcga_vga` and (`dosv!=jp` or `j3100=off`): Start DOSBox-X with English DCGA mode
3. VGA machine type and `dosv=jp` and `j3100!=off`: Start DOSBox-X with Japanese DOS/V mode, and J-3100 DCGA mode can be entered with DCGA command (video mode 0x74)
4. VGA machine type and (`dosv!=jp` or `j3100=off`): Start DOSBox-X with a non-Japanese mode, and English DCGA mode can be entered with DCGA command (video mode 0x74)
5. Non-VGA machine type: Neither Japanese DOS/V nor video mode 0x40/0x74 can be entered; DCGA command will be unavailable

nanshiki:
@Wengier 
On the J-3100, int 0x60, int 0x6f is set, the address of Kanji ROM is allocated from E000:0000, and the frame address of EMS must also be moved to D000:0000. This has to be initialized at boot time.
This is a problem for users who want to use DOS/V only, because it reduces the area of UMB and uses extra resources.

I think that the DOS / V function may be used when the user is set to use the J-3100. Both were available on the late VGA-equipped models of the J-3100.
However, if the user is set to use DOS / V, it is better not to be able to switch to J-3100 mode. 
Users who use DOS/V are not likely to run J-3100 applications.
In the case of machine type VGA, ```dosv=jp``` and ```j3100=false```, the J-3100 will not be set at startup, which means that the J-3100 application will not work even though it can switch to DCGA screen in video mode 0x74.

The video mode for DCGA is 0x40 for Olivetti M24 and AT&T 6300, and 0x74 for Toshiba T3100.
From what I've seen on Youtube, the video mode for Olivetti M24 seems to be 0x03 text mode at startup, so I think it's better to start up in 0x03 text mode instead of 0x40/0x74 even if you use the English DCGA mode.
The English display in video mode 0x40 or 0x74 uses graphics VRAM to display text, like CGA's video mode 0x06, and does not have text VRAM.
Therefore, I think video modes 0x40 and 0x74 are not used in normal DOS applications, but only in applications that assume DCGA rendering.
The J-3100's video mode 0x74 has a virtual text VRAM like DOS/V and has a function to get the address, so there is no problem with 0x74 being the video mode normally used.

If you want to keep the j3100 option, you can use it the way it was originally created here. In other words, if j3100=true, the initial settings of the J-3100 will be made, and the system will shift to video mode 0x74 at startup, allowing the use of J-3100 applications. When it shifts to video mode 0x03, it will be in DOS/V Japanese mode.
If j-3100=false, the J-3100 will not be initially configured, it will shift to video mode 0x03 at startup, and the language will be displayed according to the dosv setting.

Or you can eliminate the j3100 option and do the following.

1. If ```machine=dcga_vga``` and ```dosv=jp```, the J-3100 is initially configured and starts up in video mode 0x74, Japanese display is possible, and J-3100 applications run.
2. If ```machine=dcga_vga``` and ```dosv!=jp```, it will start in video mode 0x03 and the language specified by dosv will be displayed. Video modes 0x40 and 0x74 shift to the DCGA screen, but are displayed in English. 
3. If the machine is another VGA and ```dosv=jp```, it will boot in DOS/V video mode 0x03 and display Japanese and run DOS/V applications. In the case of video mode 0x40 and 0x74, the display will be in English if it is possible to switch to the DCGA screen.

Wengier:
@nanshiki Thanks a lot for the detailed explanations. So as you said enabling J-3100 support reduces available UMB memory and uses extra resources, which is not always desired by DOS/V users. It makes sense indeed. I have been thinking about how to proceed, and after some considerations I think keeping the way it was originally created is probably desired, except that we can change the option "j3100" from a boolean value to a string value optionally specifying the J-3100 machine type, as it was done in DOSVAXJ3. So for example:

1. `dosv=jp` and `j3100=off`: Only enable Japanese DOS/V, without J-3100 support (same as the original `j3100=false` with `dosv=jp`).
2. `dosv=jp` and `J3100=on`: Enable both Japanese DOS/V and J-3100 support (same as the original `j3100=true` with `dosv=jp`).
3. `dosv=jp` and `j3100=gt/sgt/gx/gl/sl/sgx/ss/gs/sx/sxb/sxw/sxp/ez/zs/zx`: Enable both Japanese DOS/V and the specific type of J-3100 machine (similar to DOSVAXJ3).

(The proposed `machine=dcga_vga` type is special not only because it is based on VGA emulation, but it also appears to not support the TTF output, so I guess it may be better to not make it a machine type for now, but instead you can switch it in and out with DCGA and VGA commands respectively, both available in Z:\TEXTUTIL).

Please check it out.

Wengier:
@nanshiki I separated ``j3100type`` option from ``j3100`` so that you can set ``j3100=manual`` which enables the J-3100 machine type but not activates it at start. You can enter J-3100 emulation at any time with DCGA command afterwards in this case. On the other hand, if you set ``j3100=on`` or ``j3100=auto`` then the J-3100 emulation will be both enabled and activated at start, or if you set ``j3100=off`` then the J-3100 emulation will be disabled altogether. Meanwhile, the ``j3100type`` option is exclusively for setting the J-3100 machine type (gt, sgt, gx, ...).

nanshiki:
@Wengier Thank you for also porting the palette settings for LCD and plasma displays.
The j3100 related options are also good.
I also understand about dcga_vga.

Could you please fix the following?
The return value of bx for int 0x16, ah=0xf6 is the J-3100 machine code.
In bios_keyboard.cpp line 1410
```
            // J-3100 machine code(dummy)
            reg_bx = 0x6a74;
```
->
```
            // J-3100 machine code
            reg_bx = J3_GetMachineCode();
```

Wengier:
@nanshiki I have applied the said fix. Thanks! The J-3100 palette code I ported was based on your work in DOSVAXJ3, which works very well for these machine types. You have done a great job in both DOSVAXJ3 and in importing code in this PR. Thanks again!

nanshiki:
@Wengier Thanks for the fix.
One more point, when j3100type=default, j3_text_color is not set and the background color and text color are the same, so the text is not visible.
It would be better to call J3_SetType() even when j3100type=default, and internally check if it is default or not.

Wengier:
@nanshiki Yes, I noticed the said issue too, and I fixed it right before you reported it, although I fixed it in a different way, by returning 0x6a74 for default in bios_keyboard.cpp instead of int_dosv.cpp, since the function J3_GetMachineCode() is also called elsewhere. Please try the new code.

By the way, there is a SETCOLOR command in DOSBox-X which can be used to change the text color palette including DCGA mode (e.g. `SETCOLOR 0 #123456` and `SETCOLOR 7 #abcdef`), so I did not add the j3backcolor/j3textcolor options as in DOSVAXJ3.

Wengier:
@nanshiki I have made a commit (https://github.com/joncampbell123/dosbox-x/commit/a92106d37fbda8441b62ac3a4479a754b3865bf5) which highlights the text when selecting text for the clipboard in J-3100 mode (`dosv=jp` and `j3100=on`). The text highlighting however does not work in pure DCGA mode (i.e. without J-3100). Do you know if there is any way to highlight the text in the pure DCGA mode as well? Thanks!

nanshiki:
@Wengier If the J-3100 mode is enabled, the character code is in the virtual character VRAM, but if the J-3100 mode is not enabled, the virtual character VRAM does not exist.
Characters are displayed in graphics just like CGA in video mode 0x06, but character codes and attributes are not saved.
I don't think highlighting is currently working in CGA in video mode 0x06 either. As for copying to the clipboard, I think it is possible to get the character code from the graphic data of the font as well as the CGA.

Wengier:
@nanshiki I tried the SimCity for J-3100 downloaded from the following URL, but it does not appear to run in the current J-3100 mode (with `dosv=jp` and `j3100=on`). SimCity will start if booted from the disk image, but neither TITLE.EXE nor SIMCITY.EXE will start properly if run directly from the J-3100 mode as implemented. Could you please check what happened?

https://vetusware.com/download/SimCity%20for%20J3100%20JA/?id=12738

I tried both simcity120.zip and simcity120.zip. Same result.

nanshiki:
@Wengier
int10.cpp
```cpp
INT10_Handler()
{
  switch(reg_ah) {

  case 0x82:

      } else if(reg_al == 0x05) {
         reg_bl = 0x01;
      }
```
Comment out reg_bl=0x01 or delete the whole reg_al == 0x05 process.
If int10h ax=0x8205 and bl=0x01 is returned, it means the color display can be used, but SIMCITY seems to use the color video mode 0x75 when color is available.
Video mode 0x75 is 640x400 16-color, but the details are unknown and DOSVAXJ3 does not support it. Even the J-3100 does not seem to support it on most models.
If this is commented out, the setup.exe for CGA/EGA on Toshiba Windows 3.1 will not start, but the setup.exe for VGA version will work, so please deal with this for the time being.

Wengier:
@nanshiki Thanks for the information. I can confirm that SimCity runs in B/W mode if the said line is commented out. I have added a new config option "j3100colorscroll" (in [dosv] section) accordingly which defaults to false but allows the user to toggle this setting. The commit is here: https://github.com/joncampbell123/dosbox-x/commit/f640e69219c45993f404d697f4846eee331fb903

Thanks!

PR #3094, opened 2021-11-21
Disable segment limits check for 8086/186

### LINKED ISSUE #3092: The "segment limits" option should perhaps default to false for at least the 8086 CPU

### Have you checked that no other similar question(s) already exists?

### Question

The PC Booter game "Karnov" will hang on starting gameplay if using the 8086 cpu with the message

```
Limit check ffff+2-1 = 10000 > ffff DS:SI
Segment limit violation
```

filling the log.

Setting "segment limits = false" in the .conf file allows the game to run with the 8086 emulation, seemingly without problems.

Before the point where the hang occurs, the game code appears to check whether the CPU is an 8086 or not, and runs different code according to the result.

cpu=8086_prefetch. Hangs with "segment limits = true".
```
6B9B:00000011  cmp  word [CD24],0186           ds:[CD24]=0086         EAX:0000DC45 EBX:00000000 ECX:00000000 EDX:00000000 ESI:0000FFFF EDI:0000D970 EBP:000017C6 ESP:000017C6 DS:8239 ES:8239 FS:0000 GS:0000 SS:8FD0 CF:0 ZF:1 SF:0 OF:0 AF:0 PF:1 IF:1
6B9B:00000017  jc   0000001C ($+3)             (down)                 EAX:0000DC45 EBX:00000000 ECX:00000000 EDX:00000000 ESI:0000FFFF EDI:0000D970 EBP:000017C6 ESP:000017C6 DS:8239 ES:8239 FS:0000 GS:0000 SS:8FD0 CF:1 ZF:0 SF:1 OF:0 AF:0 PF:1 IF:1
```

cpu=286_prefetch. Doesn't hang with "segment limits = true".
```
6B9B:00000011  cmp  word [CD24],0186           ds:[CD24]=0286         EAX:0000DC45 EBX:00000000 ECX:00000000 EDX:00000000 ESI:0000FFFF EDI:0000D970 EBP:000017C6 ESP:000017C6 DS:8239 ES:8239 FS:0000 GS:0000 SS:8FD0 CF:0 ZF:1 SF:0 OF:0 AF:0 PF:1 IF:1
6B9B:00000017  jc   0000001C ($+3)             (no jmp)               EAX:0000DC45 EBX:00000000 ECX:00000000 EDX:00000000 ESI:0000FFFF EDI:0000D970 EBP:000017C6 ESP:000017C6 DS:8239 ES:8239 FS:0000 GS:0000 SS:8FD0 CF:0 ZF:0 SF:0 OF:0 AF:0 PF:1 IF:1
```

Should "segment limits" perhaps default to false for at least the 8086 CPU?

Allofich:
Disables the segment limits check for the 8086 and 186 cores.

Fixes #3092.

Note that although the suggested fix was to compile out the checks, I decided to do it this way for a few reasons.
- It's just a one line change this way
- If we're going to exclude the 186 from using the segment limits, it's awkward to check for it (since it uses CPU_ARCHTYPE_286) with #if statements. (I couldn't find information on whether the 186 raises an exception on exceeding the segment limit or not)
- Also compiling out those checks would potentially be beneficial for the 8086/186 core performance, but I don't think those cores are where performance is much of a concern.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -127,6 +127,9 @@
   - Always return bit 6 of port 61h as set for PC and as
     clear for PCjr and Tandy. Allows Zaxxon to load in
     PC mode. (Allofich)
+  - Disable segment limit check for 8086/186. Fixes
+    "Karnov" and "Guerilla Warfare" when run with
+    8086 CPU emulation. (Allofich)
   - Video emulation for PC-98 mode (for 400-line modes)
     is now 16:10 instead of 4:3. 480-line PC-98 modes
     are still 4:3. (joncampbell123)
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -3295,7 +3295,8 @@ public:
 		reg_ebp=0;
 		reg_esp=0;
 
-		do_seg_limits = section->Get_bool("segment limits");
+        if (CPU_ArchitectureType >= CPU_ARCHTYPE_286)
+            do_seg_limits = section->Get_bool("segment limits");
 	
 		SegSet16(cs,0); Segs.limit[cs] = do_seg_limits ? 0xFFFF : ((PhysPt)(~0UL)); Segs.expanddown[cs] = false;
 		SegSet16(ds,0); Segs.limit[ds] = do_seg_limits ? 0xFFFF : ((PhysPt)(~0UL)); Segs.expanddown[ds] = false;


joncampbell123:
@Allofich That's not going to be 100% effective. For one thing if the user changes the CPU type on the fly it means it will either load the setting or use whatever the variable happened to be at the time. I'll make the proper change to disable segment limits below 286.

I'm fairly certain the 80186 does not have segment limit exceptions because that wouldn't make any sense prior to the introduction of 286 protected mode.

Allofich:
> For one thing if the user changes the CPU type on the fly it means it will either load the setting or use whatever the variable happened to be at the time.

I didn't think about that, because I've never used that feature.

PR #3095, opened 2021-11-21
Don't run mouse setup code if there is no mouse.

Allofich:
I was testing the PC Booter game "Trilogy", which would hang almost immediately. I found out that the hang occurred if the mouse was moved (the game writes over the vector table, including the address for int 74, which is involved with the ps/2 mouse. Moving the mouse would send execution into the middle of bogus code because int 74's address was overwritten.) Probably the game is designed for an original IBM PC with no mouse.

The hang can be avoided by setting `int33` to false in the .conf file. However, if `int33` is left as true, the hang will happen even if you have no serial mouse or PS/2 mouse specified in the .conf file.

I found that `MOUSE_Startup`, which installs int 33 handling, had a TODO for doing nothing if there is no serial mouse or PS/2 mouse emulated. This PR implements that TODO, which also allows the problem with Trilogy to be avoided by specifying no mouse in the .conf file, which I think is better than setting `int33` to false, since int 33h is a DOS interrupt and shouldn't be relevant when using a PC booter game.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -145,6 +145,8 @@
     the remaining block of undefined/undocumented EGC
     ROPs. This fixes "Atomic Punker" which uses EGC ROP
     0xBE for sprite rendering. (joncampbell123)
+  - Don't run mouse setup code if there is no emulated
+    mouse. (Allofich)
   - Integrated commits from mainline (Allofich)
     - Add proper opl3 handling of the waveform select
     to dbopl.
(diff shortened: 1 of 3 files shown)


joncampbell123:
I'm OK with this, even though on real hardware, BIOSes tend to leave other IRQs it doesn't use with NULL interrupt vectors. One Pentium MMX 133MHz system I test with tends to leave IRQ 5, 10, and 11 with NULL interrupt vectors with those IRQs masked which is why the DOSLIB Sound Blaster code checks the previous vector to decide whether the interrupt handler chains to the previous or not.

Issue #3096, opened 2021-11-21
Zool hangs for extended periods


rderooy:
### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

### Have you checked that no other similar bug report(s) already exists?

- [X] I have searched and didn't find any similar issues.

### What operating system(s) this bug have occurred on?

Linux F35

### What version(s) of DOSBox-X have this bug?

0.83.20git

### Describe the bug

This seems to be specific to DOSBox-X. Starting the game "Zool", it seems to hang on startup, but eventually continues. There is one error in the log at this point:
```
LOG:   21487228 ERROR EXEC:stack underflow/wrap at EXEC SS:SP=0000:0000
```

Proceeding to start the actual gameplay, again seems to hang, but after a long wait it continues. The log shows these errors:
```
LOG:   90155528 ERROR DOSMISC:End of cluster chain reached, but maybe good afterall ?
LOG:   90826479 ERROR DOSMISC:End of cluster chain reached, but maybe good afterall ?
LOG:   91229544 ERROR DOSMISC:End of cluster chain reached, but maybe good afterall ?
LOG:  302255032 ERROR DOSMISC:End of cluster chain reached, but maybe good afterall ?
LOG:  303048204 ERROR DOSMISC:End of cluster chain reached, but maybe good afterall ?
```

Launching the game from dosbox-svn or dosbox-staging, and it starts straight away.

### Steps to reproduce the behaviour

The game was started as follows:
```
dosbox-x -defaultconf -c "imgmount a Zool.img" -c a: -c Zool
```

### Additional context

Zool2 also seems to suffer from slowdowns on startup.

rderooy:
It seems the above issue only happens if you run the game directly from the 720KB floppy disk image. If you copy the contents to the (folder mounted) HDD, it runs fine.

Allofich:
I recently added separate I/O delay for floppies, which defaults to slower than hard drive/CD delay. Does setting "floppy drive data rate limit" under the [dos] section in the .conf file to 0 fix the hangs?

rderooy:
I will need to check when I get back home. I had seen that option, but thought it was disabled by default.

Allofich:
By default it is -1, which is supposed to give a reasonable delay (same as "hard drive data rate limit").

I set that default delay to 5000 bytes/second, in order to display PC booter startup screens for a few seconds instead of having them instantly disappear.

[...]

PR #3098, opened 2021-11-22
Raise default floppy speed

Allofich:

May be related to #3096.

According to an answer on https://stackoverflow.com/questions/52841124/how-fast-could-you-read-write-to-floppy-disks-both-3-1-4-and-5-1-2, floppy disk speeds "maxed out around 100-250kbps and I've never seen above 250kbps on a floppy."

This raises the default floppy speed from 5000 Bps (which I chose because it allowed some PC booter game screens to show for a few seconds) to 175 kbps, taking the average of 100 and 250.

Testing with the PC booter game "Tapper", the opening screens are still visible (though they go by faster than with 5000 of course) with the new value.

### DIFF
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -3687,7 +3687,7 @@ public:
                 ::disk_data_rate = 3500000; /* Probably an average IDE data rate for early 1990s ISA IDE controllers in PIO mode */
         }
         if(::floppy_data_rate < 0) {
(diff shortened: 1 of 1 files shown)


Allofich:
As I'm pretty sure this is related to #3096, I'll make this draft status until there is a bit more testing there and this is confirmed as a good default value, or a better one is chosen.

Allofich:
Looks like #3096 is also due to something else.

Anyway this should be a better default than the current one.

PR #3136, opened 2021-12-12
Add bash-completion script

rderooy:
Partly based on the work done by @FeignInt for dosbox-staging

With this installed, you will have bash-completion for DOSBox-X.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -341,7 +341,7 @@ contrib/linux/com.dosbox_x.DOSBox-X.metainfo.xml
 *.a
 *.la
 *.deps
-dosbox-x
+src/dosbox-x
 Makefile
 
 doxygen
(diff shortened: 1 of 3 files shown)


Wengier:
I think this is the bash-completion function from outside of DOSBox-X to launch DOSBox-X. On the other hand, DOSBox-X already supports bash-completion from its internal DOS shell by pressing the Ctrl+Tab key.

rderooy:
Indeed this is bash-completion for Linux (and perhaps macOS?) host systems.

I would not call the completion inside DOSBox-X bash-completion as there is no bash shell inside DOSBox-X.

Wengier:
@rderooy Yes, there is no bash shell inside DOSBox-X, but the function of **bash-style completion** is supported by DOSBox-X's internal DOS shell, in addition to the standard DOSKEY-style completion. In other words, the built-in shell supports both styles of file completions.

An example:

image

PR #3149, opened 2021-12-17
(WIP) Dosbox remote debugging server via TCP

yetmorecode:
Adds remote debugging via TCP functionality to DOSBox-X

None

Introduces a "debug server". Running in a seperate thread, waiting for incoming connections via TCP.

I'm also working on a Ghidra dosbox agent to connect to dosbox via this means. So far, dosbox does not provide any means to remotely control the debugger.

No, it's scoped by C_DEBUG_SERVER. Though there is work to do on the integration into the build system.

## Additional information

It's only proof of concept for WIN32. But I thought I might share this early on.

Second I'd like to thank you guys for providing DOSBox-X. Without further ado, it's a great project to build upon.

I couldn't resists to move the inlined class in debug.cpp to own files and do some some minor cleanups in debug/ in general, so that's in the PR as well :/

The protocol is fully text-based and implemented in debug_server.cpp. I didn't write any documentation for it yet though.

I will keep working on other platforms, support for (eventually) all debugger commands and hopefully some provided feedback regarding things I didn't understand correctly.

Yet what are the general odds of getting this eventually accepted into DOSBox-X?


joncampbell123:
I like the idea as long as you don't write it in a Windows-specific (Winsock) way.

Stick with the functions that both Winsock and Linux have in common (Berkeley sockets) and you should have some good cross-platform code there. If it works for Linux it would likely also work on Mac OS X as well.

It should be OFF by default, obviously, so as not to scare the Windows firewall.

Wengier:
This function is certainly useful, but it will be even more useful if it is cross-platform rather than Windows only. Please keep up the work.

yetmorecode:
@Wengier @joncampbell123 I'm sorry if I wasn't clear in the beginning. I will of course add support for Linux/macOS once I finish the general workings of the protocol handling. It's quite some way to go though in general (I got stuck in some weird rabbithole the last time). 

Also, thanks for the advice regarding netwoking APIs. Much appreciated, it's quite confusing what is what with all those standards tbh, but I will try to work on it.

yetmorecode:
I also realized by now that there is way more C++ code in the codebase than I initially thought. So I suppose it's better to put DebugServer in a class? I would certainly prefer this.

Wengier:
@yetmorecode I think you can implement the function with a structure you see fit, as long as the code eventually works for different platforms. Thanks for it.

joncampbell123:
@yetmorecode This work in progress is going to bitrot if not maintained against master branch, just so you know.

yetmorecode:
I will have to rebase that all once I get to it. I ran into some rabbithole with the protected mode debugger and unfortunately didn't find time to look at it so far. I think the PR can be closed for now. I will bring that back later when it's in better state (hopefully)

Issue #3179, opened 2021-12-27
Planned Pentium III support?


PythonBlue:
### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

### Have you checked that no other similar feature request(s) already exists?

- [X] I have searched and didn't find any similar issues.

### What you want

First of all, I must say that, having found this emulator recently, I'm impressed with the open source effort. Even managed to find a way to get MIDI input working via an internet connection courtesy of the pcap library, though I'll delve into that another time unless asked.

That being said, one reason I took an interest was to emulate an old Windows 98 MIDI driver, which, judging from my additional experiments in the horribly sluggish UTM/QEmu, seem to require a Pentium III processor for some of the included sounds. I realize Pentium II support was only just rolled out, so I don't expect this to be worked on any time soon: just am curious if this is a planned feature.

### Describe alternatives you've considered

I realize this might be a fringe case, especially insofar as how I'm using DosBox-X. For that matter, the most responsive alternative I found using said MIDI driver, including support for the sounds of concern, was VMWare Fusion, an alternative I'd prefer to avoid long-term due to VMWare's apparent refusal to support Intel emulation on Apple Silicon.

### Additional context

Not that it should matter beyond the scope of emulated system requirements, but the MIDI program I'm trying to get fully functional is the VL Virtual Acoustic plugin for Yamaha's S-YXG100. Doubt it's a licensing issue, because I even went as far as copying related system files from my VMWare setup, with no luck.

joncampbell123:
Pentium II was fairly easy to add in that you take pentium_pro and add SYSENTER/SYSEXIT and update CPUID accordingly. Some fake MSRs are added because Windows ME assumes they are valid (else it will BSOD) and it even uses those to try to do a CPU microcode update (!) if it detects a Pentium II. SYSENTER/SYSEXIT was tested using the Linux kernel built to optionally support it.

Pentium III emulation will require the addition of SSE registers and instructions. Unlike MMX, the SSE registers are their own register file. This also means support for the newer CPU instructions that save and restore the SSE registers in addition to FPU state.

I'd like to see that, it's just that it needs some work to build up.

PythonBlue:
Like I said, I don't expect this idea to be implemented immediately. Good to know you're open to it, though, and also good to know the technical ease of enabling Pentium II relative to III.

joncampbell123:
[...]

PR #3188, opened 2021-12-30
Pentium III emulation

joncampbell123:

Pentium III emulation request #3179. This is not going to be quick or easy but it can be done incrementally.

Pentium III emulation, SSE registers, SSE instructions and control registers, SSE enable bit in CR4, and MSRs.


PythonBlue:
Having found this development branch, you might be happy to know I tested the code as it is now. That said, it appears it's currently not an option to boot with Pentium III set as the processor due to a CPU fault, at least that's the explanation I found when booting with a disk preinstalled with 98SE. This issue might also explain the still lacking results for my fringe use case, since I'm pretty sure that as a MIDI driver, the software of concern loads on boot.

joncampbell123:
@PythonBlue I haven't added any actual SSE instructions to normal core, much less the FXSAVE/FXRESTOR instructions that Windows 98SE will need to task switch with SSE enabled. But thanks for testing. :laughing:

PythonBlue:
> @PythonBlue I haven't added any actual SSE instructions to normal core yet, much less the FXSAVE/FXRESTOR instructions that Windows 98SE will need to task switch with SSE enabled. But thanks for testing. 😆
> 
> What I have done is set the framework to begin adding instructions, and I made sure the debugger can show the SSE instructions. So while the debugger recognizes one group of SSE instructions, normal core doesn't execute them yet.

Yeah, it took me having to see the description of this pull request while giving feedback to figure out that it's not even fully ready for testing yet. XD Thanks for the clarification. Needless to say, I am excited for this possibility when it's developed enough to be at least somewhat useable.

qeeg:
According to the Intel SDM, SSE alignment exceptions are literally just GPFs. HTH.

PythonBlue:
> According to the Intel SDM, SSE alignment exceptions are literally just GPFs. At least, it works for my emulator, PCBox. HTH.

I would consider PCBox as an alternative if it actually worked on a Mac. Alas, just built that from source due to no build for that platform being available, only to see that the menu bar necessary for changing settings beyond a 1981 IBM computer does not exist.

qeeg:
I mean, there is a Qt port of the UI being worked on by 86Box, it's just taking some time. Please be patient.

PythonBlue:
Just tried the latest commits out of excitement, and while it seems not all of the CPU opcodes may be implemented yet judging from the status of my use case, I'm happy to say that there is progress all the same insofar as testing: my Win98SE guest no longer crashes on startup when set to this new CPU type!

PythonBlue:
> I mean, there is a Qt port of the UI being worked on by 86Box, it's just taking some time. Please be patient.

I'm guessing from this that you did see my response to your initial message, and I apologize for how blunt my words were at the time. From what I hear, it seems to be a plausible solution for Windows users: I was just a bit frustrated that PCBox is the only branch I could find of 86Box that supports Mac at all, and even then, that it was still a WIP. Still, I realize it was wrong to be that impatient about it, especially since the reason I care so much about Pentium III emulation has to do with future proofing from Apple Silicon some possible tools: I can get by perfectly fine with my current setup on an Intel Mac for the time being.

joncampbell123:
> Just tried the latest commits out of excitement, and while it seems not all of the CPU opcodes may be implemented yet judging from the status of my use case, I'm happy to say that there is progress all the same insofar as testing: my Win98SE guest no longer crashes on startup when set to this new CPU type!

Yeah, and that's actually kind of weird, because I haven't yet added the instructions needed to save or restore SSE state (FXSAVE/FXRESTOR) and Windows ME boots just fine with it anyway (needing only an additional MSR to read "platform ID" for Pentium III?).

I suspect crashes will happen when you run anything that uses SSE.

PythonBlue:
> > Just tried the latest commits out of excitement, and while it seems not all of the CPU opcodes may be implemented yet judging from the status of my use case, I'm happy to say that there is progress all the same insofar as testing: my Win98SE guest no longer crashes on startup when set to this new CPU type!
> 
> Yeah, and that's actually kind of weird, because I haven't yet added the instructions needed to save or restore SSE state (FXSAVE/FXRESTOR) and Windows ME boots just fine with it anyway (needing only an additional MSR to read "platform ID" for Pentium III?).
> 
> I suspect crashes will happen when you run anything that uses SSE.

Good to know that you didn't implement those particular instructions yet. Curious why it boots fine now, that in mind...

And yeah, probably true, though I do know that the MIDI driver I hope to operate is running fine despite some of its sounds requiring Pentium III instructions.

joncampbell123:
Windows ME is leaving me confused here actually... it looks like it switched on bit 9 of CR4, because my SSE test program runs just fine in a DOS Box and as a 16-bit Windows 3.1 application (Win32s build crashes for some reason?) but nothing in the DOSBox log says anything about an invalid opcode.

Did Microsoft just retrofit the task switcher in Windows 98SE/ME with something to load/save SSE manually and just flat out ignore FXSAVE/FXRESTORE? I would have expected an Invalid Opcode exception by now trying to execute those instructions! For that matter, I also expected an invalid opcode exception for LDMXCSR and SDMXCSR (load/save MXCSR) since that's a pretty important register too! It's the SSE equivalent of the FPU control word.

joncampbell123:
On a related note, I just found out that Pentium II and (WIP) Pentium III emulation was NOT reporting MMX support through CPUID! Maybe that's what Windows ME is confused about. Checking...

joncampbell123:
While we're getting Pentium III emulation up, here's what my real Pentium III shows for CPUID.

There's clearly a lot there that DOSBox-X is not likely to implement for a long time, if ever. Especially support for > 4GB of RAM and PAE.

IMG_20220101_200423170 cpuid

joncampbell123:
Also on real hardware, the FreeDOS debugger (or the Pentium III?) will apparently skip over the following instruction if it encounters RDMSR with a debugger.

The three mystery values queried by Windows ME on boot up:
IMG_20220101_201039731 rdmsr1
IMG_20220101_201039731 rdmsr2

PythonBlue:
> While we're getting Pentium III emulation up, here's what my real Pentium III shows for CPUID.
> 
> There's clearly a lot there that DOSBox-X is not likely to implement for a long time, if ever. Especially support for > 4GB of RAM and PAE.
> 
> However my guess for stepping/model/family was right on!
> 
> IMG_20220101_200423170 cpuid

Understood about some things being unlikely to be implemented, though I doubt 4 GB RAM would be necessary for DosBox-X anyway given Windows 98SE, for instance, only supports up to 512 MB.

joncampbell123:
It's one thing to go up to just below 4GB because DOSBox-X inherits DOSBox SVN code using minimally 32-bit datatypes for physical memory addresses. That was easy, in fact, I still think DOSBox SVN's paranoid 63MB limit is still excessively low, but the devs refused to budge when I argued about it (and a bug in xms.cpp related to that).

I think the original paranoia is old DOS/Windows 3.1 code that was stupid enough to return extended memory size in kilobytes as a signed 16-bit integer, which breaks when you have 32MB (32768KB) or more of memory in your system.

However to go beyond 4GB might be tougher. Developers seem to use "Bitu" whenever they feel like it, which happens to be 64-bit on 64-bit, but 32-bit on 32-bit, and PhysPt elsewhere.

Also remember that in 1998 or so, 64MB to 128MB was considered a lot of memory back then. DOSBox-X aims to emulate up until the Windows XP era (when Microsoft forced motherboards to drop ISA slots through the PC "standards"), which is when the average machine might have had 128MB to 512MB.

joncampbell123:
Good news... Windows ME is deceitfully tolerant of such minimal SSE emulation but the Linux kernel is brutally honest about what it expects of a Pentium III, and that's *good*.

Right now I see it crash at startup attempting to execute FXSAVE/FXRESTOR.

joncampbell123:
The Linux kernel apparently will require PREFETCHNTA/PREFETCH0/etc. if it sees a Pentium III. Okay.

joncampbell123:
Hey, found an old copy of CPU-Z that RIGHT AWAY tries to use the MOVUPS instruction (and BSODs Windows doing it because I haven't added that one yet :laughing: ) so progress will be made with that one.

PythonBlue:
> Hey, found an old copy of CPU-Z that RIGHT AWAY tries to use the MOVUPS instruction (and BSODs Windows doing it because I haven't added that one yet 😆 ) so progress will be made with that one.

Yeah, noticed the BSOD issue with the current commit, myself. That said, however, there is sort of good news for my use case in that the BSOD also happened when I disabled the MIDI driver's support for DirectSound, not just during bootup, so obviously we're touching on the required instructions.

DSN9990:
@joncampbell123 Is PSE/PAE emulation really difficult to implement?

joncampbell123:
@DSN9990 I'm not certain, it depends on how hardcoded everything is towards the 386-style page structure lookup.

RNMB15:
I saw an MMX, SSE and SSE3 test program on twitter.

https://twitter.com/simk98l
https://drive.google.com/file/d/1DYTF9EY2w7BUK9xFuO8NR2xVY3Iq9ND-/view

PR #3274, opened 2022-02-11
fixes joncampbell123/dosbox-x#3175

halcyon00:
Makes volume name matching use the standard 8.3 codepath.  This seems to be expected behavior as the volume for Stellar7 floppies is of the form
"Stellar7 1", however the installer expects to be able to search for both "Stellar7??" as well as "Stellar71" implying that trimming SHOULD be happening on both the name as well as the extension before wildcard matching even for volume labels.

This adds an additional variable vol_name to construct the intended output and cached volume label, but does not involve this variable in any wildcard matching.

If there are any related issues, reference them here. Ex. #1234, Fixes #1234, Closes #1234, etc.

If yes, describe the feature(s) introduced.

## Additional information


halcyon00:
This change is incorrect. 

After looking into #3187 there's a bug in how DOS_MakeName assumes that it can follow filename rules even for volume labels, and it does an implicit trim to the volume's "filename" portion to remove spaces, and thus fails the comparison.

PR #3287, opened 2022-02-19
Change default value of mt32.rate option to 48kHz

### LINKED ISSUE #3277: MT-32 emulation is lower pitch compared to other versions of DOSBox

### Have you checked that no other similar bug report(s) already exists?

Windows 10

0.83.22 SDL1

### Describe the bug

I have found out that DOSBox-X seems to have some weirdness when it comes to audio pitch when emulating the MT-32 when compared to other versions of DOSBox. 

Included as well is a video where the first minute and a half is recorded using DOSBox-X and the other minute and a half is recorded using DOSBox ECE to let everyone hear how the Red Baron main title tune is supposed to sound like.

https://youtu.be/tyR-RiIEZN8

### Expected behavior

_No response_

1. Configure Red Baron to use MT32 in DOSBox-X
2. Audio is lower pitched
3. Configure Red Baron to use MT32 in DOSBox ECE
4. Audio sounds as expected

### Used configuration

_No response_

### Emulator log

_No response_

### Additional context

Slight edit as i just tried other games that can use the MT32 in DOSBox-X and compared them to how they sound in DOSBox ECE and DOSBox-X definitely has an issue with its MT32 emulation as Space Quest 3, Police Quest 3 and Ultima 6 all exhibits the same lower pitched MT32 audio that Red Baron has. 
But in DOSBox ECE they all sound as they should sound like.

maron2000:
The default value of mt32.rate option is 44.1kHz at present but the sampling rate defined by default value of mt32.analog option is 48Khz and therefore result in lower pitch sound than expected.
Therefore, this PR  fixes the default values of respective options to be consistent, and prints warning in log if usere set values which may result in unexpected pitches. 

Closes  #3277

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,7 +14,10 @@
     Pharkas when sbtype=sb16. (joncampbell123)
   - Avoid crashes due to assertion failure when media 
     keys are pressed (maron2000)
-
+  - Fix CD audio playback failure (cue + mp3) when
+    trying to play from pregap sectors. (maron2000)
+  - Change default value of mt32.rate to 48kHz to match
+    default value of mt32.analog option. (maron2000)
 0.83.22
   - Added Pentium 3 Processor Serial Number emulation.
     Serial number can be set from dosbox-x.conf or not
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -3085,7 +3085,7 @@ void DOSBOX_SetupConfigSections(void) {
     Pint->Set_values(mt32reverbLevels);
     Pint->Set_help("MT-32 reverb level");
 
-    Pint = secprop->Add_int("mt32.rate", Property::Changeable::WhenIdle, 44100);
+    Pint = secprop->Add_int("mt32.rate", Property::Changeable::WhenIdle, 48000);
     Pint->Set_values(rates);
     Pint->Set_help("Sample rate of MT-32 emulation.");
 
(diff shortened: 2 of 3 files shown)


rderooy:
Should this not match mixer.rate? because that defaults to 44.1kHz.

But all modern soundcards should be 48kHz, so setting mixer.rate to 48kHz could be a good idea.

maron2000:
@rderooy Thanks for your comment. 
Yes indeed, we can set mixer.rate to 48kHz as well, but as far as I tried, the pitch is okay even it is 44.1kHz.
So while I do agree with your suggestion, I decided not to touch other settings than proposed in this PR.

Cerrseien:
@rderooy
An incongruent mixer.rate doesn't change the pitch of the audio, and is thus not nearly as affected as mt32.rate is. With mt32.rate being left to its current default value the pitch is audibly lowered, causing people to suspect the DB-X emulation to be flaky.

Monkey Island 1 + 2 for example support both AdLib and MT32 output (using the a/r parameters when starting the games, like "MONKEY a" or "MONKEY2 r"). mixer.rate=44100 will still sound fine with AdLib, but mt32.rate=44100 with MT32 does not. Unfortunately I cannot even test this behaviour with 44100 since my sound device doesn't even support the option anymore, it's fixed to 48K.

I honestly don't know why mt32.rate isn't just supplanted by midi.rate.

PR #3392, opened 2022-04-06
Fix egregious bashisms in configure.ac; cleanup various build scripts

roflcopter4:
This is a bit of a random mashup of changes, but I couldn't quite help myself. The more of the build system I looked at, the angrier and angrier I got. It's just...

Eh. Look. I composed a long angry rant in my head about this, and it should probably stay in my head. Long story short this PR should allow shells other than bash to grok the configure script without breaking much. Probably. I poked around at a few sub-projects that wanted to override `AR` and `RANLIB` (infuriating: can break the build if compiling with lto), and moved the optimization flags that were being set in `configure.ac` behind `$CFLAGS`, where they should be. Unless you're the Linux kernel or glibc, don't override my flags.

The vast majority of the actual changes here are simple quoting of shell variables. I also removed every usage of `test -a` and `test -o` which are ill defined and pointless. Fixed a few cases of `[ $foo == $bar ]` and even one case of `[[ ...`. Also, one script forgot to put an exclamation point in its shebang. Nice.

The Perl scripts also now look a bit less like they were written in the 90s.

Pointless new feature: allow compilation with an arbitrary C++ version (minimum C++11). Feel free to veto this dubious addition. C++20 does not compile (enumeration constants cannot be compared with integers without a cast, nor with constants from other enumerations), C++17 has a cascade of warnings and won't compile unless the C keyword `register` is defined away.

Given the messiness of this PR and the rudeness of this write-up I'm certainly willing to fix up any obvious complaints and try again. Or maybe just throw tomatoes at me until I leave.


Jookia:
Is there a reason to support other shells than Bash and not just hardcode Bash?

roflcopter4:
The various independent shell scripts can be bash specific, so long as the shebang is properly set to `/bin/bash`. I actually did this for a few of them that would have been non-trivial to fix otherwise. Just don't use bashisms on a script set to run on `/bin/sh`.

As for autoconf though, I don't think that's even possible. The tools automatically set the shebang to `/bin/sh`. One can run the configure script with whatever shell they please, or use the environment variable `CONFIG_SHELL` to cause `configure` to exec itself with whatever it's set to. There isn't really any way to force the use of a particular shell, and there isn't supposed to be. The whole point of autoconf is maximal compatibility with every shell and Unix system that has every existed. An awful lot of work was clearly put into achieving this, and you're along for the ride if you use autotools as your build system.

Check the (online) manual for autoconf. There's a substantial section dedicated to writing shell scripts in the prescribed, compatible format. It's part of the autoconf "language". Not only is it not really possible to force the use of bash, but doing so would be antithetical to the whole ethos of the system. You're basically stuck either ensuring that at least `configure.ac` and other core files are POSIX compliant, or switching to a new build system entirely. Which, frankly, might not be as crazy as it sounds: cmake is pretty nice. It would be a lot of work and a lot of headache though.

Also, bash is slow.

For what it's worth I think I've gotten rid of most of the worst offenders in this PR. To be on the safe side I mostly did it by hand.

Jookia:
Yikes. I wonder why I haven't gotten errors running the autoconf junk on Ubuntu then?

My main worry is people accidentally adding bashisms back.

roflcopter4:
Oops. I completely forgot about this PR. My apologies.

As for your case, at the risk of insulting your intelligence, I have to ask whether you may have linked `/bin/sh` to `/bin/bash` at some point? By default it ought to be `/bin/dash`. Even if you have not, you still would not likely see true _errors_. A configure file is a shell script, obviously, and therefore evaluates conditions by the exit code of programs. Even if said programs are built into the shell, they must be have as if they were true programs. Especially in the case of the `test` command, an error is taken to mean that the condition was false, regardless of what may have caused the error.

Nothing is in place to distinguish an syntax error from a false condition except for a message printed to the standard error stream, which is easily overlooked. This is how I noticed the problem. The configure script will continue to run all the same, and may produce incorrect results. Worse, it could be that the condition was actually false after all and the results may be accidentally correct. Undefined behavior at its best (worst?).

It shouldn't be too difficult to keep things "correct". The GNU manual is detailed and well written on this matter. Many of the changes I made aren't strictly necessary. I'm just a little OCD about quoting shell variables. The `shellcheck` program is worth its weight in gold sometimes, though running it on `configure.ac` will give plenty of false diagnostics by virtue of it not being strictly a shell script.

Jookia:
I haven't messed with any shell stuff on my Ubuntu development VM. I know on Arch and other distributions /bin/sh is /bin/bash, so it's just super easy to add these Bashisms. Not everyone is going to read manuals or run shellcheck.

joncampbell123:
There are conflicts with configure.ac and make-flatpak.sh. Can you resolve those please?

brunocastello:
Ehhh, I understand him. Since Catalina (I think), the default shell in macOS is zsh, but I did revert back to bash after every OS update (Catalina - Big Sur - Monterey).

PR #3401, opened 2022-04-08
Make building a flatpak easier

Jookia:
This makes it so building a Flatpak is easier on distributions like Ubuntu.

### DIFF
--- a/make-flatpak.sh
+++ b/make-flatpak.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 
-if [[ ! -x $(which flatpak-builder) ]]; then
-	echo Please install flatpak-builder
+if [[ ! $(flatpak list|grep org.flatpak.Builder|wc -l) -ge 1 ]]; then 
+	echo Please ensure the flatpak builder is installed by running:
+	echo flatpak install flathub org.flatpak.Builder
 	exit
 fi
 
@@ -11,7 +12,7 @@ if [[ ! $(flatpak list|grep org.freedesktop.Sdk|wc -l) -ge 1 ]]; then
 	exit
 fi
 
-flatpak-builder --repo=myrepo --force-clean build-flatpak contrib/linux/com.dosbox_x.DOSBox-X.yaml
+flatpak run org.flatpak.Builder --repo=myrepo --force-clean build-flatpak contrib/linux/com.dosbox_x.DOSBox-X.yaml
 if [ $? -eq 0 ]; then
 	echo
 	echo You can now install the flatpak by running the following commands:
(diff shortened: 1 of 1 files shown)


rderooy:
Is there no flatpak-builder on ubuntu? I find my way cleaner and shorter.

Also, your changes will conflict with the PR https://github.com/joncampbell123/dosbox-x/pull/3392 from @roflcopter4

Jookia:
There is flatpak-builder on Ubuntu, but it gives errors related to certificates. This actually works.

Currently my fix is necessary for the script to work on Ubuntu, I don't really know what to do about #3392 . Maybe they can rebase.

roflcopter4:
By the look of things it would be trivial to work around any conflicts from this.

Jookia:
I just tried the 'test it without installing by running' command and it fails, probably due to sandboxing. I'm considering doing a PR to remove that line, but it feels like I've made a mistake here.

Any ideas @rderooy ?

rderooy:
To be honest, I'm not a fan of these changes. I fail to see why you would get certificate errors before. And indeed when I now try to run it, this is what I get.

```
$ flatpak run org.flatpak.Builder --run build-flatpak contrib/linux/com.dosbox_x.DOSBox-X.yaml dosbox-x
LOG: Early LOG Init complete
LOG: DOSBox-X's working directory: /home/rderooy/src/github/dosbox-x
LOG: Logging init: beginning logging proper. This is the end of the early init logging
LOG: Logging: No logfile was given. All further logging will be discarded.
LOG: DOSBox-X version 0.83.25 (Linux SDL2)

** (process:48177): WARNING **: 10:51:52.758: Failed to connect to bus: Could not connect: No such file or directory
LOG: E_Exit: Can't init SDL No available video device
E_Exit: Can't init SDL No available video device
```

It seems that SDL cannot find the video output for some reason.
I tried to force it to `SDL_VIDEODRIVER=x11`, but it seems you cannot use both the `flatpak run` and `flatpak override` option at the same time.

Jookia:
I don't see why I get certificate errors either. But I did update to a new Ubuntu version so I'll try seeing if they fixed it.

Jookia:
Yeah latest Ubuntu is still broken with flatpak-builder:

```
Downloading sources
Downloading https://mesa.freedesktop.org/archive/glu/glu-9.0.2.tar.xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   162  100   162    0     0    116      0  0:00:01  0:00:01 --:--:--   116
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
Failed to download sources: module GLU: server certificate verification failed. CAfile: none CRLfile: none
```

Jookia:
I've opened https://github.com/flatpak/flatpak-builder/issues/468

Jookia:
Yeah, given flatpak-builder is broken on Ubuntu I don't think it's worth changing back. Maybe a better idea is to dump the --run instruction and instead just have people install the flatpak? That's what official documentation have.

Issue #3372, opened 2022-04-01
Device Info bit fields


cimarronm:
### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

### Have you checked that no other similar question(s) already exists?

- [X] I have searched and didn't find any similar issues.

### Question

It is very unclear on the usage of DOS device information bits throughout dosbox which doesn't seem entirely consistent with the bit fields defined (see http://www.ctyme.com/intr/rb-2820.htm#Table1423). Specifically, dosbox is using bit 15 and bit 9 which are reserved from what I can tell. Are these bytes actually defined in later versions of dos but just not in ralf brown's list or is dosbox using these in a proprietary way? From what I can tell, it almost seems like bit 15 was mistaken for bit 7 so where bit 7 should have been used (character device instead of file), bit 15 was used instead.

cimarronm:
It looks like the DOS device attribute and device information flags were intermixed resulting in the confusion. I added PR #3382 to clean up the usage.

_From the IBM DOS technical reference_
### Device Attribute Flags
<img width="400" src="https://user-images.githubusercontent.com/1505899/161399838-bebdd02a-6619-4490-9e44-121ac712d5f6.png">

### Device Information Flags
<img width="400" alt="image" src="https://user-images.githubusercontent.com/1505899/161399909-9699cdea-0476-476c-8a2c-59a94ec249f9.png">

cimarronm:
It seems DOS copies the upper byte of the attribute flags into the information word resulting in bit 15 (marked as reserved in documentation) as a duplicated device flag.

From the MS DOS 2 source
```
ioctl_read:
        XOR     AH,AH
        TEST    AL,devid_ISDEV          ; Should I set high byte
        JZ      ioctl_no_high           ; no
        LES     DI,DWORD PTR ES:[DI+sf_fcb+fcb_FIRCLUS]  ;Get device pointer
        MOV     AH,BYTE PTR ES:[DI.SDEVATT+1]   ;Get high byte
ioctl_no_high:
        invoke  get_user_stack
        MOV     DX,AX
        MOV     [SI.user_DX],DX
        transfer    SYS_RET_OK
```

Device detection code in DOS and other Microsoft programs all appear to use bit 7 for device detection but bit 15 should be set as well for any program using reserved bit 15 for device detection.

PR #3416, opened 2022-04-15
Get Information IOCTL system call returned value

cimarronm:
Sets bit 15 on the returned value for devices (marked as reserved bit in documentation) as done in MSDOS.

See #3372 for more information

@Wengier found a program, Softscape Tools, which uses bit 15 from the Get Information IOCTL system call for device detection on `stdout`. There are probably other programs as well that use this reserved bit for device detection.


Wengier:
Thanks for the fix. The Softscape PC-Tools that I use indeed appears to use bit 15 which is the reserved bit, and this should make the emulation match the actual MS-DOS behavior.

PR #3417, opened 2022-04-17
SVN r4474

Allofich:
https://sourceforge.net/p/dosbox/code-0/4474/

This is supposed to fix https://www.pouet.net/prod.php?which=65584, however that program does not work properly in DOSBox-X, whether before or after this commit.

I don't know in what matter SVN was supposedly fixed, but in DOSBox-X the player character is invisible and pieces of wall disappear as you move around.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -19,6 +19,10 @@
     for IOCTL. (cimarronm)
   - Clean up of DOS device and attribute flag usage
     (cimarronm)
+  - Integrated commits from mainline (Allofich)
+    - Use real pointer with font data for reading
+    characters. Also ignore page number for modes with
+    only one page.
 0.83.24
   - Release notes for this version:
     https://dosbox-x.com/release-0.83.24.html
--- a/src/ints/int10_char.cpp
+++ b/src/ints/int10_char.cpp
@@ -1499,13 +1499,13 @@ void ReadCharAttr(uint16_t col,uint16_t row,uint8_t page,uint16_t * result) {
         switch (machine) {
         case MCH_CGA:
         case MCH_HERC:
-            fontdata=PhysMake(0xf000,0xfa6e);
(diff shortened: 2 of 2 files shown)


Wengier:
Git bisect shows that this is the first commit that having the said issue (invisible player character, etc):

https://github.com/joncampbell123/dosbox-x/commit/9a07cb807cca9c79dc4a6a655e31b68fd284f962

Setting ``ansi.sys=false`` is a workaround for the issue though.

Allofich:
Thanks Wengier. Whatever problem(s) SVN had with this game might have been different than those I saw in DOSBox-X. I haven't done a before/after test of the SVN change.

I still submitted this PR in the hope that the "ignore page number for modes with only one page" is a correct change, and because as long as we're going to continue taking in updates from SVN, it is helpful to have the base code match because future SVN commits may depend on the changes of earlier ones.

I looked for a discussion about the SVN change on VOGONS or the DOSBox SVN issues forum but couldn't find one.

joncampbell123:
If it takes disabling ANSI.SYS emulation to make the game work, maybe the game is somehow using the ESC character code in it's graphics. Maybe it was coded for systems where ANSI.SYS is not running, and was never tested otherwise.

The best way to confirm that is to put it on an MS-DOS boot disk, boot the disk in DOSBox-X, and run the game. Then add ANSI.SYS to CONFIG.SYS, boot the disk again, and run the game.

joncampbell123:
@Allofich Your fix does allow the game to render correctly. You still have to turn off ANSI.SYS emulation though, because some chars are control characters that ANSI.SYS would interpret rather than print.

Wengier:
I can confirm that in real DOS system with ANSI.SYS loaded, the game will show the player character just fine. So I don't think the game as is is incompatible with ANSI.SYS.

Issue #1431, opened 2020-03-04
Yellow screen in highcolor and truecolor modes on macOS


degaart:
**Describe the bug**
15-bit, 16-bit, 24-bit and 32-bit color VBE applications give corrupted yellowish output on macOS.

**To Reproduce**
1. Download univbe
2. Run vbetest.exe
3. Choose interactive visual compliance tests
4. Choose a high-color or true-color mode
5. Output lacks blue component and is tinted yellow

**Expected behavior**
Correct output

**Screenshots**

**Environment (please complete the following information):**
 - macOS 10.14.6
 - DOSBox-X release version 0.83 SDL2
 - machine=svga_s3

**Additional context**
Looks like issue #614 

Attached screenshot and debug log

<img width="752" alt="Screen Shot 2020-03-04 at 11 22 53" src="https://user-images.githubusercontent.com/12181081/75859553-498f4a00-5e0b-11ea-9ec4-2a813162f313.png">

[dosbox-x-0.83-sdl2.log](https://github.com/joncampbell123/dosbox-x/files/4286134/dosbox-x-0.83-sdl2.log)

msikma:
Just wanted to note that this goes for every VGA/SVGA machine. CGA, EGA and many others work just fine and show a normal screenshot with blue channel in it. So there's something specific to the VGA code that's causing this.

I recall vanilla DOSBox having a similar bug that only occurred on high color depth, like for example when playing Archimedean Dynasty.

msikma:
Maybe the same bug as #727 even though that one's specifically about SDL1 and this is about SDL2.

brunocastello:
This issue is still present with the SDL2 build as of 0.83.10

brunocastello:
This issue is still persistent with the SDL2 build 0.83.24.

Anyone know how can we fix it? It'd be quite important because apparently the SDL2 build is more stable than SDL1, which I had been using for years because of this bug in particular...

myon98:
I've experienced this too, and from a quick glance through the codebase I think it just assumes the pixel format of the `SDL_Surface` returned from `SDL_GetWindowSurface` without checking. And on MacOS it seems like it is BGRA on SDL1 while ARGB on SDL2.

It seems that some places check if `defined(MACOSX) && !defined(C_SDL2)` but others only check if `defined(MACOSX)`, causing it to incorrectly apply the correction to SDL2 as well.

[...]

Issue #614, opened 2018-03-13
Yellow screen running Windows 98SE in more than 256 colours on macOS [16bpp renderer does not account for Mac OS X odd RGBA byte order]


jamesfreeman959:
Hi there,

First of all, a huge thank you for continuing the development of DosBox - I'm a huge fan and have been playing with your builds recently. I run a Mac day to day so I am testing your latest macOS binary (v0.82.4). 

Yesterday I installed Windows 98SE (retail build, 2222A) into dosbox-x. It installs flawlessly and runs really well. However, I notice that as soon as I increase the color depth beyond 256 colours, I get what looks like a yellow filter across the whole screen. This occurs at any resolution, and at 16, 24 and 32-bit colour depths.

Is there a way to fix this - for example a setting that needs tweaking on the Mac? If there's anything I can do to help debug this please let me know.

Kind regards,

James

joncampbell123:
If logging is enabled, provide that log file. The SDL init code provides the RGBA pixel layout given by the Mac OS X system.

As I understand it, Mac OS X uses an odd RGBA pixel layout and this is likely related to that.

jamesfreeman959:
Thank you - I've generated a log file from a clean boot to Windows 98 and then shutdown. Not sure if my logging options are correct - let me know if you need something more generating.

[dosboxw98.log](https://github.com/joncampbell123/dosbox-x/files/1808568/dosboxw98.log)

joncampbell123:
Please try running it from the Terminal app, at the command line, using the --debug switch. The logging level was not high enough for the GFX code to write the SDL RGB bitmasks.

jamesfreeman959:
Sorry about that - please find a new log attached with the --debug switch specified.

[dosboxw98.log](https://github.com/joncampbell123/dosbox-x/files/1809006/dosboxw98.log)

joncampbell123:
297067970       VGAMISC:VGA:S3:SEQ:Read from illegal index  9
 297067972       VGAMISC:VGA:S3:SEQ:Write to illegal index  9
 297068669       VGA:h total 336 end 256 blank (256/270) retrace (260/268)
 297068669       VGA:v total 806 end 768 blank (777/797) retrace (780/782)
 297068669       VGA:VGA refresh rate is now, 70.052
 297068669       VGA:screen: 1.333, scanfield: 1.345, scan: 0.817, vratio: 1.333
[...]

PR #3428, opened 2022-04-24
Fixed color in some modes under x86-64 MacOS SDL2

myon98:
From issue #1431, originally fixed for SDL1 at #614, there still seem to be some issue with regards to video and MacOS when using SDL2.

I'm not too familiar with the codebase, but from a quick glance through it it seems that DosBox-X assumes the pixel format returned from `SDL_GetWindowSurface()` without checking, which seems to be ARGB on everywhere else while BGRA on MacOS SDL1.

It seems that in some places where the conversions for MacOS occur there is a check to specifically target SDL1 OSX, while others apply it to SDL2 as well. I've added some at render_templates.h which fixed the issue for my particular situation.

There are other places that looks like this however, such as in sdlmain.cpp, vga_draw.cpp, output_opengl.cpp, but I didn't include them in this commit, as I'm not too sure about my understanding of the codebase.

I've only tested it on my Win 3.1 guest, as I don't have a convenient way to test many video configurations. Also as this Mac is not mine I don't have a way to test it very frequently. I'd like to ask for comments and contributions from people more experienced with this project or using the MacOS SDL2 port frequently.

Both screenshot from MacOS 11 (Big Sur) x86-64. First is the latest binary release MacOS SDL2 0.83.24, next is the latest git master with this PR applied. Guest is Windows 3.10 Japanese, with svga_s3 in 1024x768 64k colors.

dosbox-x-sdl2
dosbox-x-sdl2-fixed

### DIFF
--- a/src/gui/render_templates.h
+++ b/src/gui/render_templates.h
@@ -123,7 +123,7 @@
 #elif DBPP == 16 // xRRRRRGggggBBBBB -> RRRRRGggggGBBBBB
 #define PMAKE(_VAL) ((_VAL & 31)|((_VAL & ~31)<<1)|((_VAL&0x0200)>>4))
 #elif DBPP == 32 // xRRRrrGGGggBBBbb -> RRRrrRRRGGGggGGGBBBbbBBB
-# if SDL_BYTEORDER == SDL_LIL_ENDIAN && defined(MACOSX) /* Mac OS X Intel builds use a weird RGBA order (alpha in the low 8 bits) */
+# if SDL_BYTEORDER == SDL_LIL_ENDIAN && defined(MACOSX) && !defined(C_SDL2) /* Mac OS X Intel builds use a weird RGBA order (alpha in the low 8 bits) */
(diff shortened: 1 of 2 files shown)


myon98:
Reading the issue I've linked to, I think I should have tested it with UniVBE. I'll post the results if I can test it tomorrow.

brunocastello:
How can I test by compiling my own ARM build with this PR?

rderooy:
You need to clone the git repo and then you can use the github `gh pr checkout 3428` command to switch to the pr branch and build it. You do need the github cli installed.

Other option is to clone the repo and download the changed files.

brunocastello:
> You need to clone the git repo and then you can use the github `gh pr checkout 3428` command to switch to the pr branch and build it. You do need the github cli installed.
> 
> Other option is to clone the repo and download the changed files.

`git fetch REMOTE pull/PRNUMBER/head:BRANCHNAME` also works without having to isntall the git cli.

brunocastello:
@rderooy @myon98 just built a SDL2 version for M1 Mac, it works. I booted a fresh new Win 98 SE install and I've tested with both 16 and 32 bit colors. No more yellow tinted screen.

Wengier:
I think people like @emendelson had complained several times about such an issue. Very nice fix indeed. Thanks @myon98.

PR #3430, opened 2022-04-26
Update manpage and bash completion for -machine option

rderooy:
Got tired of manually trying to write the manpage in troff and
converted it to asciidoc which can convert to a manpage using:

asciidoctor -b manpage dosbox-x.asciidoc

This is not part of the build process, so for now whenever the
asciidoc is updated, it needs to be converted manually.

With https://github.com/joncampbell123/dosbox-x/commit/f462870dacf2cde74ec3850a1fc4598cd7c4ee86 there is now support for the `-machine` option which is missing from the manpage and bash completion. This adds the support.


Wengier:
Thanks for the option update in the script. I removed the commit in my PR to avoid potential conflict with your PR.

Issue #3433, opened 2022-04-28
Building from source, build script cannot find SDL2 on my machine


brunocastello:
### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

### Have you checked that no other similar bug report(s) already exists?

- [X] I have searched and didn't find any similar issues.

### What operating system(s) this bug have occurred on?

macOS Monterey latest version

### What version(s) of DOSBox-X have this bug?

Latest code + Pull Request 3428

### Describe the bug

While building with script `build-macosx-sdl2` :

checking for sdl-config... no
checking for SDL - version >= 1.2.0... configure: error: SDL 1.x or SDL 2.x is required to compile this program

I can confirm that I have SDL2 installed on my Mac from brew, version 2.0.20, latest. However the build script can't find it? I may have to edit the script to point it to where SDL2 is?

### Expected behavior

Build successfully a binary.

### Steps to reproduce the behaviour

Clone the repo, fetch the pull request 3428, and proceed to try and compile from build-macosx-sdl2 script.

Wengier:
I think the build-macosx-sdl2 script does try to compile the builtin SDL2 library. How about try the automated DOSBox-X builds here:

https://github.com/joncampbell123/dosbox-x/actions

brunocastello:
Lst night in my test (not the automated builds yet) the build script did use the built in version of SDL2, but the version I already had installed was not detected before. I am trying to determine if the SDL version has something to do with my sound issues or not

Jookia:
Could you dump your config.log and also location of SDL homebrew files?

brunocastello:
[config.log](https://github.com/joncampbell123/dosbox-x/files/8603184/config.log)

@Jookia there you go:

Location of SDL homebrew files:

`MacBruno-Air:~ bruninho$ brew --prefix sdl2`
`/opt/homebrew/opt/sdl2`

brunocastello:
@Jookia list of files:

[list.txt](https://github.com/joncampbell123/dosbox-x/files/8603292/list.txt)

following the output of the list, `/opt/homebrew/opt/sdl2` is a symlink it seems.

sdl2-config appears to be a program. I just ran it to say the version number as seen in my previous comment.

Jookia:
The config.log you've posted seems to imply if found SDL2 properly:

```
configure:8280: checking for sdl2-config
[...]

PR #3448, opened 2022-05-03
configure.ac: Fix error in setpriority test

Jookia:
On some compilers this error would happen:

> conftest.c:39:5: error: second parameter of 'main' (argument array) must be of type 'char **'
> int main(int argc,char * argv) {
>     ^

This is absurd because the test clearly has argv[] in it, but it's
getting dropped for some reason. Just avoid this by writing char**

Some weird configure bug?

No

Not sure

## Additional information

I noticed this in #3433 . I will need @brunocastello to test this and send me or upload their config.log if they can.


Jookia:
I believe this affects macOS given this seems to fix the check in CI and the issue was discovered on a macOS system.

PR #3449, opened 2022-05-03
acinclude: Enable SDL2 by default

Jookia:
This sets configure to have --enable-sdl2 enabled by default

Issues like #3433 show that having SDL2 installed isn't enough for configure to build with it, --enable-sdl2 needs to be set by default too.

No

Yes: Builds now use SDL2 by default. This breaks every non-SDL2 build script.
However, this is not my doing: The code in configure.ac prefers SDL2 if it is available already, this just enables finding SDL2.

I'm putting this PR in draft status to figure out what to do about that. @joncampbell123 and @Wengier have experience with building the actual application so I'd really enjoy the feedback.

It might be worth flipping things around and disabling SDL1 by default and updating build scripts so they pass --disable-sdl1.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -8,7 +8,7 @@ AC_ARG_WITH(sdl2-exec-prefix,[  --with-sdl2-exec-prefix=PFX Exec prefix where SD
 AC_ARG_ENABLE(sdl2test, [  --disable-sdl2test      Do not try to compile and run a test SDL program],
 		    enable_sdl2test=no, enable_sdl2test=yes)
 AC_ARG_ENABLE(sdl2,     [  --enable-sdl2           Enable SDL 2.x],
-		    enable_sdl2enable=$enableval, enable_sdl2enable=no)
+		    enable_sdl2enable=$enableval, enable_sdl2enable=yes)
 
   AH_TEMPLATE(C_SDL2,[Set to 1 to enable SDL 2.x support])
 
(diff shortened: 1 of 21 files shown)


joncampbell123:
If you change whether SDL2 is default enabled, then the build-debug and build-debug-sdl2 scripts need to be updated so that build-debug still does SDL1.

Jookia:
Is it just the build-* scripts? Aren't there more?

joncampbell123:
No, that's about it. The build scripts and configure.ac.

Your change shouldn't affect the VS2019 project files.

Jookia:
Okay, I've fixed things up, time to see what breaks. :)

Jookia:
Uh I accidentally closed this.

Wengier:
I wonder if it is better to name the option ``--enable-sdl`` as ``--enable-sdl1`` to better reflect that it is for enabling SDL1 instead of SDL in general?

Jookia:
That could be an idea but I don't want to break anyone's build scripts.

Jookia:
Is there an update on this?

Jookia:
Is there any interest in this?

joncampbell123:
Is this no longer a work in progress?

Jookia:
I don't know, nobody's told me what to do next.

Jookia:
I'm going to close this PR for now and later replace it with a different PR later.

Issue #3463, opened 2022-05-05
Output TTF mode switching breaks when running ELKS


Jookia:
### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

### Have you checked that no other similar bug report(s) already exists?

- [X] I have searched and didn't find any similar issues.

### What operating system(s) this bug have occurred on?

Ubuntu 22.04 amd64

### What version(s) of DOSBox-X have this bug?

latest master SDL, SDL2, flatpak

### Describe the bug

Triggering a TTF output switch to OpenGL by trying to open a GUI or take a screenshot causes the virtual machine to lock up and the TTF mode to become unavailable.

### Expected behavior

The machine should switch back to TTF mode like it does in DOS mode.

### Steps to reproduce the behaviour

1. Boot an ELKS image
2. Try to open a menu while in TTF mode

### Used configuration

```ini
Default configuration with TTF enabled.
```

Jookia:
(I understand this bug report is sub-par but I intend to dig more in to it myself)

Jookia:
It seems to happen when I switch from non-TTF output to TTF in ELKS.

Jookia:
Log for when switching to TTF in ELKS:

```
SDL: TTF activated DejaVuSansMono.ttf
Loaded system codepage: 437
DOS MCB dump:
FINAL: Type=0x7c(|) Seg=0x021d size=0xfae5 name='F5­1^^AKP'
End dump
E_Exit: Corrupt MCB chain
```

Log for switching from booted TTF in ELKS:

```
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  25 (X_SendEvent)
  Resource id in failed request:  0x2e0000d
  Serial number of failed request:  69
  Current serial number in output stream:  70
```

Booting FreeDOS in the VM and switching doesn't create this issue.

grapeli:
If you start with the `ttf` video output then you can switch to other outputs or go back to `ttf` without any problems.

I tested the ELKS on this image.
https://github.com/jbruchon/elks/releases/download/v0.5.0/fd1440.img
`dosbox-x -fastlaunch -set "dosbox quit warning=false" -set "dos floppy drive data rate limit=0" -set output=ttf -c "imgmount 0 ./fd1440.img -t floppy -fs none" -c "boot a:"`

Jookia:
[...]

PR #3467, opened 2022-05-06
output_ttf: Don't initialize code page fonts with a disabled DOS kernel

Jookia:
This fixes some buggy TTF behaviour.

Switching to TTF output when ELKS (or a non-DOS kernel) will corrupt the VM. See #3463

No

No

## Additional information

Unsure if this is the right approach. I get the implication that as initcodepagefonts runs DOS commands it's not meant to run here.

Strangely enough it doesn't crash FreeDOS. I guess it implements enough of a DOS kernel to work.


Wengier:
Thanks @Jookia for fixing the issue!

PR #3390, opened 2022-04-05
src/cpu: Implement hinting NOPs

Jookia:
On Pentium Pro and later, opcodes 0x0f 0x18 through 0x0f and 0x1f nops.
These were later defined to be prefetch and multi-byte nop instructions.

This fixes PR #3209 and adds a feature only I seem to care about.
It implements it on all CPU cores and tested on amd64 Linux.

Test assembly:

```org 100h
; Print Hello world
mov dx,msg
mov ah,9
int 21h
; Test all hints
db 0x0f, 0x18
db 0x0f, 0x19
db 0x0f, 0x1a
db 0x0f, 0x1b
db 0x0f, 0x1c
db 0x0f, 0x1d
db 0x0f, 0x1e
db 0x0f, 0x1f
; Test nops
nop
db 0x66, 0x90
db 0x0f, 0x1f
db 0x0f, 0x1f, 0x40, 0x0     
db 0x0f, 0x1f, 0x44, 0x0,  0x0       
db 0x66, 0x0f, 0x1f, 0x44, 0x0,  0x0 
db 0x0f, 0x1f, 0x80, 0x0,  0x0,  0x0, 0x0    
db 0x0f, 0x1f, 0x84, 0x0,  0x0,  0x0, 0x0, 0x0
db 0x66, 0x0f, 0x1f, 0x84, 0x0,  0x0, 0x0, 0x0, 0x0
db 0x66, 0x2e, 0x0f, 0x1f, 0x84, 0x0, 0x0, 0x0, 0x0, 0x0
; Quit
mov ah,4Ch
int 21h
msg db 'Hello, World!',0Dh,0Ah,'$'
```

Compile and test:

```
nasm -f bin test.asm -o out.com
out.com
```

PR #3498, opened 2022-05-16
cpu: Don't dynamically recompile PREFETCH instructions

Jookia:

This should fix booting Linux and other operating systems that rely on it.

Found by https://github.com/joncampbell123/dosbox-x/pull/3390#issuecomment-1126700686 .

No.

No.


grapeli:
I confirm with this patch now linux boots correctly with `core=dynamic_x86` or `dynamic_rec` and `cputype=pentium_iii`.

Issue #930, opened 2018-11-13
PC-98: Graphic LIO not implemented


hdk1983:
**Describe the bug**
Graphic LIO is a set of routines in BIOS ROM.  I found it was used by the following games/applications:

- KATALITH
- MAZE_999 https://www.vector.co.jp/soft/dos/game/se023680.html
- Microsoft Basic programs that use graphics (probably MAZE_999 is built on Microsoft Basic)

Graphic LIO is in segment F990h.  If you see some errors around in segment F990h, probably it is related to this issue.  For example:

- KATALITH: ERROR CPU:Write 0 to rom at f9900
- MAZE_999: ERROR CPU:Write 20 to rom at f9900

**To Reproduce**
1. Run a program that uses graphic LIO
2. The program does not start

**Expected behavior**
The program works

**Additional context**
To verify the graphic LIO is used, I added a dummy implementation to DOSBox-X as follows:

[dummyliopatch.txt](https://github.com/joncampbell123/dosbox-x/files/2576252/dummyliopatch.txt)

With this patch, KATALITH shows the following copyright screen:

![katalith_000](https://user-images.githubusercontent.com/1381939/48414108-1ac01100-e78d-11e8-949f-f04b5e411b2e.png)

And, MAZE_999 shows the following error (maybe because of the dummy implementation), looks like an error of Microsoft Basic:

![game_999_000](https://user-images.githubusercontent.com/1381939/48414129-26133c80-e78d-11e8-8169-279670ea7cec.png)

I made a hello world program using MSGBOX statement of Visual Basic for DOS:

[HELLO2.zip](https://github.com/joncampbell123/dosbox-x/files/2576255/HELLO2.zip)

With the above patch, this program starts with broken graphic screen (maybe because of the dummy implementation):

![hello2_000](https://user-images.githubusercontent.com/1381939/48414149-388d7600-e78d-11e8-8e2b-82f021da9f0c.png)

The hello world works on Neko Project:

![hello2_np21w](https://user-images.githubusercontent.com/1381939/48414320-b81b4500-e78d-11e8-9adc-086fefcf3bb9.png)

joncampbell123:
Actually, yes, I noticed "Sword of Legend" is quite consistent at jumping to segment F990, evidently without checking properly.

joncampbell123:
If anyone has a PDF or link to documentation on the LIO interface that would be helpful.

The source code of Neko Project II appears to explicitly implement the LIO interface in bios_lio()

[...]

Issue #692, opened 2018-04-30
PC-98: Text not scrolling correctly in Legend of Heroes II's staff roll


yksoft1:
It scrolls like this in DosBox-X with following messages generated.
![ffano](https://user-images.githubusercontent.com/37106774/39413864-1786c410-4c20-11e8-9e0b-0768409d9b33.gif)

> LOG: PC98 CRTC w: port=0x70 val=0x00 unknown
> LOG: PC98 CRTC w: port=0x72 val=0x0F unknown
> LOG: PC98 CRTC w: port=0x74 val=0x10 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x00 unknown
> LOG: PC98 CRTC w: port=0x78 val=0x1F unknown
> LOG: PC98 CRTC w: port=0x7A val=0x16 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x01 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x02 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x03 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x04 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x05 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x06 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x07 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x08 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x09 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x0A unknown
> LOG: PC98 CRTC w: port=0x76 val=0x0B unknown
> LOG: PC98 CRTC w: port=0x76 val=0x0C unknown
> LOG: PC98 CRTC w: port=0x76 val=0x0D unknown
> LOG: PC98 CRTC w: port=0x76 val=0x0E unknown
> LOG: PC98 CRTC w: port=0x76 val=0x0F unknown
> LOG: PC98 CRTC w: port=0x76 val=0x00 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x01 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x02 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x03 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x04 unknown
> LOG: PC98 CRTC w: port=0x76 val=0x05 unknown
[...]

Issue #3434, opened 2022-04-29
DOSBox-X Discord channel ☎📱


Wengier:
While DOSBox-X is mostly based on GitHub, there is also an official DOSBox-X Discord channel, which some people may find useful for chatting since communications via Discord is real-time and thus may be more effectively. There are both public and private message sections, and Discord app also exists for smart phones and tablets, which may notify you when there are new messages. The DOSBox-X Discord channel is available from:

https://discord.gg/5cnTmcsTpG

People who use Discord can surely join the official DOSBox-X Discord channel for chatting about DOSBox-X.

Wengier:
The post was intended to announce the existence of the official DOSBox-X Discord channel, but since DOSBox-X is mostly GitHub based you may not expect many people around in the Discord channel, and GitHub is still recommended for most purposes.

scrutinizer80:
Can this be synched to LiberaChat IRC?

Wengier:
Discord and IRC are different types of communication systems, but I think there may be tools available to do such synchronization’s between the two.

roblox34:
Invite delete

ckeisc:
The invite seems to have expired. Got new one or could it be fixed? Thanks

Wengier:
Try this one then:

https://discord.gg/2N4wnYaEYF

Dobby233Liu:
> there may be tools available to do such synchronizations between them.

One of the server I am in has a bot called dIRCord that may be for that, though it seems that it may be private

MasterO2:
> While DOSBox-X is mostly based on GitHub, there is also an official DOSBox-X Discord channel, which some people may find useful for chatting since communications via Discord is real-time and thus may be more effectively. There are both public and private message sections, and Discord app also exists for smart phones and tablets, which may notify you when there are new messages. The DOSBox-X Discord channel is available from:
> 
> https://discord.gg/5cnTmcsTpG
> 
> People who use Discord can surely join the official DOSBox-X Discord channel for chatting about DOSBox-X.

@Wengier, that discord link no longer works. Can you re-post one that works?

joncampbell123:
The Discord server is invite-only now. Look in the Discussions sections for how to request access.

This is due to repeated instances where some users kept posting spam and crypto spam.

PR #3508, opened 2022-05-21
PC-98 GDC and LIO drawing support

nanshiki:
GDC and LIO drawing are implemented based on the Neko project II.
If the GDC clock is 5 MHz, DGD must be set to 1 in VECTW. However, this drawing process does not refer to the PITCH value, so if the screen settings pass normally, DGD can draw normally even if DGD is 0.
GDC command interpretation time and drawing time are not taken into consideration.
LIO adds GCIRCLE ellipse and fill and GPAINT1/GPAINT2 functions.
#930 I could not get KATALITH, but MAZE_999 and HELLO2 can be displayed.
#2421 FILMTN now displays correctly.

Other bug fixes.
int 18h, ah=42h does not change the read/write page of VRAM.
#692 Corrected the valid range of the setting value for the number of text scrolling lines (port 0x76). Some applications expect the set value 0x10 to display the same as 0x00.
Bit7 of memory 0000:054D is changed by int 18h ah=4dh. In PC-9821, Bit7=0 by default. int 18h ah=4dh ch=01h will make Bit7=1. This is confirmed to be the case on an actual PC-9821.
Fixed incorrect reading of odd addresses when EGC port 04A4h Bit13=0.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -92,6 +92,8 @@ HQ2X and HQ3X render scaler (ScummVM, Maxim Stepin; GPLv2+) src/gui/render_templ
 
 PC-98 FM board emulation (Neko Project II; BSD 3-clause) src/hardware/snd_pc98/*
 
+PC-98 GDC and LIO drawing support (Neko Project II; BSD 3-clause) src/hardware/vga_pc98_gdc_draw.cpp src/ints/pc98_lio.cpp
+
 QCOW image support (Michael Greger; GPLv2+) src/ints/qcow2_disk.cpp
 
 JEGA and DOS/V support (nanshiki, Wengier; GPLv2+) include/jfont.h src/ints/int_dosv.cpp
--- a/include/pc98_gdc.h
+++ b/include/pc98_gdc.h
@@ -50,6 +50,22 @@ struct PC98_GDC_state {
     void take_reset_sync_parameters(void);
     void cursor_advance(void);
 
+    void draw_reset(void);
+    void vectw(unsigned char bi);
+    void exec(uint8_t command);
+    void prepare(void);
+    void draw_dot(uint16_t x, uint16_t y);
+    void pset(void);
+    void line(void);
+    void text(void);
+    void circle(void);
+    void box(void);
+    void set_vectl(int x1, int y1, int x2, int y2);
+    void set_mode(uint8_t mode);
+    void set_csrw(uint32_t ead, uint8_t dad);
+    void set_textw(uint16_t pattern);
+    void set_textw(uint8_t *tile, uint8_t len);
+
     void begin_frame(void);
     void next_line(void);
 
@@ -60,13 +76,50 @@ struct PC98_GDC_state {
     bool fifo_empty(void);
     uint16_t read_fifo(void);
 
+    enum {
+        RT_MULBIT   = 15,
+        RT_TABLEBIT = 12,
+        RT_TABLEMAX = (1 << RT_TABLEBIT)
+    };
+    struct VECTDIR {
+        int16_t x;
+        int16_t y;
+        int16_t x2;
+        int16_t y2;
+    };
+    static const VECTDIR vectdir[16];
+    static const PhysPt gram_base[4];
+    static uint16_t gdc_rt[RT_TABLEMAX + 1];
+
+    struct GDC_DRAW {
+        PhysPt base;
+        uint32_t ead;
+        uint16_t dc;
+        uint16_t d;
+        uint16_t d1;
+        uint16_t d2;
+        uint16_t dm;
+        uint16_t pattern;
+        uint16_t x;
+        uint16_t y;
+        uint8_t tx[8];
+        uint8_t dad;
+        uint8_t ope;
+        uint8_t dir;
+        uint8_t dgd;
+        uint8_t wg;
+        uint8_t dots;
+        uint8_t mode;
+        uint8_t zoom;
+    } draw;
+
     /* NTS:
      *
      * We're following the Neko Project II method of FIFO emulation BUT
      * I wonder if the GDC maintains two FIFOs and allows stacking params
      * in one and commands in another....? */
 
-    uint8_t                 cmd_parm_tmp[8];            /* temp storage before accepting params */
+    uint8_t                 cmd_parm_tmp[11];            /* temp storage before accepting params */
 
     uint8_t                 rfifo[PC98_GDC_FIFO_SIZE];
     uint8_t                 rfifo_read,rfifo_write;
--- a/include/pc98_gdc_const.h
+++ b/include/pc98_gdc_const.h
@@ -3,11 +3,21 @@ enum {
     GDC_CMD_RESET = 0x00,                       // 0   0   0   0   0   0   0   0
     GDC_CMD_DISPLAY_BLANK = 0x0C,               // 0   0   0   0   1   1   0   DE
     GDC_CMD_SYNC = 0x0E,                        // 0   0   0   0   1   1   1   DE
+    GDC_CMD_MODE_REPLACE = 0x20,                // 0   0   1   0   0   0   0   0
+    GDC_CMD_MODE_COMPLEMENT = 0x21,             // 0   0   1   0   0   0   0   1
+    GDC_CMD_MODE_CLEAR = 0x22,                  // 0   0   1   0   0   0   1   0
+    GDC_CMD_MODE_SET = 0x23,                    // 0   0   1   0   0   0   1   1
     GDC_CMD_CURSOR_POSITION = 0x49,             // 0   1   0   0   1   0   0   1
+    GDC_CMD_CSRW = 0x49,                        // 0   1   0   0   1   0   0   1
     GDC_CMD_CURSOR_CHAR_SETUP = 0x4B,           // 0   1   0   0   1   0   1   1
+    GDC_CMD_ZOOM = 0x46,                        // 0   1   0   0   0   1   1   0
     GDC_CMD_PITCH_SPEC = 0x47,                  // 0   1   0   0   0   1   1   1
+    GDC_CMD_VECTW = 0x4C,                       // 0   1   0   0   1   1   0   0
+    GDC_CMD_TEXTE = 0x68,                       // 0   1   1   0   1   0   0   0
     GDC_CMD_START_DISPLAY = 0x6B,               // 0   1   1   0   1   0   1   1
+    GDC_CMD_VECTE = 0x6c,                       // 0   1   1   0   1   1   0   0
     GDC_CMD_VERTICAL_SYNC_MODE = 0x6E,          // 0   1   1   0   1   1   1   M
     GDC_CMD_PARAMETER_RAM_LOAD = 0x70,          // 0   1   1   1   S   S   S   S    S[3:0] = starting address in parameter RAM
+    GDC_CMD_TEXTW = 0x78,                       // 0   1   1   1   1   0   0   0
     GDC_CMD_CURSOR_ADDRESS_READ = 0xE0          // 1   1   1   0   0   0   0   0
 };
--- a/src/hardware/Makefile.am
+++ b/src/hardware/Makefile.am
@@ -12,7 +12,7 @@ libhardware_a_SOURCES = adlib.cpp dma.cpp gameblaster.cpp hardware.cpp iohandler
 			vga.cpp vga_attr.cpp vga_crtc.cpp vga_dac.cpp vga_draw.cpp vga_gfx.cpp vga_other.cpp \
 			vga_memory.cpp vga_misc.cpp vga_seq.cpp vga_xga.cpp vga_s3.cpp vga_tseng.cpp vga_paradise.cpp \
 			cmos.cpp disney.cpp gus.cpp mpu401.cpp ipx.cpp ipxserver.cpp ne2000.cpp hardopl.cpp dbopl.cpp innova.cpp dongle.cpp \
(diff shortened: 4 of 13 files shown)


maron2000:
@nanshiki 
It's great to have GLIO implemented.
I confirm FILMTN graphic mode works, and also "46-okunen Monogatari" works as well.
st1_000

joncampbell123:
Just my two cents... I suppose masking the CRTC rows by 0x0F is correct for the 200-line modes, but according to a real PC-9821 laptop I test with, it seems to respond to the low 5 bits, not 4 bits.

What does masking by 0x0F fix?

joncampbell123:
I think I found a case where your GDC draw support needs a little work. Windows 3.1 uses the GDC draw commands too.

joncampbell123:
Notice from an earlier commit the Paintbrush interface is drawn properly.
krnl386_000

joncampbell123:
I'm going to modify the GDC draw code where it concerns memory access. mem_readb() and mem_writeb() are CPU-side memory I/O, don't use them from hardware access, please. Your VGA card cannot cause the CPU to read/write memory and neither should the GDC. I just went over this with @Wengier, and for the same reason, the graphical issues and possible instabilities are resolved by commenting out any part of GDC draw that calls mem_readb()/mem_writeb().

This isn't an issue for PC-98 games because they either run in real mode or they run in virtual 8086 mode where the video memory isn't virtualized in any way. It's protected mode environments like Windows 3.1 through ME that issuing mem_readb()/mem_writeb() like that can cause problems.

joncampbell123:
Final question: I get the impression that the GDC functions entirely on it's own from the EGC hardware. That would mean that perhaps when the GDC draws on planar memory it does so directly to the planar memory, not by reading/writing through the EGC. Right?

nanshiki:
I checked on an actual PC-9821, and in 640x400 mode, the scrolling once returns to 0 at port 76h = 10h, and after 11h, it seems to shift one by one. However, after 10h, the character display of the first line becomes strange.
In 20-line mode, the scrolling seems to return to 0 at port 76h=14h because the scanning line of one line is 20.

The software is "joshikousei-no-nazo-time-slip-toujankyou-maboroshi-ware-neo-kobe-pc98-ia" from the DOSBox game compat testing chart.
Scrolling occurs after the mahjong game, and the character display has shifted since then. It seems that we are expecting the scroll position to return to 0 at port 76h = 10h.

I did not understand about mem_readb()/mem_writeb(). Please correct the code.
The book I have has an example of using EGC to speed up GDC drawing, so I assume that GDC writes also go through EGC.

joncampbell123:
@nanshiki I will test that game. I am adding a function to vga_memory.cpp to give your GDC draw code a more direct interface to the read/write functions of PC-98 VRAM than through mem_readb/mem_writeb.

I will work on the code assuming that GDC read/modify/write cycles go through EGC the same as CPU I/O.

I'll restore your masking of that register by 0x0F in that case. When I first wrote that code I assumed that since other registers were 5 bits that maybe the register in question was 5 bits.

nanshiki:
As for port 76h, I think it is valid for 5 bits because the scroll position changes up to 13h in 20-line mode (character 80x20).
The proposed fix is as follows.
```
pc98_text_row_scroll_lines = (unsigned char)val % (pc98_text_first_row_scanline_end - ((pc98_text_first_row_scanline_start & 0x10) ? -(0x20 - pc98_text_first_row_scanline_start) : pc98_text_first_row_scanline_start) + 1);
```
This still works slightly differently from the actual machine when the number of scanlines is exceeded.
I don't think that settings that exceed the number of scanlines are usually made, except for mahjong games.

Please correct the GDC and LIO.

maron2000:
It seems that it scrolls upward if 0x00 - 0x0f is thrown to port 76h,
and downwards if 0x1f- 0x10 so it seems to be 5-bits.

joncampbell123:
> It seems that it scrolls upward if 0x00 - 0x0f is thrown to port 76h, and downwards if 0x1f- 0x10 so it seems to be 5-bits.

So perhaps it should be treated as 5 bits that are sign extended?

joncampbell123:
Some games in my test set that previously didn't work, seem to work fine now. Despite what I pointed out, the LIO BIOS does work. :+1: 

Test case "Ancient Dragon" needed it, which is apparently like Tetris, but you build swords block by block instead. :raised_eyebrow: 
guest os_000

joncampbell123:
@nanshiki Can you look at Balance of Power? I know it uses LIO BIOS functions but it still doesn't look right. It doesn't seem to clear bitplanes correctly.

guest os_000

joncampbell123:
@nanshiki Speaking of using EGC to enhance the GDC line draw, I just noticed that I can make 99.99999999% of the line drawing artifacts disappear in Windows 3.1 if I modify the REPLACE/SET case of draw_dot() only write, but not read from, the video memory. Which means Windows 3.1 sets the GDC draw pattern to 0xFF and the EGC to a XOR raster op.

I'm also noticing that the GDC_CMD_MODE_*: cases are documented to then accept data bytes that, in graphics modes, set the pattern to either 0xFFFF or 0x0000 based on the LSB. The pattern is 0xFFFF, therefore the line draw is effectively XORing the entire 8 bits and that's why the black blocks around the white line. I think I can fix that without breaking any games.

joncampbell123:
Please try the latest code. I don't think I've broken anything your code fixes. I've added a hack to the GDC drawing code that does dummy reads when EGC is in use which is necessary to completely eliminate line drawing artifacts in Windows 3.1.

As for the LIO BIOS functions, your code has enabled working display and graphics in many games of the msdos test set that were previously marked as not displaying anything. :+1:

joncampbell123:
@nanshiki Can you make another pull request adding your changes to the CHANGELOG so you can be credited properly for your work in the next release?

nanshiki:
@joncampbell123 My English is poor, so can you please correct the CHANGELOG there?

joncampbell123:
@nanshiki What would you like me to add to the CHANGELOG?

nanshiki:
I will leave the contents to you.
But please write that the GDC and LIO drawing were implemented based on the Neko project II source.

I have tried drawing with EGC by GDC and there seems to be a problem.
I have prepared a test program.
[line.zip](https://github.com/joncampbell123/dosbox-x/files/8769036/line.zip)

When executed, the program first draws with GDC only, and when the key is pressed, it draws with GDC using EGC.
On the actual machine, both are displayed the same, but in DOSBox-X, the color of the line drawn later seems to change the color of the dots existing in the same word.
If the actual machine is GDC 5MHz
\> line 1
add a dummy option and execute.

joncampbell123:
@nanshiki The GDC according to this datasheet is said to draw using read/modify/write memory cycles. Prior discussion says that those apparently goes through the EGC.

However R/M/W with EGC seemed problematic with Windows 3.1 until I added a hack to issue a dummy read (result ignored) then to write the pattern. It only worked if the R/M/W acted as if the read got back 0x00.

joncampbell123:
@nanshiki I see nothing in LINE.C that changes anything about EGC. Where does it write to EGC registers?

nanshiki:
@joncampbell123 EGC drawing in LINE.C uses the GRCG tile register, which is set via outp(0x7c,0xc0) in linee() to simultaneously write multiple planes to a color line with a single GDC call. This source is taken from the EGC speedup sample source in the GDC Technical Book (ISBN: 4890520902).
LINE.EXE is a compilation of LINE.C with MS-C 6.0.

EGC is upward compatible with GRCG. GRCG could only be used for writing from the CPU, but EGC supports writing from the GDC. See the following (in Japanese)
https://kei-sakaki.jp/2017/06/09/reason-talked-closed-specification-of-pc-9800-egc/
My knowledge is limited to the above book and some internet resources, so I don't know the details of GDC and EGC integration.

nanshiki:
@joncampbell123 Balance of Power PC-9801 version, but the screen images I found in my search seem to be in monochrome.
Perhaps it uses the PC-9801 monochrome mode. This mode is set when port 68h = 03h, and the display color is 8x16 units and text attributes are used.
I believe this is not supported by DOSBox-X.

joncampbell123:
@nanshiki Though reading through Google Translate, I think that website you linked to is saying that the EGC has a GRCG compatible mode and EGC enhanced mode, and the GRCG mode is available to both the CPU and the GDC when either one does memory I/O. It does *not* say that the GDC has access to enhanced EGC functions.

However, the behavior of Windows 3.1 suggests that maybe some enhanced EGC functions are possible with GDC line drawing. I checked using debug logging with GDC drawing state and Windows 3.1 uses the REPLACE mode of WDAT commands and the EGC raster op for XOR, not the COMPLEMENT mode as you'd expect.

GDC memory I/O is said to be read/modify/write, however if I do that with Windows 3.1 artifacts occur because the read registers the white background and XORs it with the white background producing the black vertical strips as seen in the video I posted sometime back.

This complicated behavior will have to be determined precisely by writing more test code in DOSLIB and running it on real PC-9821 hardware, which is also an opportunity to examine your other test program regarding the text scroll region TEST.EXE.

I suppose DOSBox-X could make more progress at this point adding the monochrome mode you mentioned.

joncampbell123:
Basic test with DEBUG.EXE on a real laptop: The 8-color "digital" mode is active. Switching on monochrome mode makes bitplane 2 the only displayed bitplane, as if bitplane 2 becomes all three planar bits. I used DEBUG.EXE to poke some bytes into planar graphics.

It helps to remember that bitplanes from B800 to A800 are G R B order. In 16-color modes memory appears at E000 and is the "E" plane to provide the 4th bit.
G = B800
R = B000
B = A800
E = E000

https://user-images.githubusercontent.com/6245486/170411649-05e20ba0-2185-4227-86d9-57054f13b43c.mp4

joncampbell123:
Which bitplane becomes the one monochrome bitplane is affected by the 8-color digital palette registers, confirmed.

joncampbell123:
If you switch on monochrome mode and 16-color analog mode, there are some strange results. It seems to only show white only if BITPLANE2 == 1 && BITPLANE3 == 0. However perhaps the monochrome mode was never intended for anything but 8-bit digital mode.

nanshiki:
The aforementioned book also states, "EGC has a compatibility mode with GRCG, and even allows it to be used from GDC." and it does not say that EGC's ROPs, etc. can be used.
However
https://www.webtech.co.jp/company/doc/undocumented_mem/io_egc.txt
shows that the EGC "enters between the CPU and VRAM, or GDC and VRAM, and performs various logic operations on the data to be written to VRAM." So I think it is possible to use extended functions.
The "PC-9800 Series Technical Data Book, Supplement Edition" on the previous page seems to have a detailed explanation of EGC, but I have the "New Edition PC-9800 Series Technical Data Book", which does not have that explanation.

As for the monochrome mode, we will try various things here as well.

joncampbell123:
@nanshiki I think monochrome mode is simple enough. It's for 8-color digital graphics mode. Bitplane 2 becomes the only displayed bitplane, after palette remapping through A8h-AEh even. It acts strangely in 16-color analog graphics mode, but probably because you're not supposed to do that :)

joncampbell123:
If there is another register that controls which bitplane is THE bitplane in monochrome mode, that would be good to know.

nanshiki:
Monochrome mode, but it is not an operation that can be handled by the palette.
A sample program has been created. Please execute it on the actual machine.
[mono.zip](https://github.com/joncampbell123/dosbox-x/files/8776650/mono.zip)
Pressing the key once switches to monochrome mode, and then each subsequent press changes the color of the text attribute.
You can see that the color of the text attribute colors the graphic portion of the same character area.

It is strange that it is called monochrome mode but colors are applied, but this seems to be the same behavior as the 640x400 of the PC-8801 (8-bit machine before the PC-9801). Like the simple graphics mode, this feature is for compatibility with past models, and its use was not recommended.
The monochrome mode and simple graphics mode seem to have been removed from the late PC-9821.

joncampbell123:
What I mean is that, at least on the PC-9821 laptop I have (a 486), monochrome mode seems to function like it gets a 3-bit color from the 3 bitplanes, then it remaps it through the 8-color digital palette, then *after* mapping, bit 2 determines whether the pixel is white or black.

joncampbell123:
Balance of Power does not appear to use monochrome mode. The log would say so if it saw unknown port 68h write 02h/03h.

However the log does say "LIO GCIRCLE not support flags: 05".

joncampbell123:
@nanshiki Just to confirm, are you suggesting that the color of the monochrome graphics display plane is determined by the color attribute of the text layer at that point on screen?

So that would mean that despite two separate genlocked GDCs in the system, the text and graphics output work together in monochrome mode to determine if the pixel is set or not and what color?

By the way, bitplane 2 at B800 represents green, and I am aware that traditionally in RGB hookups that carry monochrome, the green pin is usually used to carry that signal, which is probably why that plane was chosen. I think MDA does the same thing while using the same 9-pin connectors as CGA in the IBM PC world.

I am away from the test unit at this time and I cannot confirm on real hardware until later tonight.

nanshiki:
@joncampbell123 MONO.EXE, but it seems to display the same as the actual device. It is wonderful.
I think the Balance of Power is not using the port, but the GSCREEN function of LIO or int 18h ah=42h. If int 18h ah=42h and bit 5 of ch register is 1, it will be monochrome mode.
GSCREEN of LIO also calls this, so please try the following modification for now.
In bios.cpp INT18_PC98_Handler(), add the following to case 0x42:
```
    IO_WriteB(0x68, (reg_ch & 0x20) ? 0x03 : 0x02);
```
If the display still does not work with Balance of power, you may need to add a process to change the palette according to the display screen selection in the parameter ds:[bx+3] of GSCREEN in LIO.

joncampbell123:
According to the GSCREEN code it is indicating a preference for mono, but then it calls INT 18h AH=42h with CH=C0h to cancel it out.

I am using the copy available on The Internet Archive "Neo Kobe" collection:

https://archive.org/download/NeoKobe-NecPc-98012017-11-17/ASCII.zip/ASCII%2FBalance%20of%20Power%2FBalance%20of%20Power%20%5BFD%5D.zip

nanshiki:
@joncampbell123 Neko Project also has the same symptom without BIOS.ROM, so I still think it is a problem with the monochrome mode setting in GSCREEN in LIO.
I will continue to investigate.

joncampbell123:
@nanshiki DOSBox-X has a Discord channel if you're interested. https://github.com/joncampbell123/dosbox-x/issues/3434

PR #3575, opened 2022-06-15
FAT SFN quirk when the path starts with 0xe5

myon98:
This checks if the short filename starts with 0xe5, replaces it with 0x05 before storing, and vice versa when reading back from disk.

I recently found out a quirk of FAT short file names, where if the first character happens to be 0xe5 it has to be replaced with 0x05 due to clashing with the value to mark the dirent as free[0]. In CP932 this corresponds to Kuten codepoints row 73 and 74[1] which are mostly fairly obscure Level 2 Kanji characters, but an example is 蜃 (E5 87), which means that current DOSBox-x refuses to create a file named 蜃気楼.TXT (mirage, haze) in IMGMOUNTed filesystems[2].

mtools cannot handle this either, but Linux[3] and FreeBSD[4] seem to do.

This patch blindly translates initial 0x05 in dirent to 0xe5 and vice versa, just like in FreeBSD and Linux.

No

I do not believe so, but considering that introducing a bug in filesystem implementation can be catastrophic, it might be necessary to treat this patch carefully.

### Example disk image
[e5sfn.hd4.gz](https://github.com/joncampbell123/dosbox-x/files/8907307/e5sfn.hd4.gz)
This is a 1.44MB raw floppy image that contains a SFN-only 蜃気楼.txt in CP932.
Confirmed on MS-DOS 6.2/V, MS-DOS 5.0A NEC, MS-DOS 6.2 EPSON, FreeDOS(98), Mac OS J1-9.2.1, Windows 2000 JPN, Linux, US English MS-DOS 6.22 (shows 蜃 as σç instead of ♣ç)
### References
[0]: MS FAT spec: https://download.microsoft.com/download/1/6/1/161ba512-40e2-4cc9-843a-923143f3456c/fatgen103.doc
[1]: JIS codepoint table: https://en.wiktionary.org/wiki/Appendix:Japanese_kanji_by_JIS_X_0208_kuten_code#Row_73
[2]: FAT guide by elm chan: http://elm-chan.org/docs/fat_e.html#dir_struct
[3]: Linux: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fat/namei_msdos.c?h=v5.18#n56
[4]: FreeBSD: http://fxr.watson.org/fxr/source/fs/msdosfs/msdosfs_conv.c?v=FREEBSD-12-STABLE#L504

### DIFF
--- a/src/dos/drive_fat.cpp
+++ b/src/dos/drive_fat.cpp
@@ -292,6 +292,11 @@ static void convToDirFile(const char *filename, char *filearray) {
                 if(charidx >= 11) break;
                 if(filename[i] != '.') {
                         filearray[charidx] = filename[i];
+                        if (charidx == 0 && filearray[0] == (char)0xe5) {
+                                // SFNs beginning with E5 gets stored as 05
+                                // to distinguish with a free directory entry
+                                filearray[0] = 0x05;
+                        }
                         charidx++;
                 } else {
                         charidx = 8;
@@ -2410,6 +2415,9 @@ nextfile:
 	memset(find_name,0,DOS_NAMELENGTH_ASCII);
 	memset(extension,0,4);
 	memcpy(find_name,&sectbuf[entryoffset].entryname[0],8);
+	// recover the SFN initial E5, which was converted to 05
+	// to distinguish with a free directory entry
+	if (find_name[0] == 0x05) find_name[0] = 0xe5;
     memcpy(extension,&sectbuf[entryoffset].entryname[8],3);
 
     if (!(sectbuf[entryoffset].attrib & DOS_ATTR_VOLUME)) {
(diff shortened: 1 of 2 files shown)


myon98:
I think an additional fix is required for handling volume label, which needs the 0xe5 -> 0x05 substitution not only for the dirent in the root directory but also for the one in BPB. Also [this](https://github.com/joncampbell123/dosbox-x/blob/5a526541dd89d9247262abfd9409d4b1015ccf1a/src/dos/drive_fat.cpp#L773) and [this](https://github.com/joncampbell123/dosbox-x/blob/5a526541dd89d9247262abfd9409d4b1015ccf1a/src/dos/drive_fat.cpp#L837) doesn't seem to be DBCS aware when doing upper case conversion (as confirmed by タチツテト getting corrupted to タチイウエ).

PR #3604, opened 2022-07-02
make icon square

rderooy:
Change the icon from a rectangle to a square

flathub has a new mandatory check during build to see if the icon is square, or the build simply fails. This fixes the issue.

The steps I took using Inkscape:
- Modify canvas to be square
- Centre icon on canvas


joncampbell123:
That seems a bit OCD for an app store, but, OK.

PR #3614, opened 2022-07-05
Rotate icon to look like an X

rderooy:
@Wengier requested that I rotate the icon some more such that it looks more like an X.


rderooy:
I am uncertain about the PNG icon. It seems the original may be an animated icon from what github is showing?

Also for the Windows ICO file, the original contains 6 images consisting of 3 resolutions (16x16, 32x32 and 48x48) in two different colour maps (RGB and RGBA). I re-created all 6 images and packed them as a new ICO file but it would be good for somebody to check that it is working as intended.

The BMP and PNG file were simply exported as such from inkscape using the following commands:
```
inkscape -w 55 -h 55 -o dosbox-x.png dosbox-x.bmp
inkscape -w 48 -h 48 -o dosbox-x.png dosbox-x.svg
```

The ICO file required some more steps using a combination of inkscape and ImageMagick.
```
inkscape -w 16 -h 16 -o 16-rgba.png dosbox-x.svg
inkscape -w 32 -h 32 -o 32-rgba.png dosbox-x.svg
inkscape -w 48 -h 48 -o 48-rgba.png dosbox-x.svg
convert 16-rgba.png -channel RGB 16-rgb.png
convert 32-rgba.png -channel RGB 32-rgb.png
convert 48-rgba.png -channel RGB 48-rgb.png
convert *rgb*.png dosbox-x.ico
```

rderooy:
I checked the original PNG with a few tools on Linux, and none are saying it is an animated PNG or contains multiple images. Including Gimp and the ImageMagick `identify -verbose` command.

Issue #3603, opened 2022-07-02
Use year.month as version number?


emendelson:
### What you want

Perhaps it would make sense to use the year and month for the version number, rather than the very arbitrary 0.84.xx? The current version would be 2022.07. This would make it easier for everyone to keep track of when the version they're using was created. This is just a suggestion...

### Have you checked that no similar feature request(s) exist?

- [X] I have searched and didn't find any similar feature request.

### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

Jookia:
Yeah, it's a little silly to still be at version 0 when we're shipping monthly releases of a mature software project with backwards compatibility in mind. Having predictable version numbers also makes it easy to just know if someone's up to date. Jumping to 2022.07 would solve this.

Jookia:
Solved. Next version will be 2022.08.0

PR #3640, opened 2022-07-20
Jump to version 2022.08.0

Jookia:
Move the version from the planned 0.84.2 to 2022.08.0 (earlier revision was 22.08)

https://github.com/joncampbell123/dosbox-x/issues/3603

No

The version number is now no longer 0.xx.x


Jookia:
I see two criticisms of this PR:

The first is that we've dropped a digit. I guess we could go with YY.MM.DD instead?
The second is that it's not safe for 21XX dates, but that can be solved by just increasing the YY. Any number will be larger than before.

Jookia:
Given some feedback from @Wengier in private and my own above I've changed this PR to use version 2022.08.0 instead

PR #3642, opened 2022-07-23
update-version-number: Fix MyAppVersion not escaping quotes

Jookia:
update-version-number is not escaping quotes properly, fix this

None

No

No

## Additional information

N/A


Jookia:
I added another quick patch to make it executable.

PR #3673, opened 2022-08-10
Adding German translation

Link-Mario:
This pull request adds 2 german translation files with one additional task.

Because this is a quick translation, please review it, do corrections (where necessary), then add it to master.

The second file is necessary, because there are not enough strings exclusively for PC-98-mode, to display all translations correctly.
This means, more placeholder strings are necessary.

### DIFF
--- a/contrib/translations/de/de.lng
+++ b/contrib/translations/de/de.lng
@@ -1,0 +1,3474 @@
+:DOSBOX-X:LANGUAGE:deutsch
+:DOSBOX-X:CODEPAGE:437
+:DOSBOX-X:VERSION:0.84.0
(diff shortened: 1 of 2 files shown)


rderooy:
I had a quick look, and there are a lot of problems with the line lengths. For instance, run a command such as `CONFIG` or `MOUNT -?` or `MOUNT -examples` and the lines are way too long and the output is a mess due to line wraps.

Translating strings is one thing, getting it to fit in the space available is often the real challenge.

Link-Mario:
Mission accomplished.

There may be some strings, that look OK when using `help /a`, but not when using e.g. `for /?`.

Issue #3535, opened 2022-06-01
Mount zip file doesn't work since version 0.83.23 ?


aetchris:
### Describe the bug

Programs mounted in zip files don't launch since after version 0.83.23. 

### Steps to reproduce the behaviour

Mount a zip file as drive c
Try to launch program
Dosbox-x hangs and program doesn't launch

### Expected behavior

In version 0.83.23, the program launches.
After this version, dosbox-x seems to hang up with a black screen

### What operating system(s) this bug have occurred on?

Windows 10

### What version(s) of DOSBox-X have this bug?

Since 0.83.24

### Used configuration

```ini
mount C "pathtozipfile\file.zip:\"
C:
cd\
program.exe
```

### Output log

```text
Many "ERROR CPU:Illegal Unhandled Interrupt Called 6" in output

373108514 ERROR CPU:Illegal Unhandled Interrupt Called 6
 373108611 ERROR CPU:Illegal Unhandled Interrupt Called 6
 373108708 ERROR CPU:Illegal Unhandled Interrupt Called 6
 373108805 ERROR CPU:Illegal Unhandled Interrupt Called 6
 373108902 ERROR CPU:Illegal Unhandled Interrupt Called 6
 373108999 ERROR CPU:Illegal Unhandled Interrupt Called 6
 373109096 ERROR CPU:Illegal Unhandled Interrupt Called 6
```

### Have you checked that no similar bug report(s) exist?

- [X] I have searched and didn't find any similar bug report.

### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

grapeli:
I confirm does not work from the moment of introducing these changes #3347.

aetchris:
The issue seems to be resolved in version 0.84.1. Thanks a lot!

grapeli:
I have not noticed any improvement. It still does not work.

[keen6.zip](https://github.com/joncampbell123/dosbox-x/files/9032749/keen6.zip)
[bomber.7z.txt](https://github.com/joncampbell123/dosbox-x/files/9032750/bomber.7z.txt)

```
dosbox-x -c "mount c ./bomber.7z" -c "c:" -c "cd 3dbomber" -c bomb
dosbox-x -c "mount c ./keen6.zip" -c "c:" -c keen6
```

After undoing these unfortunate code changes #3347, it works.

aetchris:
[...]

PR #3678, opened 2022-08-14
Fix issues 3535 (dosbox-x hang on zip-mounted drives) and 3347 (minor bug fix and related)

tbr:

This PR proposes a fix for issue #3535 and some code fixes/cleanup for the merged #3347.

### Longer explanation

I was trying to work with a ZIP file mounted as drive in read-only, with an overlay mount on top for changes. In order to collect the contents of the ZIP, I first used a regular folder mount as drive, testing all the programs inside of the drive, and making sure everything worked (which was the case). However, putting the same contents in a ZIP file (and mounting as described) makes dosbox-x crash with a "ERROR CPU:Illegal Unhandled Interrupt Called 6" when running the same content (DOS programs that worked fine under a regular folder mount).

Even a simple "edit bla.txt" produces the same hang of dosbox-x (when bla.txt is in a ZIP mounted drive).

I saw that #3535 described this problem, and the comments say that the temporary solution is to revert #3347 and rebuild. Digging deeper into this, I think I found two bugs:

1) In the code that was merged by #3347, there seems to be a mistake in the 0x3f code (DOS File read). This code now calls MEM_BlockRead instead of MEM_BlockWrite in one of the possible paths (actually in the added path for DOS file I/O device drivers). And, in certain true conditions MEM_BlockWrite is not called at all. To me, it seems to be a minor mistake. So, this PR fixes this and cleans up the code a little bit.

2) The issue described in #3535 seems to be caused by the PhysFS code that sets the EXT_DEVICE_BIT (0x200) on the file handle objects for files in a ZIP (when doing open or create file). I believe this bit should NOT be set, as the files are just regular DOS files (not special device files). This bit causes the #3347 code to think it needs the file I/O device driver path to be called when instead it's just reading or writing regular files.

### Disclaimer

Now, I'm not (yet) familiar with the inner workings of dosbox and dosbox-x, so it's perfectly possible that this PR is completely wrong. If that is the case, then my apologies and just reject the PR (maybe let me know why). At least, for me it seems this PR solves the issue in #3535.

I hope not. Please review.

### DIFF
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -1941,30 +1941,24 @@ static Bitu DOS_21Handler(void) {
                 }
 
                 dos.echo=true;
-
-                if(handle >= DOS_FILES) {
+                if(handle >= DOS_FILES || !Files[handle] || !Files[handle]->IsOpen()) {
                     DOS_SetError(DOSERR_INVALID_HANDLE);
-                } else 
-                if(!Files[handle] || !Files[handle]->IsOpen())
-                    DOS_SetError(DOSERR_INVALID_HANDLE);
-                else if(Files[handle]->GetInformation() & EXT_DEVICE_BIT)
-                {
+                }
+                else if(Files[handle]->GetInformation() & EXT_DEVICE_BIT) {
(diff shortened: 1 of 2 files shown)


leecher1337:
FYI: Unfortunately, functionality of #3347 is broken now.

PR #3719, opened 2022-09-08
Makefile.am: Don't ignore setcap error with || true

Jookia:
This just changes the output a bit.

Suggested by @rderooy

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -132,11 +132,11 @@ install: src/dosbox-x
 	install -m 644 $(srcdir)/contrib/linux/dosbox-x.1 $(DESTDIR)$(mandir)/man1
 	mkdir -p $(DESTDIR)$(prefix)/share/bash-completion/completions
 	install -m 644 $(srcdir)/contrib/linux/dosbox-x $(DESTDIR)$(prefix)/share/bash-completion/completions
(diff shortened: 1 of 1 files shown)


VVD:
FreeBSD don't have command `setcap` - why is it mandatory to run it from Makefile?

VVD:
```
# setcap
-bash: setcap: command not found
# uname -srm
FreeBSD 13.2-RELEASE-p8 amd64
```

rderooy:
Your point being?

I suspect there may be slightly more Linux users than *bsd users

As such, either fix the makefile in such a way that it does not remove the setcap for Linux systems, or don't use `make install`.

VVD:
Run it on Linux only. Isn't it obvious?

Jookia:
It's not mandatory. This PR checks if it exists. Are you having build errors or is this nitpicking?

VVD:
This PR remove ` || true`.

During build I get error:
```
command -v setcap >/dev/null && setcap cap_net_raw=ep /wrkdirs/usr/ports/emulators/dosbox-x/work/stage/usr/local/bin/dosbox-x
*** Error code 127 (ignored)
```
But look like "build farm" can ignore this error.

P.S. I'm a FreeBSD ports maintainer and working on update port emulators/dosbox-x.

Jookia:
Do you not have the command 'command'?

VVD:
Thanks for explanation.
Never used command `command`.
And sorry for noise.

PR #3722, opened 2022-09-09
Make system buildable on FreeBSD and NetBSD

Jookia:
This is a long PR that tries to fix most if not all bugs on the BSDs preventing them from building and running out of the box using ./build with the embedded SDL variant. Building with external SDL2 works using ./configure .

No PR but a lot of packagers are working around these bugs.

Working BSD support.

Not sure. I did some refactoring of the configure and build scripts. Everything seems fine from my tests. Let's see what CI thinks.


Jookia:
I'm converting this to a draft because I'd really appreciate some feedback and I'd like to try and get the embedded SDL2 working.

rderooy:
I'm a bit worried that this will break something on the other platforms. And as all the CI builds currently fail, that seems to be the case.
Once the CI builds run, I would like to try building flatpak and RPM packages to see if anything is broken.

Jookia:
I think what I'm going to do is try and split out some of the autotools 'refactoring' from this branch which is creating a lot of noise.

Jookia:
Okay I think I nailed out most the bugs. CI seems to be running.

Jookia:
Okay I'm fairly happy with this patchset. The only downside is my NetBSD environment throws a skippable SDL2 assertion with the internal SDL2.

rderooy:
successfully compiled and launch tested on F36 x86_64:
- local SDL1 build
- local SDL2 build
- local flatpak SDL2 build
- local RPM SDL2 build

p.s. should the internal SDL2 v2.0.9 not be bumped to the latest upstream release (2.24.0)?

joncampbell123:
> successfully compiled and launch tested on F36 x86_64:
> 
>     * local SDL1 build
> 
>     * local SDL2 build
> 
>     * local flatpak SDL2 build
> 
>     * local RPM SDL2 build
> 
> 
> p.s. should the internal SDL2 v2.0.9 not be bumped to the latest upstream release (2.24.0)?

There are probably many good reasons to update to the latest SDL2.
Let @Wengier know when you do it though. He'll need to reapply his SDL2 IME composition code to the new library.

joncampbell123:
Are you ready for me to merge this?

There is only a conflict in the CHANGELOG that needs to be addressed, and then I can merge.

Jookia:
Yes I'm ready. I'll fix the CHANGELOG today.

Jookia:
Fixed. Ready for merge.

Jookia:
I don't have the time to keep resolving changelog conflicts. Consider this PR dropped.

Torinde:
it's a pity that was dropped :(

Jookia:
Realistically this just means packagers have to patch each version to work on the BSDs. They're already doing this, so nobody is really impacted outside people compiling from source and developers.

Torinde:
Still better if your work above isn't in vain and gets utilized.
Isn't end of the month when main developers prepare new releases? The other two PRs don't touch the CHANGELOG (from what I see), so if you fix it now - it'll be adopted fast? 
Or maybe you can exclude the CHANGELOG from the PR and leave it for the maintainer to add a log entry for this change...

Jookia:
If there's more interest in merging I'll re-open but there's also the issue of code rot if someone starts touching places this PR touches, and also the issue that there's not BSD users to really review or test this.

Issue #3355, opened 2022-03-25
Overlay Filesystem: existing files in archives


thesnable:
### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

### Have you checked that no other similar question(s) already exists?

- [X] I have searched and didn't find any similar issues.

### Question

Hi!

(see my other question)

i mount a zip file, and add an overlay. Scenario: sndsetup.ini exists in the archive. I want to update it with new setting from the setup program. The new sndsetup.ini is saved to the overlay drive: but still the old one is used and visible in the directory listing.

Bug or Feature?

I would recommend to use the overlay file first....

thanks alot
marco

Wengier:
An overlay over zip files is not a standard overlay drive. It internally uses PhysFS for processing, instead of overlay file system. PhysFS and standard overlay drives may have subtle differences in handling this case.

thesnable:
Ok is it possible to implement this? Actually i have to delete those files from the archive first... also does it know to which archive it belongs?

Wengier:
Archives and write directories are separate, but connected together by the PhysFS library. If PhysFS supports this, then it will be supported by DOSBox-X automatically. Hope there are updates to PhysFS for such issues.

thesnable:
Ok thanks alot then!

NebularNerd:
I'll mention this here as well as in #3535, the most awesome @icculus has offered a solution to the contents of overlays not superseding the original zip contents here: icculus/physfs#18 

https://github.com/joncampbell123/dosbox-x/blob/3b002fa724a67b3cf26d5030391cbf3d5dea90a0/src/dos/drive_physfs.cpp#L371

>Make that 1 a 0 so the overlay gets checked for files before the thing it overlays.

This should then let mounted .zip/overlay combo's work as we expect them to (maybe in conjunction with an unmount/remount or rescan depending on how it behaves), saving the need to delete files inside the .zip just to experiment with a mod or sound configuration.

icculus:
(To be clear, I didn't personally test this, and someone mentioned that there's PhysicsFS and a different mounting thing, too, but for things going through PhysicsFS, it will favor files in mounts that are _earlier_ in the search path, so calling PHYSFS_mount with a 0 for the third parameter puts that thing at the _start_ of the search path instead of the _end_, which is what you would want for an overlay mechanism.)

NebularNerd:
[...]

PR #3723, opened 2022-09-09
Update drive_physfs.cpp to support overlay files in place of .zip files

NebularNerd:

Permits overlay contents to be used in place of .zip contents. Replaces missing ; that causes some .zips to not work.

Per joncampbell123/dosbox-x#3535, joncampbell123/dosbox-x#3355 and icculus/physfs#18 this fixes contents of overlay directory not superseding the original .zip contents, for example: changing a config file or testing a patch.  

This should have uses for many people, adding/removing files from the .zip could risk breaking the game or program or corrupt the file. With this in place you can edit existing files and use the overlay version in place of the .zip original. If it breaks something you can delete the overlay and go back to the stock version. Another use could be be multiple overlays with one base .zip for testing CGA, EGA, TANDY versions without needing to duplicate the base .zip (for NTFS users you can then use folder compression so the overlays take even less space).

There is also the missing ; before -mypos on line 746 discussed in joncampbell123/dosbox-x#3535 which breaks .zip functionality with some software. Restoring this fixes a lot of games that did not behave with it removed.

No

Not that I am aware of, @grapeli and myself have tested it in #3535 and seems to behave as expected.

## Additional information

My first Pull Request 😁

### DIFF
--- a/src/dos/drive_physfs.cpp
+++ b/src/dos/drive_physfs.cpp
@@ -368,7 +368,7 @@ bool physfsDrive::setOverlaydir(const char * name) {
         return false;
 	} else {
         if (oldwrite) PHYSFS_unmount(oldwrite);
-        PHYSFS_mount(newname, NULL, 1);
+        PHYSFS_mount(newname, NULL, 0);
         dirCache.EmptyCache();
     }
 	if (oldwrite) free((char *)oldwrite);
--- a/src/dos/drive_physfs.cpp
+++ b/src/dos/drive_physfs.cpp
@@ -743,7 +743,7 @@ bool physfsFile::Seek(uint32_t * pos,uint32_t type) {
 	switch (type) {
 	case DOS_SEEK_SET:break;
 	case DOS_SEEK_CUR:mypos += PHYSFS_tell(fhandle); break;
-	case DOS_SEEK_END:mypos += PHYSFS_fileLength(fhandle)-mypos; break;
+	case DOS_SEEK_END:mypos += PHYSFS_fileLength(fhandle); -mypos; break;
 	default:
 	//TODO Give some doserrorcode;
 		return false;//ERROR


NebularNerd:
@joncampbell123 Following an update from the PhyFS author on #3535 I suggest blocking the https://github.com/joncampbell123/dosbox-x/pull/3723/commits/e7eca5d9f8dc29ffb2b34e4cc00ddc7bd8abc115 semicolon 'fix' on this PR for now. It appears it fixes things by simply bodging something else that's incorrect into working.

The overlay commit https://github.com/joncampbell123/dosbox-x/pull/3723/commits/5410de76ffc698bc8f2cac6713e97b3d22791eba should be fine.

joncampbell123:
Just curious, what does the ';' fix?

Note that newer GCC compilers are quite good at pointing out possible problems with if() indention and whether it encloses code.

NebularNerd:
> Just curious, what does the ';' fix?
> 
> Note that newer GCC compilers are quite good at pointing out possible problems with if() indention and whether it encloses code.

@joncampbell123 I wish I knew, I'm capable of throwing a working .exe together with some changes and testing but I'm not sure what the `;` fixes (or more possibly bodges) elsewhere in the code.  

In #3535 it was highlighted that https://github.com/joncampbell123/dosbox-x/commit/dc798b56a1bc2cb25c90e4b057815a22822cd1e2 (which points to drive_local.cpp not drive_physfs.cpp, was it split off at some point?) is programatically correct in removing the `;` and it's also confirmed by [PhyFS author](https://github.com/joncampbell123/dosbox-x/issues/3535#issuecomment-1242161237) a bit later in the thread.   

However doing so prevents many .zip based games from starting or running correctly, once you reinsert the `;` a load of games suddenly start working. It's a longish thread now, but you'll see we've been busy testing a few cases and post reinsertion a lot more works, there are still fringe cases but they seem to handle file access in much stranger ways.

joncampbell123:
Looking at the semicolon fix, I can see why.

SEEK_END is supposed to seek relative to the end of the file. That's how it works here on Linux, in Windows, and MS-DOS. The added semicolon makes it work properly by severing the -mypos from the expression (leaving it by itself to do nothing).

lseek(fd,-4,SEEK_END) is supposed to seek to 4 bytes before the end of file (EOF-4), for example.

Perhaps the fix was needed because someone didn't fully understand what SEEK_END is supposed to do.

So the best fix for that change is to remove the -mypos entirely from the SEEK_END case.

image

joncampbell123:
I'll go ahead and merge the pull and remove the now useless extra.

maron2000:
Sorry if I didn't understand correctly but shouldn't it be like below?
I don't know the rationale of once making `mypos` beyond file length, and correcting it in the next conditional check.
```cpp
case DOS_SEEK_END:mypos += PHYSFS_fileLength(fhandle); -mypos; break; // before
case DOS_SEEK_END:mypos = PHYSFS_fileLength(fhandle); break; //after
```

NebularNerd:
So the `;` was fixing it by causing the -mypos to be ignored. I look forward to testing the next build.

@maron2000 Yes it looks like @joncampbell123 will remove the extraneous code to look like your example.

joncampbell123:
> Sorry if I didn't understand correctly but shouldn't it be like below? I don't know the rationale of once making `mypos` beyond file length, and correcting it in the next conditional check.
> 
> ```c++
> case DOS_SEEK_END:mypos += PHYSFS_fileLength(fhandle); -mypos; break; // before
> case DOS_SEEK_END:mypos = PHYSFS_fileLength(fhandle); break; //after
> ```

The file offset in lseek() is a signed integer. You can use SEEK_END with 0 or a negative number to seek relative to the end of the file. A positive offset allows seeking past the end of the file, which does not extend the file unless you then issue a write or use INT 21h to "truncate" out to that position to extend the file.

PR #3726, opened 2022-09-09
fpu: Drop support for using 'long double' for 80-bit floats

### LINKED ISSUE #3499: FPU Long Double Emulation with host CW Synchronization
### Question

I'm trying to understand the use case for when one would want the long-double FPU instruction set being used (fpu_instructions_longdouble.h) and `FPU_SyncCW()` being defined to set the control word of the x87 FPU. If one is going to bother to set the CW of the x87 FPU on the host, why not bother in using the actual x87 instructions instead? I'm not seeing the point in having `FPU_SyncCW()` at all.

### Have you checked that no similar question(s) exist?

- [X] I have searched and didn't find any similar question.

Jookia:
The only systems I can find where 'long double' is 80-bit IEEE is on x86 and Itanium. For x86's case it can use the FPU directly instead of going through a structure. Itanium is dead and hard to support.

Having this code has already caused confusion on non-x86 platforms and we already special-case ARM. Instead of special casing this to be x86 and GCC only we might as well just dump the code entirely.

This also drops the glibc-specific fpu_control.h dependency.

Portability to non-x86 system.

Closes #3499 question.

No.

No.

## Additional information


joncampbell123:
The long double code is intended for x86 and x86_64 because the underlying FPU uses it.
If any other platform is accidentally triggering it, then the headers or config to enable should prevent that.
This code is the only way for normal core to get the full 80-bit precision when available. Without this code, the only way for full precision is to use dynamic core (same limitation as DOSBox SVN).

Jookia:
Interesting. I'll do another PR that adds an arch check then.

PR #3727, opened 2022-09-09
Clean up unused autotools junk

Jookia:
This cleans up some autotools junk.

None.

No.

Drops support for macOS 10.3.


joncampbell123:
What's wrong with checking for socklen_t? At the very least it helps to check whether it is there.

Jookia:
It was only added as a hack for macOS 10.3 and lower. Are there systems where it's not there? If it's code that's not actually used it means we're keeping it and maintaining it for no reason and won't notice if it breaks.

joncampbell123:
Good point. I am aware there are some out there who like to compile DOSBox-X for Mac OS X 10.13 or 10.15 or such but I doubt anyone will compile against 10.3. For one thing, Mac OS X versions that old only run on PowerPC because Apple had not yet moved to x86.

Jookia:
If someone is actually compiling it against 10.3 I'd be in favor of keeping it. I guess if someone starts complaining we could revert?

PR #3729, opened 2022-09-10
Fix PC-98 AVSDRV.SYS INITDRV

nanshiki:
Fixed a case of freezing at the start of the prologue after the opening in POLICENAUTS.
If the internal variables are not initialized with INITDRV, it remains in the playing state and does not proceed to the next step.


nanshiki:
Fixed that the background music stops and the training does not end when the shooting trainer starts in POLICENAUTS.

PR #3745, opened 2022-09-16
More autotools cleanup

Jookia:
Just some more cleanups to the autotools code.

None

No

Probably not

## Additional information

Ok


joncampbell123:
pdcurses is used in MinGW builds, ncurses on Linux builds. Does this change retain the ability to detect and use curses in both cases.

curses is needed for the debugger interface. Without curses, there is no debugger.

Jookia:
Pretty sure it preserves all existing behaviour- it just puts the resulting curses library in a variable instead of having a separate one for each curses implementation.

PR #3794, opened 2022-10-27
Fix serial port handling on unix systems

### LINKED ISSUE #2995: OPL3 Duo! playback corruption in Linux

### Have you checked that no other similar bug report(s) already exists?

Arch Linux

0.83.18 both SDL1 and 2

### Describe the bug

The playback from the OPL3 Duo! board in Linux is very wrong. Hanging notes, notes not playing, instruments are wrong and distorted. I would guess it's timing issue, but it could be something with communication over USB. The board I am using works fine in Windows.

### Expected behavior

Playback should be accurate

Plug in the OPL3 Duo!

Make sure your user account on the system is in the group that the /dev/ttyACM0 device is (in Arch, it's uucp, but other distros apparently use dialout) so Dosbox-X can access it, or run Dosbox-X with sudo.

Edit the Dosbox-X config file to use OPL3 Duo! with the oplport set to ttyACM0

Run Dosbox-X and make sure it was able to initialize the board successfully:

OPL3 Duo! Board: Connecting to port ttyACM0... 
OPL3 Duo! Board: COM Port OK.

Run any game that uses OPL3. I tested with Commander Keen4 and Budokan.

### Used configuration

```ini
Default apart from these changes:

oplemu                                           = opl3duoboard
oplport                                          = ttyACM0
```

### Emulator log

```text
There are no messages in the log after initialization.

OPL3 Duo! Board: Connecting to port ttyACM0... 
OPL3 Duo! Board: COM Port OK.
```

### Additional context

_No response_

### LINKED ISSUE #3694: OPL2board audio distortions and cut-outs (Linux)
### Describe the bug

OPL2board Audio playback is unusably out of time, stuttery and distorted on Linux.

- Compile or install DOSbox-X from source or via Flatpak.
- Make sure your user account is added to the dialout group (required for ttyACM0 access in Debian/Ubuntu/Pop!_OS)
- Connect OPL2board and Arduino (Uno R3 in my case), open Arduino IDE and upload [SerialPassthrough](https://github.com/DhrBaksteen/ArduinoOPL2/tree/master/examples/OPL2AudioBoard/SerialPassthrough) for OPL2board
- Set up DOSBox-X.conf as below
- Load a game with OPL2 audio support, having selected appropriate setup options for the games themselves. I've been testing with Elvira, Quest for Glory 2 and Ultima VI, so that's Adlib, /s and Adlib respectively.

### Expected behavior

Undistorted music to play through OPL2board's 3.5mm port.

Pop!_OS 22.04

0.83.13 to 0.84.2. Affects SDL1 and SDL2 builds, as well as Flatpak

### Used configuration

```ini
[mixer]
nosound         = false
sample accurate = true
swapstereo      = false
rate            = 49761
blocksize       = 8192
prebuffer       = 250



[sblaster]
sbtype                       = none
sbbase                       = 220
irq                          = 7
dma                          = 1
hdma                         = 5
enable speaker               = false
sbmixer                      = false
oplmode                      = opl2
oplemu                       = opl2board
oplrate                      = 49716
oplport                      = ttyACM0
retrowave_bus                = serial
retrowave_port               = 
hardwarebase                 = 220
goldplay                     = true
blaster environment variable = true
```

### Output log

```text
Output log isn't visibly helpful, as correct performance is indistinguishable from errored performance. Here it is with --debug enabled. See Additional information for both a log and a video of the issue with debugging enabled for the opl2board.h.
```

### Additional information

Compiling DOSBox-X with #define OPL2_AUDIO_BOARD_DEBUG 1 in opl2board.h allows us to view every signal being sent to the OPL2board, and the timing of the messages appears to match that of the garbled audio. Occasionally, audio will be sent/received at the correct rate, under which conditions the messages also appear at the correct rate for the music being played.

This can be most  easily observed by putting mixer setting block size at maximum (8192) and prebuffer at 250, but this is by no means reliably reproducible, to the point where the impact may be coincidental. Noise and loss of event timing may be delayed by as much as 30 seconds, but inevitably occurs. 

I've videoed this comparison at:
https://theos-cloud.eu/index.php/s/JWwoeoqPyegdZDp
Cut-out occurs at around 27 seconds, example of distortion at 54 seconds.

Here's the dosbox-x.conf file that's performed best so far and was used when making that video:
https://theos-cloud.eu/index.php/s/BWg63cdP9mkLoJA

And a log of the same effect, taken separately:
https://theos-cloud.eu/index.php/s/XnWZMjpZiAx3XYQ

I previously commented on a [similar bug report](https://github.com/joncampbell123/dosbox-x/issues/2995) about the OPL3 Duo! board, in the same family as my OPL2board, but as they're handled separately in code and I don't have an OPL3 Duo to test with, I've opened a dedicated issue report for this specific hardware.

Wondering if this might be a reversion, I went back to version 0.83.13, when opl2board support was introduced, and the issue is present there. I also reverted to a contemporary version of the ArduinoOPL2 SerialPassthrough sketch. This has only a minor difference which made no practical impact on the problem. 

Do we know what Linux system + OPL2board spec the original implementation was tested under? I've tested on multiple Ubuntu-derived distros since autumn 2021. Perhaps I'm missing a key configuration step that isn't documented in either the DOSBox-X or [ArduinoOPL2 docs](https://github.com/DhrBaksteen/ArduinoOPL2/wiki/DosBox-X-configuration), or there's something unexpected happening with the handling of virtual com ports in this family of distros.

stengun:
Serial port initialization on unix systems is done incorrectly.

We have to consider the fact that serial devices on linux are not reset once a program opens the corresponding device (if device is still plugged in)
This means that every time we open a serial port we must explicitly set / reset all the bits we need (and never use assignment to do this).
~Lastly, opening the serial port with O_NDELAY will make all the read() calls non-blocking, returning -1 if no data is ready to be read. However, some devices could want to set this feature differently on open. A sane default would be using simple Blocking I/O~

Also, when we set COM port parameters, parity bit stuff is set without setting input parity mode.

This PR will fix the above behavior.

No.

No.

### DIFF
--- a/src/hardware/serialport/libserial.cpp
+++ b/src/hardware/serialport/libserial.cpp
@@ -305,11 +305,14 @@ bool SERIAL_open(const char* portname, COMPORT* port) {
 	termios termInfo;
 	memcpy(&termInfo,&cp->backup,sizeof(termios));
 
-	// initialize the port
-	termInfo.c_cflag = CS8 | CREAD | CLOCAL; // noparity, 1 stopbit
-	termInfo.c_iflag = PARMRK | INPCK;
-	termInfo.c_oflag = 0;
-	termInfo.c_lflag = 0;
+	// initialize the device, we first reset any bits we don't need.
+	// this step is needed because edits to serial devices are persistent.
+	termInfo.c_cflag &= ~(PARENB | CSTOPB | CSIZE | CRTSCTS | CMSPAR);
+	termInfo.c_iflag &= ~(IGNBRK | BRKINT | ICRNL | INLCR | PARMRK | INPCK | ISTRIP | IXON); // reset input controls
+	termInfo.c_oflag &= ~OPOST ; // raw output
+	termInfo.c_lflag &= ~(ECHO | ECHONL | ICANON | IEXTEN | ISIG); // no line processing, raw input
+	// Use preferred default options
+	termInfo.c_cflag |= CREAD | CLOCAL; // Standard default, must always be set
 	termInfo.c_cc[VMIN] = 0;
 	termInfo.c_cc[VTIME] = 0;
 
@@ -408,17 +411,35 @@ bool SERIAL_setCommParameters(COMPORT port,
 	termios termInfo;
 	int result = tcgetattr(port->porthandle, &termInfo);
 	if (result==-1) return false;
-	termInfo.c_cflag = CREAD | CLOCAL;
+	// reset flags that will be affected by this function
+	termInfo.c_iflag &= ~(INPCK | ISTRIP);
+	termInfo.c_cflag &= ~(CSIZE | PARENB | PARODD | CSTOPB | CBAUD | CMSPAR);
 
 	// parity
+#ifndef CMSPAR
+	// Mark or space parity bit.
 	// "works on many systems"
-	#define CMSPAR 010000000000
+	#define CMSPAR 010000000000 // Many systems have this define, not posix standard
+#endif
+	// set options according to parity
 	switch (parity) {
-	case 'n': break;
-	case 'o': termInfo.c_cflag |= (PARODD | PARENB); break;
-	case 'e': termInfo.c_cflag |= PARENB; break;
-	case 'm': termInfo.c_cflag |= (PARENB | CMSPAR | PARODD); break;
-	case 's': termInfo.c_cflag |= (PARENB | CMSPAR); break;
+	case 'n': break; // no parity
+	case 'o': // Odd parity
+		termInfo.c_iflag |= INPCK | ISTRIP;
+		termInfo.c_cflag |= (PARODD | PARENB);
+		break;
+	case 'e': // Even parity
+		termInfo.c_iflag |= INPCK | ISTRIP;
+		termInfo.c_cflag |= PARENB;
+		break;
+	case 'm': // Mark parity
+		termInfo.c_iflag |= INPCK | ISTRIP;
+		termInfo.c_cflag |= (PARENB | CMSPAR | PARODD);
+		break;
+	case 's': // Space parity
+		termInfo.c_iflag |= INPCK | ISTRIP;
+		termInfo.c_cflag |= (PARENB | CMSPAR);
+		break;
 	default:
 		return false;
 	}


joncampbell123:
Given this change, how is serial I/O handled without blocking the emulator when no data is moving either direction? Is select() or poll() involved?

stengun:
I tried to look where libserial is actually used inside dosbox-x. Usage is found only in [opl2board](https://github.com/joncampbell123/dosbox-x/blob/38082111bc252e0023913bd59cd3058b0f1fd0fc/src/hardware/opl2board/opl2board.cpp#L13) and [opl3duoboard](https://github.com/joncampbell123/dosbox-x/blob/38082111bc252e0023913bd59cd3058b0f1fd0fc/src/hardware/opl3duoboard/opl3duoboard.cpp#L14).
Those two boards don't need read(), but despite this the flag had an effect also on writes.

Technically, how reading is done using this serial library should depend solely on the implementers. It's up to the programmer to write a proper way of detecting data without blocking the entire emulator. Or at least, with this implementation would be like this. Personally, I think that the option of using blocking IO should be exposed by the opening method, because different hardware could either support or not support blocking IO.

Taking another example, in the codebase is present a board called RetroWave which communicates with dosbox using serial IO. The funny part is that it didn't use libserial but its own library which opens the socket with blocking IO ([here](https://github.com/joncampbell123/dosbox-x/blob/38082111bc252e0023913bd59cd3058b0f1fd0fc/src/hardware/RetroWaveLib/Platform/POSIX_SerialPort.c#L135)). RetroWave doesn't neet do read() either.

I share your concerns, We may try to find a better way of doing this if you think dosbox should offer a libserial that is emulation-safe and then on the hardware protocol implementation they use ioctl commands to se blocking IO on the socket. In my testing, those two OPL* boards work ok without breaking the emulator (I used adlib tracker 2 music to test for that and it indeed works). I'm really open to suggestions.

ps: other than non blocking IO, the other issues addressed by this PR should be put in to have a proper way of setting flags.

stengun:
I'm about to change this PR. I found a better way to fix the related issues without forcing blocking IO open mode (which should not be part of this pr anyway).
The concerns about blocking IO are real, but have to be addressed in a different way than this one. Specifically, I will address them in the relevant hardware files and give them separate PRs.

Device flags are a different matter tho, this PR corrects flag set on open and fixes a possible bug when setting parity mode (Input parity is not being set). I can't test this last thing because there's nothing that uses it.

Let me know if you prefer to close this instead.

maron2000:
MacOS builds complain and fails that flags CBAUD and CMSPAR are not available.

stengun:
Those two defines are not in posix standard, so either I redefine them if not present or set it to 0.

The reason behind using CBAUD is that I always reset the current baud rate before setting the new one just in case it happens to be unsupported, but since the switch case returns false for unsupported baud rates, flag reset is not needed.

CMSPAR on the other hand is tied to parity flag so its presence should be cleared when setting parity. Since it's mandatory to call set_commParameters after opening a serial port, it should be safe to avoid resetting CMSPAR on port open and doing it only when setting parity parameters.

PR on the way

PR #3827, opened 2022-11-15
Fix compilation on macOS builds

stengun:
Fix latest code compilation on macOs.

Due to the usage of non posix defines, macOS builds were broken. I reviewed and fixed those linux-isms.

No

No


maron2000:
Thank you. I haven't tested it but build is successful for macOS as well.

PR #3845, opened 2022-11-22
Update CHANGELOG

maron2000:
Update CHANGELOG.
Also includes not only mine but commits made by other contributors as well.


NebularNerd:
I shall try to remember to do this if I make any future pulls.

PR #3850, opened 2022-11-26
Fix bashisms, add quoting, clean perl scripts

roflcopter4:
Some time ago [I made this PR](https://github.com/joncampbell123/dosbox-x/pull/3392). I admit that the title was unnecessarily adversarial, but I was not at all prepared to see people overtly defend incorrectness. I'm afraid I quickly lost both interest in following up and a not insignificant portion of my will to live.

Well, here it is again, more or less remade by virtue of the many commits in the meanwhile. Mostly the same changes, some which fix bugs that make it impossible to compile this project on systems that don't have bash as a default shell. Others are less critical. All of it takes a surprisingly long time to do, so if I'm wasting said hours do feel free to say so.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -63,9 +63,10 @@ AC_CANONICAL_BUILD
 
 dnl Setup for automake
 AM_INIT_AUTOMAKE([foreign subdir-objects])
-AC_CONFIG_HEADER(config.h)
+AC_CONFIG_HEADERS([config.h])
 
 dnl Checks for programs.
+AC_PROG_AWK
 AC_PROG_MAKE_SET
 AC_PROG_CC
 AC_PROG_CPP
@@ -73,6 +74,8 @@ AC_PROG_CXX
 AC_PROG_OBJCXX
 AC_PROG_INSTALL
 AC_CHECK_TOOL(AR, ar)
+AC_PROG_MKDIR_P
+AC_PROG_OBJCXX
 AC_PROG_RANLIB
 AC_PREFIX_DEFAULT([/usr/local])
 
@@ -131,18 +134,44 @@ CXXFLAGS="$OLDCXXFLAGS"
 CPPFLAGS="$OLDCPPFLAGS"
 CFLAGS="$OLDCFLAGS"
 
-if test x$warn_cpp0x = x1; then
+if test "x${warn_cpp0x}" = 'x1'; then
     AC_MSG_ERROR([ Your compiler must support C++11 or better ])
 fi
 
 dnl Allow devs to force compile as C++11 as long as this code supports C++11 as a baseline
 AC_ARG_ENABLE(force-cxx11,AC_HELP_STRING([--enable-force-cxx11],[Force compile as C++11]),force_cxx11=yes)
+AC_ARG_WITH([cxx-version], 
+            AS_HELP_STRING([--with-cxx-version],
+                [Force compile as given C++ version (11, 14, 17, 20, 2a, 23). This option is for testing purposes only and if used will probably break the build. Caveat emptor.]),
+            [custom_cxx_version="$withval"], [])
 
-if test x$warn_cpp11 = x1; then
+if test -n "${custom_cxx_version}"; then
+    case "${custom_cxx_version}" in
+        2a|23)
+            CXXFLAGS="$CXXFLAGS -std=gnu++2a"
+            AC_MSG_WARN([Compiling with C++ version '-std=gnu++${custom_cxx_version}': this version is known not to work.])
+            ;;
+        17)
+            CXXFLAGS="$CXXFLAGS -std=gnu++17 -Dregister="
+            AC_MSG_RESULT([[Compiling with C++ version '-std=gnu++${custom_cxx_version}']])
+            ;;
+        11|14)
+            CXXFLAGS="$CXXFLAGS -std=gnu++${custom_cxx_version}"
+            AC_MSG_RESULT([[Compiling with C++ version '-std=gnu++${custom_cxx_version}']])
+            ;;
+        gnu++*|std++*|-std=*)
+            AC_MSG_ERROR([Please specify only a bare number for the C++ version (e.g. --with-cxx-version=17)])
+            ;;
+        *)
+            AC_MSG_ERROR([Unknown or invalid C++ version given "${custom_cxx_version}."])
+            ;;
+    esac
+    CFLAGS="$CFLAGS -std=gnu11"
+elif test "x${warn_cpp11}" = 'x1'; then
     AC_MSG_WARN([ Your compiler does not support C++14. Currently it is optional. At some point in the future, it will be required ])
     CFLAGS="$CFLAGS -std=gnu11";
     CXXFLAGS="$CXXFLAGS -std=gnu++11";
-elif test x$force_cxx11 = xyes; then
+elif test "x${force_cxx11}" = 'xyes'; then
     CFLAGS="$CFLAGS -std=gnu11";
     CXXFLAGS="$CXXFLAGS -std=gnu++11";
 else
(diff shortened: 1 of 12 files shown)


vs/libpng/scripts/makefile.linux
```diff
@@ -19,13 +19,14 @@ LIBSOREL=$(LIBSOMAJ).$(RELEASE)
...
 
 # Utilities:
-AR_RC=ar rc
-CC=gcc
-MKDIR_P=mkdir -p
-LN_SF=ln -sf
-RANLIB=ranlib
-CP=cp
-RM_F=/bin/rm -f
+AR ?= ar
```
skitt, line 22: All this is copied from `libpng` and should be fixed [there](https://github.com/glennrp/libpng).

vs/sdl/Makefile.minimal
```diff
@@ -1,9 +1,11 @@
 # Makefile to build the SDL library
 
-INCLUDE = -I./include
-CFLAGS  = -g -O2 $(INCLUDE)
-AR	= ar
-RANLIB	= ranlib
+INCLUDE += -I./include
```
skitt, line 3: This should be fixed [upstream](https://github.com/libsdl-org/SDL/blob/main/Makefile.minimal).

vs/sdl2/Makefile.minimal
```diff
@@ -1,9 +1,11 @@
 # Makefile to build the SDL library
 
-INCLUDE = -I./include
-CFLAGS  = -g -O2 $(INCLUDE)
-AR	= ar
-RANLIB	= ranlib
+INCLUDE += -I./include
```
skitt, line 3: This should be fixed [upstream](https://github.com/libsdl-org/SDL/blob/main/Makefile.minimal).

vs/sdl2/build-scripts/androidbuildlibs.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
skitt, line 1: This should be fixed [upstream](https://github.com/libsdl-org/SDL/tree/main/build-scripts).

vs/sdl2/configure.ac
```diff
@@ -4060,9 +4060,9 @@ fi
 SDLTEST_SOURCES="$srcdir/src/test/*.c"
 
 if test x$video_wayland = xyes; then
-    WAYLAND_PROTOCOLS=`cd $srcdir/wayland-protocols ; for p in *.xml ; do echo -n "\$p" |sed 's,\\.xml\$, ,g' ; done`
-    WAYLAND_PROTOCOLS_SOURCES=`for p in $WAYLAND_PROTOCOLS ; do echo -n "\\$(gen)/\$p-protocol.c " ; done`
-    WAYLAND_PROTOCOLS_HEADERS=`for p in $WAYLAND_PROTOCOLS ; do echo -n "\\$(gen)/\$p-client-protocol.h " ; done`
+    WAYLAND_PROTOCOLS=`cd $srcdir/wayland-protocols ; for p in *.xml ; do printf '%s' "\$p" |sed 's,\\.xml\$, ,g' ; done`
```
skitt, line 4063: This came from upstream, but upstream has since switched to CMake so…

skitt (changes requested):

Torinde:
So, there are:
- 4 items to be fixed upstream
- 1 item which would've been good to fix upstream, but upstream has more non-merged changes anyway
- 1 item which can be modified a little bit
- many others that are fine

Can this PR be merged if the "4 to fix upstream" items are excluded?

vs/sdl/Makefile.minimal
Torinde: I think SDL1 is not supported anymore upstream?
Does that mean it's OK to fix it directly here?

Torinde:
Can somebody make PRs at SDL and libpng for the items to be fixed upstream?

vs/sdl2/build-scripts/androidbuildlibs.sh
skitt: https://github.com/libsdl-org/SDL/pull/7440

vs/sdl/Makefile.minimal
skitt: I guess it does!

vs/sdl2/Makefile.minimal
Torinde: I can't find that file at https://github.com/libsdl-org ?

skitt: It’s on the SDL2 branch now: https://github.com/libsdl-org/SDL/blob/SDL2/Makefile.minimal

I’ve got a PR ready for submission.

configure.ac
```diff
@@ -170,35 +199,38 @@ dnl Optimize for speed by default
...
-if test x$enable_optimize != xno; then
-    if test x$enable_emscripten = xyes; then
-	    CFLAGS="$CFLAGS -Os"
-	    CXXFLAGS="$CXXFLAGS -Os"
+dnl
+dnl Never, EVER override the user's CFLAGS: it's Evil and Wrong.
+dnl
+dnl CFLAGS=["`echo $CFLAGS' ' | sed -e 's/-O[^ ]* //g'`"]
+dnl CXXFLAGS=["`echo $CXXFLAGS' ' | sed -e 's/-O[^ ]* //g'`"]
+
```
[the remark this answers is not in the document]
Torinde: @roflcopter4, can you please update or react on that one?
Skitt is preparing upstream PRs for the remaining items.

Torinde:
@skitt, how far are you with the upstream PRs?

Can the remaining edits be re-implemented?

vs/sdl2/Makefile.minimal
skitt: https://github.com/libsdl-org/SDL/pull/7508 (only `CFLAGS`; the other settings can be overridden on Make’s command-line anyway).

vs/libpng/contrib/testpngs/makepngs.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
skitt, line 1: https://github.com/glennrp/libpng/pull/472

vs/libpng/contrib/tools/intgamma.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
skitt, line 1: Addressed differently in https://github.com/glennrp/libpng/pull/472

skitt:
> @skitt, how far are you with the upstream PRs?
> 
> Can the remaining edits be re-implemented?

I’ve pushed the ones I care about most; I would like to improve the `libpng` Makefiles too, but I’m still on the fence about some of the changes there.

In general, Make variables can always be overridden from the command-line, so using `?=` everywhere isn’t _necessarily_ warranted and I wouldn’t feel comfortable pushing that on behalf of someone else. Referring to the original PR, if `AR` is a problem, `make AR=foo` will work (but might not be appropriate, that would require more discussion).

The `CFLAGS` handling in `configure.ac` should be redone; I can’t promise anything there however.

I don’t feel qualified to comment on the Perl changes.

The remaining shell script changes are debatable IMO; `x$foo` → `"x$foo"` is good (quoting variables), but really should be `"$foo"`, removing `x` from the expected value as well; `x$foo` → `x${foo}` is a matter of style; quoting static strings isn’t necessary. As far as the Atari `freetype` script is concerned, I would want to check the environment where it is expected to be used before making changes.

vs/sdl2/Makefile.minimal
roflcopter4: No they can't. If you assign a value in a makefile with `=` it will override whatever was set before or on the command line. Use `? =`.

skitt: @roflcopter4 what version of Make are you using?

```
$ cat > Makefile <<EOF
TEXT = hello

text:
	echo "$(TEXT)"
EOF

$ make text
echo "hello"
hello

$ make text TEXT=world
echo "world"
world
```

You need to explicitly `override` variable declarations in Makefiles if you want to override values set on the command line.

roflcopter4: This just isn't what always happens when one uses autotools (badly). The flags are supposed to be substituted in the `configure` stage, not when running make. If that doesn't work, for any reason, then the build is broken. And it doesn't work.

skitt: `Makefile.minimal` is only relevant if `configure` _isn’t_ used.

PR #3879, opened 2022-12-08
Fix build on OpenBSD

iyzsong:
Hello, this change will fix the build of DosBox-X on OpenBSD.
- Update `whereami` to its latest version.
- Change `configure.ac` to always link with `iconv` on OpenBSD.
- Fix some `NULL` issues for OpenBSD.
- Only call `X11_ErrorHandlerInstall` on Linux.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -647,6 +647,11 @@ dnl LIBRARY TEST: libiconv
 dnl This may be provided by glibc itself, -liconv may not be needed.
 AC_CHECK_HEADER(iconv.h,have_iconv_h=yes,)
 AC_CHECK_LIB(iconv, libiconv_open, have_iconv_lib=yes, ,-liconv)
+case "$host" in
+  *-*-openbsd*)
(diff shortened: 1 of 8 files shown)


Jookia:
Yeah my PR covered all this and more

Jookia:
Oops, not the bump for the where lib. That wasn't included.

Feel free to include all my commits if you want to take this on. I'm happy to answer anything about them here or on email.

iyzsong:
My change only allows it build on OpenBSD with system's SDL2 library, tested with autoconf, then `./configure --enable-sdl2; make`.

IIUC the changes from @Jookia allow it build with bundled SDL1 or SDL2.  What's the advantange using bundled SDL2 over a system one?

joncampbell123:
> IIUC the changes from @Jookia allow it build with bundled SDL1 or SDL2. What's the advantange using bundled SDL2 over a system one?

For SDL2, not much, except for small IME changes by @Wengier.

PR #3881, opened 2022-12-09
Fix build on FreeBSD and NetBSD

Torinde:
Manual copy of the edits from #3722. Due to subsequent changes, especially #3879, some of the edits were adjusted a little bit and also a few `#define ENABLE_IM_EVENT 1` were added for OpenBSD (I assume more has to be done for SDL1 to be OpenBSD buildable), such as in:
- vs/sdl/include/SDL_platform.h.default
- vs/sdl/include/SDL_platform.h
- src/output/output_surface.cpp
- configure.ac - **this one should be checked in detail**.

@Jookia @iyzsong please review thoroughly as I have no experience with any of that.

PR #3883, opened 2022-12-12
Improve support for FreeBSD, NetBSD and OpenBSD

iyzsong:
Hello, this change will improve support for dosbox-x on FreeBSD, NetBSD and OpenBSD.

- `./build`, `./build-debug`, `./build-sdl2`, `./build-debug-sdl2` works on NetBSD now.
- `./autogen.sh; ./configure --enable-sdl2; make` can be used to build dosbox-x againest external SDL2 library.
- Fixed dynrec on 32bit FreeBSD (Don't use 64-bit registers for pushl) 
  @Jookia, I can't test this..  Is it only for 32bit FreeBSD?
- Fixed dynrec on OpenBSD.

This change is rebased upon #3722 @Jookia and replaces #3881  @Torinde .

I tested builds on OpenBSD and NetBSD, x86_64 system.
* OpenBSD fails with bundled sdl1 due to audio driver error, its system one use sndio instead of bsdaudio, I have no plan to fix this.
* NetBSD pass all builds, but failed to enable dynrec, can be fixed later by other I think...
* FreeBSD not testsed.


Torinde:
Great. One last item: OpenBSD - mention in the CHANGELOG

PR #3884, opened 2022-12-13
Feature/gamelink

jwalt:
This patch set adds support for Grid Cartographer's Game Link protocol, which enables automapping in various old games like Ultima, Eye of the Beholder, and Wizardry.

Details are in src/gamelink/README.md, including a big caveat which I could only work around awkwardly.

It is only implemented for SDL2. Apart from that it is somewhat clean, but the working principle requires invasive changes. I tried to move as much as possible into separate code files.

The code is based on the reference client at https://github.com/hiddenasbestos/dosbox-gridc/ but it had to be modified a lot to become compatible with DOSBox-X


maron2000:
If your code is only valid for SDL2, you have to make it ignored when building it in SDL1.
Such as adding `#if defined(C_SDL2)` in the code.

jwalt:
> If your code is only valid for SDL2, you have to make it ignored when building it in SDL1. Such as adding `#if defined(C_SDL2)` and/or `#if C_GAMELINK` in the code.

Will do. While playtesting, I also found two bugs, expect an update in the next days.

jwalt:
The feature should now be pretty solid. Switching outputs works now, pause also works. The only other known issue is that there is no way to use the sdldraw menu when using the gamelink output. This can't be fixed easily, but the native menu works just fine.

jwalt:
I tried to fix VS compilation, but I have no VS available here, so this is probably not entirely right.

Also, I am still working on one extra feature: better address offset detection, because that is a pain. By now I come to believe it can't be fixed, because DOSBox-X and DOSBox SVN have too much differences in memory layout. So at least I want to make the process easier for end users.

kcgen:
@jwalt - are there any open-source cross-platform clients that support this?

jwalt:
I have no idea, I really just ported the changes from the dosbox-gridc repository. There is the All-Seeing Eye for the EoB game series that does memory snooping as well (and might have similar memory layout problems), but I have not tried it nor do I know what mechanism it uses. My personal motivation for porting this is GC really, since it is pretty generic and supports a lot of classic games.

Unless I forgot something, this should now compile fine on all platforms. On Windows it is fine. I tested it on Linux and it seems to work there as well. The address offset detection is now also sufficiently user friendly.

PR #3896, opened 2022-12-26
improve out-of-the-box compatibility of game link mode

jwalt:
The previous version of the game link code was pretty awkward to use due to big changes in memory layout. I found settings that almost fix this, removing the complicated setup procedure. This patch documents the required settings for maximum compatibility in the accompanying README file.

The patch keeps the "snoop" mode just in case some game is more sensitive to the remaining minimal memory differences. It does eliminate the hackish configuration via GC XML profiles, making this feature fully self-contained, not dependent on a specific client.

### DIFF
--- a/src/gamelink/README.md
+++ b/src/gamelink/README.md
@@ -35,7 +35,28 @@ gamelink master = true
 
 [render]
 scaler=xbrz
+
+[dos]
+hma                                              = false
+drive z expand path                              = false
+minimum dos initial private segment              = 0
+minimum mcb segment                              = 0
+minimum mcb free                                 = 187
+maximum environment block size on exec           = 32768
+additional environment block size on exec        = 83
+kernel allocation in umb                         = true
+shellhigh                                        = true
+
+[config]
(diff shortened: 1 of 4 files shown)


joncampbell123:
Just a suggestion, perhaps the game link external program should have a way to locate the emulator window and *ask* it where key addresses are, instead of requiring specific memory addresses.

As a good example of this, consider how long ago Nullsoft WinAMP allowed you to find it's main window and send well defined WM_USER messages to query song title and other useful functions.

jwalt:
Well, the trouble is that DOSBox itself doesn't really know what the memory addresses will be. Take Ultima 1 for example, U1 changes its memory layout depending on how long the environment variables are. There is no way for DOSBox to anticipate that. The "Load Address" workaround did work for the games I tested, but I wouldn't bet on this to always work. I think the current approach is more robust. The other pretty robust solution I've seen is to scan memory for known byte sequences and calculate everything relative to that. The current protocol supports that, if a client chooses to use that approach.

PR #3905, opened 2022-12-28
Fix build failure for Visual Studio builds in CI workflow

maron2000:
CI workflow builds fails on Visual Studio x86 and x64 builds when build tool is set to v142.
This PR fixes such failure.


joncampbell123:
I'm going to go ahead and update to VS2022 or whatever latest version is there so that later toolsets are available for DOSBox-X to use. Any objections?

maron2000:
Microsoft says that VS2022 can co-exist with older versions, so I think there are no obstacles in upgrading the build tool by just installing the Community edition for free.
The current code is compatible with v142(2019) and v143(2022), and not with v141(2017).

PR #3920, opened 2023-01-08
Fix PIC_AddEvent() interval (timer.cpp)

### LINKED ISSUE #3916: Crash - Days of Thunder
### Describe the bug

The game Days of Thunder is causing Dosbox-X to freeze after it exits.  There is a few seconds of delay before DosBox-X freezes, where it no longer accepts any input (including the menu or even resizing the window).  

Mainline versions of Dosbox aren't affected. The freeze isn't specific to one computer.

1. Launch "DOT.exe".  (Older versions of DosBox-X mention Loadfix, although that could be auto-configuration.) 
2. Press ESC to reach main menu.
3. Choose Quit to Dos. 

### Expected behavior

Avoiding the crash. 

Windows 10

v2022.12.26, MinGW SDL1, 64-bit

### Used configuration

_No response_

### Output log

_No response_

### Additional information

Additional versions tested, all having the freeze:
- v2022.12.26, MinGW SDL1, 64-bit
- v2022.12.26, MinGW SDL2, 64-bit
- v.0.83.22, Windows SDL2, 64-bit (January 2022) 
- v0.83.5, SDL1
- v2022.12.26, MinGW SDL1, 32-bit

maron2000:
The game Days of Thunder is causing Dosbox-X to freeze after it exits.
This is due to that the interval of PIC_AddEvent() is negative, and this PR fixes it.
However, this does not fix the root cause to make the interval negative value, so further fix may be required.

Fixes #3916

### DIFF
--- a/src/hardware/timer.cpp
+++ b/src/hardware/timer.cpp
@@ -324,7 +324,11 @@ static void PIT0_Event(Bitu /*val*/) {
 			LOG_MSG("PIT0_Event timing error %.6fms",err);
 #endif
 
-		PIC_AddEvent(PIT0_Event,pit[0].delay - (err * 0.05));
+        if(pit[0].delay - (err * 0.05) <= 0) {
+            err = 0;
+            //LOG_MSG("PIT0_Event delay value negative, reset to 0", err);
+        }
+        PIC_AddEvent(PIT0_Event,pit[0].delay - (err * 0.05));
 	}
 }
 


maron2000:
Commit 9da7ec8a12c1153c0703dfa5010f62782750a29c needs to be fixed in order to make Visual Studio builds successful.

NebularNerd:
I wonder if this will fix my old copy of Trivial Pursuit for DOS, very similar thing in that it just locks up when you use the 'Exit to Dos' function, look forward to trying this once it's merged. 🙂

maron2000:
@NebularNerd You can already try it by downloading the artifacts from the "Checks" menu found above.(or the following link)
https://github.com/joncampbell123/dosbox-x/pull/3920/checks

NebularNerd:
> @NebularNerd You can already try it by downloading the artifacts from the "Checks" menu found above.(or the following link) https://github.com/joncampbell123/dosbox-x/pull/3920/checks

Thanks, tested it just now and does not fix Trivial Pursuit from exiting to DOS, must be something else it's not liking.

grapeli:
@NebularNerd 
These are completely different cases. After finishing Days of Thunder  DOSBox-X goes into a loop causing full CPU load, unable to exit. Reset is not possible. The menu doesn't work. All that remains is, for example, to terminate the process with the command issued from the shell `pkill -QUIT dosbox-x`, which additionally ends with a crash. All this works fine in DOSBox-X  older than version 0.82.15. Works fine under dosbox-staging and dosbox-svn.

Trivial Pursuit (1987) hangs on exit to DOS. DOSBox-X works fine. You can use the menu, reset, reboot, exit. In addition, it behaves identically under dosbox-svn and dosbox-staging (this is very important).

NebularNerd:
> @NebularNerd These are completely different cases. After finishing Days of Thunder DOSBox-X goes into a loop causing full CPU load, unable to exit. Reset is not possible. The menu doesn't work. All that remains is, for example, to terminate the process with the command issued from the shell `pkill -QUIT dosbox-x`, which additionally ends with a crash. All this works fine in DOSBox-X older than version 0.82.15. Works fine under dosbox-staging and dosbox-svn.
> 
> Trivial Pursuit (1987) hangs on exit to DOS. DOSBox-X works fine. You can use the menu, reset, reboot, exit. In addition, it behaves identically under dosbox-svn and dosbox-staging (this is very important).

It's no big, saw the original thread and thought 'I wonder if TP is affected by this as well?' tested and found it was not. It's not something I play much and you can still exit DosBox via the hotkey or menus so it's by no means the end of the world.

joncampbell123:
I'll merge this pull request and then add an additional LOG call to that case so negative times can be noted and fixed. Sound good?

maron2000:
@joncampbell123 Yeah, I think we need some logging as you proposed.
We may need to supress the number of logs, so that it doesn't appear too frequently.

joncampbell123:
Hm... I'm noticing that the bug is that the PIT handler expects pit[0].start to advance automatically as it would mode 2 and 3, but Days of Thunder leaves PIT timer 0 (system timer) in mode 1, which means that when you exit, the system timer wouldn't even tick anymore until something else put it back into mode 2, and start doesn't advance, so the computed "error" just gets bigger and bigger and bigger endlessly.

Which probably means if you ever run this game on a PC/XT without a RTC of any kind, the date and time will magically stop ticking forward once you run this game :)

PR #3922, opened 2023-01-08
Make code able to be built with Visual Studio 2017 (build tool v141)

maron2000:
The current code cannot be built with Visual Studio 2017 (build tool v141) since #3744.
This PR makes it able to be built by VS2017 again.
In addition, modified the build instructions for Visual Studio.


joncampbell123:
I was planning to update the build tool and update to VS2022 or so.

Since there is an interest in VS2017 still, I'll hold back on that for now.

maron2000:
@joncampbell123 I believe 2017 is required for XP builds only, so Ithink we should make a build workflow dedicated for XP.
My patch shouldn't affect building on 2019 and 022.

Issue #3926, opened 2023-01-10
dosbox-x-vsbuild-win32-20221226190454.zip can't run on windows xp


leoxxx:
### Describe the bug

dosbox-x-vsbuild-win32-20221226190454.zip can't run on windows xp

### Steps to reproduce the behaviour

Click the exe, then can't run.

### What operating system(s) this bug have occurred on?

windows xp

### What version(s) of DOSBox-X have this bug?

vsbuild-win32-20221226190454

### Have you checked that no similar bug report(s) exist?

- [X] I have searched and didn't find any similar bug report.

### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

Torinde:
I don't know if it matters, but you may try both SDL1 and SDL2...

leoxxx:
Try them, all can't run.

maron2000:
This is actually not a bug.
Visual Studio (2017 and after) and MinGW dropped support for Windows XP.
You can still run Visual Studio builds by applying a patch to the header.
Download the patcher and copy it to where your DOSBox-X executable is, and run it.
Remember that XP support is **obsolete and/or not recommended** and the patcher does NOT guarantee that further DOSBox-X releases can be run on XP. 

```PatchPE dosbox-x.exe```

https://github.com/joncampbell123/dosbox-x/blob/master/contrib/windows/installer/PatchPE.exe

leoxxx:
2022.08 can run on windows xp. I think you can save this version.

BridgeHeadland:
Do the patcher make dosbox-x-vsbuild-win64-20230109003356.zip\bin\x64\Release SDL2 to be run on XP? It's the only version I have right now.

BridgeHeadland:
Do the patch make the current dosbox-x-vsbuild-win32-20230110063138 to be run on XP?

maron2000:
@Torinde Before updating the documentation, did the patch work?

Torinde:
I don't have WinXP at hand now - maybe @leoxxx can test it?

maron2000:
The patch didn't work on 32-bit XP, however, 2022.09 (**not** 08) version which is possibly built with VS2017 worked.
So maybe it is due to building on VS2019/2022 since PR #3744 dated 15th Sep which broke compatibility with VS2017.
I already put a PR (#3922) to make it able to be built with VS2017, so after it is merged, we can try the latest code built on VS2017.

[...]

PR #3945, opened 2023-01-20
mention WinXP VS2017, LOADLIN latest

Torinde:
- automatic building of LOADLIN portable (last seen [0.83.25 2022-05-01](https://github.com/joncampbell123/dosbox-x/releases/tag/dosbox-x-v0.83.25)). If it's decided that this won't be build with every release, then INSTALL.md should link to the latest one.

_Originally posted by @Torinde in https://github.com/joncampbell123/dosbox-x/issues/3926#issuecomment-1386537129_

Mention in INSTALL.md the WinXP VS2017 builds and the latest version having LOADLIN.
Of course, would be also good if somebody can make the LOADLIN be automatically build like the rest of the versions.


maron2000:
I suggest not using version numbers such as 2017 & 2022, since casual users will not find those number when downloading.
The current build is built using 2019, but maybe subject to change.

TL;DR
Visual Studio 2017 builds  -> Visual studio XP builds
Delete 2022, no version number required aince it is subject to change.

Torinde:
Done.

PR #3965, opened 2023-01-24
Sync dh fpu state with normal fpu

finalpatch:

This is the same change as in https://github.com/dosbox-staging/dosbox-staging/pull/2248

Basically this fixes the fpu state out of sync problem when dynamic core calls normal core to handle self modification code.

No

No

## Additional information

I tested this with dos navigator:

* Set core to dynamic
* Run Dos Navigator
* Type some rubbish on the command line and press enter
* Repeat the above step
* After several times, dos navigator will crash (the entire dosbox app will crash in other dosbox versions but in dosbox-x I guess it has better error handling so only the program in it crashes).

I used mingw build because the visual studio build uses 64-bit floats and won't run dos navigator.

### DIFF
--- a/src/cpu/core_dyn_x86.cpp
+++ b/src/cpu/core_dyn_x86.cpp
@@ -315,7 +315,7 @@ Bits CPU_Core_Dyn_X86_Run(void) {
 			fpu.sw = dyn_dh_fpu.state.sw;
 			const uint8_t* buffer = &dyn_dh_fpu.state.st_reg[0][0];
 			for(Bitu i = 0;i < 8;i++){
-				memcpy(&fpu.p_regs[STV(i)], buffer + i * 10, 10);
+				memcpy(&fpu.p_regs[i], buffer + i * 10, 10);
 			}
 		}
 		~auto_fpu_sync () {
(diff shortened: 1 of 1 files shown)


rderooy:
As there was a mention of Vogons that this patch causes a significant performance impact on Win98, I ran the Processor benchmarks of WinBench96 on Win98SE before and after the patch.
guest os_006

LiFengSC:
Regarding the performance hit,  I believe the emulation first needs to be correct, then fast.  This PR and the next one (https://github.com/joncampbell123/dosbox-x/pull/3969) fixe a legit bug that causes certain programs to not function properly. 

Given the focus of the X project being "accurate" and to be "a platform for running DOS applications",  being correct is definitely more important here than other dosbox forks.

An example of production application not functioning properly is that on x86 dynamic core, debuggers won't be able to see FPU states when debugging code.  This is now fixed.

kcgen:
>  That said, the root cause of the bug and the need for "syncing the 2 fpu states" is because we have 2 fpu states in the first place, which I suppose is due to historical reason. I believe we have an opportunity to refactor the fpu emulation code and unify the 2 states into a single, lean data structure and this will eliminate the need for the syncing and get the speed back, not to mention the code will be shorter and simpler.

@LiFengSC -> Nice! This is exactly what @finalpatch also mentioned - Is this something you want to tackle? (Or are you @finalpatch ?)

finalpatch:
> 
> @LiFengSC -> Nice! This is exactly what @finalpatch also mentioned - Is this something you want to tackle? (Or are you @finalpatch ?)

Sorry that was me.  I accidentally commented through my work account.  Rewriting the fpu code is a fairly big chunk of work and has some risk. At this point I feel I'm not super motivated to do this, especially when things are now working.

kcgen:
hey np! maybe it's something you'll want to chip away at, even tiny bits at a time.

It's awesome having you part of the crew, and helping make DOSBox (as a whole) better!

PR #3969, opened 2023-01-26
More dynamic FPU fixes

finalpatch:
After some testing, I believe the memory layout that I assumed on fnsave was incorrect.  This fixes it. 

This also adds protection to all the other instances where CPU_Core_Normal_Run() is called in the x86 dynamic core.  After this Turbo Debugger is able to correctly display FPU updates on the dynamic core.

An optimization is added to avoid most of the copies.

### DIFF
--- a/src/cpu/core_dyn_x86.cpp
+++ b/src/cpu/core_dyn_x86.cpp
@@ -294,6 +294,28 @@ static void dyn_restoreregister(DynReg * src_reg, DynReg * dst_reg) {
 
 extern int dynamic_core_cache_block_size;
 
+class auto_fpu_sync {
+public:
+	auto_fpu_sync () {
+		FPU_SetTag(dyn_dh_fpu.state.tag);
+		fpu.cw = dyn_dh_fpu.state.cw;
+		fpu.sw = dyn_dh_fpu.state.sw;
+		const uint8_t* buffer = &dyn_dh_fpu.state.st_reg[0][0];
+		for(Bitu i = 0;i < 8;i++){
+			memcpy(&fpu.p_regs[STV(i)], buffer + i * 10, 10);
+		}
+	}
+	~auto_fpu_sync () {
+		dyn_dh_fpu.state.tag = FPU_GetTag();
+		dyn_dh_fpu.state.cw = fpu.cw;
+		dyn_dh_fpu.state.sw = fpu.sw;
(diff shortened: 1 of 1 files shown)


finalpatch:
@rderooy I implemented an optimization that is able to avoid most of the copies. Could you retest with your win95 benchmark?

finalpatch:
@kcgen  I implemented an optimization that should help reduce state copies in this PR.  I think you'll want it in your repo too.   I have tested this change with everything that I know that could verify the bug:
* dos navigator (both real and dpmi) run command line
* a borland pascal example program
* turbo debugger fpu window

rderooy:
I ran the benchmark again, and got some weird results...
guest os_009

finalpatch:
@rderooy LOL that's quite unexpected.   Do you notice anything wrong?

This code does pass all the cases that I used to have issues with, and is able to pass the artificially constructed test program here https://gist.github.com/finalpatch/91ae4449875c3cf217327bfdcbeab0c5

rderooy:
> @rderooy LOL that's quite unexpected. Do you notice anything wrong?
> 
> This code does pass all the cases that I used to have issues with, and is able to pass the artificially constructed test program here https://gist.github.com/finalpatch/91ae4449875c3cf217327bfdcbeab0c5

Thinking about it, I think that for the first two runs, I switched away from the process. While for the third run, I left it as the active process. Testing now. It takes ~25min to run one test, so it takes a while...

rderooy:
It seems that that switching focus away was the culprit. Here there are two runs. the first is actually with the old fix (I labelled it wrong, and it does not give you the opportunity to fix it later), the v2 is with the new one.

guest os_012

finalpatch:
@rderooy sorry I'm a bit confused.  Are you saying the change in this branch is actually running faster than not only the previous PR, but also faster than before the whole FPU fix thing?

rderooy:
@finalpatch for the initial benchmark runs that I did, I had switched away from the DOSBox-X window as they take a LONG time to run (~25min for the 16bit benchmark, and another 20min for the 32bit benchmark). This seems to have a significant performance impact, and therefore those runs cannot be relied on.

With the later benchmark runs, I left the system alone, so with DOSBox-X as the active process for the run. But even then there is some variance in the runs.

finalpatch:
Hi @joncampbell123,  I'm happy with the state of this PR.  Feel free to merge when you are ready.

PR #3985, opened 2023-02-05
Support DPMI idle (INT 2FH Function 1680H) and DOS idle (INT 28H)

finalpatch:

When DOSBOX is set to run at high cycles or max cycles,  it will spin the host CPU very hard even when there is nothing to do (eg. waiting for user input).  This makes tasks such as text editing and programming rather unpleasant, because we either have to set for a low cycle count, sacrificing responsiveness and compilation time, or burn lots of energy and make the host system very hot.

This adds support for DOS idle (INT 28H) and DPMI idle (INT 2FH Function 1680H)

Programs that support these features would periodically call these interrupts to give up their time slice to the host system (on real hardware, the DPMI host, and on DOSBOX, the DOSBOX app).

Before this change, when on the command line prompt (real mode) or when editing text in Free Pascal IDE (protected mode), and cycles set to max, the host CPU spins up to 4.0GHz and total CPU usage at 10%. 

After this change, when performing the same task, the host CPU stays at 2.5GHz, 2% usage.  This makes editing programs nice and cool, while still being able to compile programs quickly. 

The way it works is that when staying at command line prompt, the command shell inserts INT 28H calls to the keyboard reading loop when there is no input, this lets DOSBOX know that it's idle.  Similarly, Free Pascal's protected mode text editor would periodically call both idle functions to give up CPU time to host ( [code](https://github.com/fpc/FPCSource/blob/9f22623459866915d8769bd2e6cd9c8a75357c6e/packages/fv/src/drivers.inc#L830) )

So far, I haven't noticed any side effects.  The risk of this breaking games is low because games do not call these software interrupts and they normally do not use the DOS blocking keyboard function to read inputs because this would stop their screen updates.

I don't use Windows inside DOSBOX, so this change has **NOT** been tested with guest Windows system. 

## Additional information

Changes:

* Add INT 28H handler and install it. [reference](http://www.techhelpmanual.com/567-int_28h__dos_idle.html)
* Add implementation to INT 2FH/1680H handler and return al = 0 (means supported) [reference](http://www.delorie.com/djgpp/doc/dpmi/api/2f1680.html)
* Both implementations use the HLT instruction implementation in cpu.cpp to set the remaining cycles to 0 and enter interruptible sleep.
* Change the keyboard blocking read function to wait for available input first and call the new idle function in the wait loop

Update:
I just realised this has been attempted before in [this thread ](https://www.vogons.org/viewtopic.php?t=25117) .  I'm going to try if I could try some of the ideas there to make this PR better. 

Update:
Yeah, I replaced my INT 28H handler with the original commented out code.


joncampbell123:
It might be wise to put this behind a dosbox.conf enable, which imho should be on by default, but allow to turn it off in case it interferes with interrupts or timing in games or hardware emulation.

Also just to point out that in pure DOS mode that call doesn't do anything because there's no multitasking. But, this might be a nice feature for those who want DOSBox-X idle when the program is calling idle regardless.

I do know from experience that the particular INT 2F call is very important to keep CPU load down for your task if you're a DOS program running under NTVDM.EXE in Windows XP.

finalpatch:
You are right, I'll add a config setting for this.  

I'm not aware of Allegro calls this too.  Could you point me to the source?

kcgen:
Cross-linking @feralchild64's discovery that some commands launched in DOS Navigator's CLI still can crash it:

https://github.com/dosbox-staging/dosbox-staging/issues/2265

Quote:

> _"I suspect the incompatibility is related to using CALLBACK_Idle or similar mechanism by interactive commands, waiting for user input."_

Will be very curious to see if this solves that, too.

PR #3988, opened 2023-02-05
Add CI workflow to build Installer for Windows XP

maron2000:
The current installer for Windows does not include binaries that are able to run in Windows XP.
This PR adds some script to gather the relevant binaries and compile it in an installer for Windows XP users. 
I haven't thoroughly tested it but the majority of the code is based on the existing 32-bit installer so I don't expect critical issues.
However if there are any flaws, you can always try the portable zips.

## Additional information

XP_Installer

Edit: You can test the installer by downloading from the following link
https://github.com/joncampbell123/dosbox-x/suites/10787704695/artifacts/542303627

### DIFF
--- a/.github/workflows/vsbuild_xp.yml
+++ b/.github/workflows/vsbuild_xp.yml
@@ -63,10 +63,15 @@ jobs:
           mkdir -p $top/package/shaders
           mkdir -p $top/package/glshaders
           mkdir -p $top/package/languages
+          mkdir -p $top/vs-bin
           cp $top/bin/Win32/Release/dosbox-x.exe $top/package/dosbox-x_XPx86_SDL1.exe
           cp $top/bin/Win32/"Release SDL2"/dosbox-x.exe $top/package/dosbox-x_XPx86_SDL2.exe
           cp $top/bin/x64/Release/dosbox-x.exe $top/package/dosbox-x_XPx64_SDL1.exe
           cp $top/bin/x64/"Release SDL2"/dosbox-x.exe $top/package/dosbox-x_XPx64_SDL2.exe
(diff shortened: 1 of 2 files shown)


Torinde:
Installer runs on WinXP x86, WinXP x64, Vista x86, Vista x64.
Regular Win32 installer runs on Vista x86, Vista x64. Maybe also on WinXP x86 (but obviously the non-XP builds from it won't run)
Regular Win64 installer does NOT run on Vista x64.
image

So, what fix from the XP installer allows it to run properly on Vista x64 - the "disable cache detection"? Can that (or another fix) be implemented also for the regular win64 installer?

Also, the website is structured as follows:
image
So, either:
- has to be changed to:
XP Setup (32-bit and 64-bit)
32-bit Setup (Vista+) | 64-bit Setup (Vista+)
- installers for XP and Vista+ to be combined. Or maybe better - all builds to be in a single installer (32-bit and 64-bit)?

maron2000:
I think it's now fairly OK to merge this PR.

.github/workflows/vsbuild_xp.yml
```diff
@@ -96,3 +99,118 @@ jobs:
...
+        with:
+          name: dosbox-x-winXP-${{ env.timestamp }}-setup.exe
+          path: contrib/windows/installer/dosbox-x-winXP-*.exe
+      - name: Upload XP build installer release version
+        uses: softprops/action-gh-release@v1
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          files: |
+            contrib/windows/installer/dosbox-x-winXP-*.exe
+      - name: Clean cache
```
Dobby233Liu, line 206: I don't get it. Caches are for data that can be repurposed between multiple runs of workflows. You can download artifacts from other jobs with [`actions/download-artifacts`](https://github.com/actions/download-artifacts).

PR #3982, opened 2023-02-04
Fixes for Haiku OS

amol-:
Provide necessary fixes to make DosBox-X correctly compile and start on Haiku.
There are still some issues to be addressed, but I was able to start and play Warcraft2 on it so it's functional enough to be usable.

PR #3997, opened 2023-02-10
Fix broken MinGW build

maron2000:
This PR fixes the broken MinGW builds since #3982.

### DIFF
--- a/src/dos/drive_local.cpp
+++ b/src/dos/drive_local.cpp
@@ -19,7 +19,10 @@
  *  With major works from joncampbell123 and Wengier
  */
 
+#if defined(HAIKU)
 #include "config.h"
+#endif // defined(HAIKU)
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>


src/dos/drive_local.cpp
```diff
@@ -19,7 +19,10 @@
  *  With major works from joncampbell123 and Wengier
  */
 
+#if defined(HAIKU)
 #include "config.h"
```
amol-, line 23: This was added exactly because otherwise ``HAIKU`` wasn't defined, so probably the right fix would be to remove the ``config.h`` and understand why ``HAIKU`` isn't defined but the other OSes are

amol-: https://github.com/joncampbell123/dosbox-x/pull/4002 should fix MinGW without breaking HAIKU

maron2000: @amol- Thanks for your confirmation.
 I see that MinGW build is back with your fix  so I will withdraw this request.

PR #4054, opened 2023-03-06
Add macOS Universal Binary build support

### LINKED ISSUE #3237: Build universal binaries for macOS?

### Have you checked that no other similar feature request(s) already exists?

### Is your feature request related to a problem? Please describe.

_No response_

### What you want

Is it possible to build universal binaries for the Mac, using some variation of the method here:

https://thebreakfastpost.com/2021/01/05/on-macos-arm64-and-universal-binaries/

For my build, I build separately on Intel and ARM, and now realize that I can use the lipo command to combine the two binaries into a single binary.

If this could be added to the standard code, there could be a single macOS release instead of two separate one as there are now.

EDIT: I've now experimented with building for ARM on Intel, with no success, after trying various options for configure (host=various ways of specifying arm64, adding target= lines to the CC exports, and modifying the build scripts for the components in VS (libpng, zlib, etc.). It seems to be possible, but beyond my abilities.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

diddledani:
Add support to build both Intel and arm64 executables on an Apple Silicon Mac and combine them into a single App Bundle to simplify distribution of macOS builds.

Fixes #3237

No new features within DosBox-X itself, but does bring the new build-time feature to be able to create a Universal Binary with both Intel and Apple Silicon supported by a single App Bundle.

I do not believe it does. The changes are to the build scripts only, so the functionality of DosBox-X *should* be identical.

However, I have only been able to test the Intel portion of the Universal Binaries I have produced by running through Rosetta 2 as I only have one Mac, which has an m1 CPU.

Running through Rosetta 2 required that I set the CPU configuration option `core` to `normal` at runtime otherwise it would crash with a Segmentation Fault when starting DOOM (my test game) - I suspect this is a byproduct of running through Rosetta 2, but this needs to be verified by testing on an Intel-based Mac.

## Additional information

I have uploaded my built App Bundle of an SDL2 build to Google Drive (because it is too big to attach to this PR) so that somebody can test on a native Intel Mac: [dosbox-x.app.zip](https://drive.google.com/open?id=13udgs5YoLVJlWxazADRXf68bR3Go7wsP&authuser=diddledanillew%40gmail.com&usp=drive_fs)

The Universal build method will *only* succeed on an Apple Silicon Mac. If you build on an Intel Mac you will receive a single-architecture build for Intel only due to the requirement for a working arm64 Homebrew prefix for the build process.

You need to have both an arm64 *and* Intel prefix of Homebrew for a Universal build, because we need the Homebrew-provided shared libraries for both architectures.

The build process for a Universal Binary App Bundle is as follows...

For SDL2:

```bash
$ ./build-macos-sdl2 universal
$ make dosbox-x.app
```

Or for SDL1:

```bash
$ ./build-macos universal
$ make dosbox-x.app
```

The `build-macos` and `build-macos-sdl2` scripts will *only* build both architectures on an Apple Silicon Mac. You *must* add the `universal` parameter to trigger them to build both architectures even when you are on an Apple Silicon system. *Without* the `universal` parameter the scripts should build the same way they did before this PR so that potential breaking of anyone packaging elsewhere is minimised :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -320,8 +320,8 @@ __pycache__/
 ###############################################################################
 
 *~
-dosbox.app/
-dosbox.app/*
+dosbox-x.app/
+dosbox-x.app/*
 dosbox.icns
 dosbox.iconset/
 dosbox.iconset/*
@@ -351,6 +351,10 @@ Makefile
 config.h
 config.status
 stamp-h1
+
+# macOS Universal Binary intermediates
+/src/dosbox-x-arm64
+/src/dosbox-x-x86_64
 
 doxygen
 doxygen/*
@@ -410,6 +414,9 @@ vs/libpng/linux-build/*
 /vs/libpdcurses/wincon/firework.exe
 .vscode/settings.json
 vs/freetype/objs
+
+# Ignore mach-o-matic
+/src/tool/mach-o-matic
 
 # Ignore flatpak stuff
 .flatpak-builder/
--- a/BUILD.md
+++ b/BUILD.md
@@ -18,7 +18,7 @@ The four major operating systems and platforms of DOSBox-X are:
 
 2. Linux (with X11) 64-bit x86/x64, and on a Raspberry Pi 3/4
 
-3. macOS (Mac OS X) recent version, 64-bit Intel and ARM-based
+3. macOS (Mac OS X) recent version, 64-bit Intel, ARM-based, and Universal
 
 4. DOS (MS-DOS 5.0+ or compatible)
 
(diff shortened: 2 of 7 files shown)


diddledani:
Just want to ping this, because I see that I broke the CI but you merged it anyway. Is the broken CI ok to bypass? Would you like me to follow-up with another PR to appease the CI?

It looks like it is my poor knowledge of Makefiles that broke it - I put the arm64 and x86_64 executables as dependencies for the `dosbox-x.app` rule, but when built on CI the arm64 executable isn't created so it refuses to run the rule. I'm not sure how to allow the rule to run when only the x86_64, or only the arm64 executable are available - it should build a single-architecture `dosbox-x.app` bundle if only one architecture build is present in the tree if I've got the scripting within the rule correct, so it's only the dependencies of the rule that I got wrong (I think).

Torinde:
In my opinion: yes, please correct it. (of course, I'm not the maintainer)

From [what I see](https://dosbox-x.com/devel-build.html) - due to that currently there are no nightly builds for macOS, and even if at release time there is a way to bypass the problem - I think it's good to correct it, so that macOS users can get nightly builds between the releases...
image

Allofich:
It would also be good to be corrected because if even one CI build fails, the whole thing gets a red "x" by it. You have to then open it up to check whether one of the other builds failed or it was just the one(s) that were already failing.

Also, it may depend on settings but I get an email every time a CI build fails on one of my repositories, so probably @joncampbell123 is getting an email every time the CI build fails.

PR #4076, opened 2023-03-14
Attempt fix for macOS CI fail following PR#4054

diddledani:
PR #4054 broke the macOS CI for pull requests and the nightly builds. As a follow-up requested in the PR this is an attempt to fix it.

Regression in CI from PR #4054

It adds an optional \`./configure --enable-universal\` parameter to define a variable for use in the `Makefile` when executing `make dosbox-x.app`. This is to allow us to determine which architecture executables to depend upon in the \`dosbox-x.app\` rule in the \`Makefile\`.

I do not believe so.

## Additional information

Changes are:

* Add a parameter to `configure.am` to define a build-time variable for use in `Makefile.am` to specify the executables we will be depending upon for the `dosbox-x.app` rule.
* Fix an issue in the first library scan section of the `dosbox-x.app` rule in the `Makefile` causing a failure to build on an Apple Silicon mac when there is no `x86_64` binary present.
* Fix failure to build SDL1 variant after a previous SDL2 build attempt by deleting `vs/sdl2/linux-host` before attempting the SDL1 build.
* Use `make clean` instead of `make distclean` in the build scripts' `do_cleanup` function to not clean absolutely everything.
* Make sure we remove the achitecture-specific copies we create of the `dosbox-x` executable before building, as these are meant to be used for the universal binary creation with `lipo` and the presence of an old copy will cause the `dosbox-x.app` bundle to potentially contain mismatched copies for each architecture, or contain an architecture that is not intended to be included.
* Add a `distclean-local` rule to `src/Makefile.am` to remove the architecture-specific temporary copies of the `dosbox-x` executable when running `distclean` as these shouldn't remain in the tree.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -3,9 +3,17 @@
 EXTRA_DIST = autogen.sh
 SUBDIRS = src include
 
+if MACOSX
+if MACOS_UNIVERSAL
+MACOS_BINARIES=src/dosbox-x-arm64 src/dosbox-x-x86_64
+else
+MACOS_BINARIES=src/dosbox-x-$(shell uname -m)
+endif
+endif
+
 .PHONY: dosbox-x.app
 
-dosbox-x.app: src/dosbox-x-arm64 src/dosbox-x-x86_64 contrib/macos/dosbox.icns src/tool/mach-o-matic
+dosbox-x.app: $(MACOS_BINARIES) contrib/macos/dosbox.icns src/tool/mach-o-matic
 	rm -Rfv dosbox-x.app
 	mkdir dosbox-x.app
 	mkdir dosbox-x.app/Contents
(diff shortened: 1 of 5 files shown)


diddledani:
Please don't merge this until we're sure the CI is correctly fixed by the changes.

diddledani:
OK, I think this is good to go then. Can I get your thoughts, @joncampbell123, @Torinde, and @Allofich, please?

Torinde:
From what I see (but I can't test it) it corrected the macOS flow:
image

Can you add the universal and arm-only binaries to the automatically-generated build artifacts? Currently only the x86-64 one is shown (I assume your initial PR "makes" those somewhere, but they aren't shown in the list? or your PR only makes it possible to have arm64/universal builds, but they can't be generated automatically and it's useful only at manual-release time?)

maron2000:
I downloaded the artifact (x86_64 build) and ran in on an Intel Mac.
I just only started the binary but it seems to be working.

diddledani:
Until GitHub provides an apple silicon build environment (they're scheduling for initial public availability in the October through to December timeframe right now) we cannot generate a universal binary within CI via GitHub Hosted Runners. Until that time we still need to generate the universal build offline, but these two PRs have made it much simpler to do so.

The reason we require the build to run on an Apple Silicon system is because we require a functional Homebrew environment for *both* Intel *AND* Apple Silicon on the build system to be able to build both architectures successfully within a single run. This is only possible when using Rosetta.

The alternative would be to build everything, that we currently use Homebrew to provide, ourselves for each architecture. It would require verifying each of those dependencies can be cross-built for Apple Silicon on an Intel Mac. This would, however, allow us to build the universal bundle on an Intel Mac within CI.

Allofich:
My only concern was about the CI build failing, so since that's working now it looks good to me.

PR #4093, opened 2023-03-19
Retarget VS solution to latest

Allofich:

Retargets the Visual Studio solution to v143, which is the latest right now.

No.

No.


joncampbell123:
Before I merge this...

Past experience says that VS2019 only supports up to v142. This will break compatibility with VS2019 which I understand some who compile from source are still using.

Anyone willing to drop VS2019? Please discuss.

Allofich:
OK, well it's not a big deal to me if we keep it at v142, then.

PR #4107, opened 2023-03-26
- Should fix compilation of src/hardware/parport/parport.cpp with Vis…

dbjh:
…ual Studio ARM32 and Visual Studio ARM64.

- I say should because I do not have access to Visual Studio ARM32 nor Visual Studio ARM64. As far as I can tell this is a proper fix.

Compilation of parport.cpp caused an error with Visual Studio ARM32 and Visual Studio ARM64, because parport.cpp was referring to the class CDirectLPT if the constant C_DIRECTLPT was non-zero. However, in my previous pull request I had made CDirectLPT only available on platforms where it can actually work. CDirectLPT cannot work on Windows/ARM32 and Windows/ARM64.
This pull request introduces the constant HAS_CDIRECTLPT that has the value 1 if the class CDirectLPT is available. Using this constant is much less error prone than repeating the preprocessor logic that determines whether CDirectLPT should be compiled.

No.

No.

## Additional information

Thank you very much for accepting my previous pull request!

### DIFF
--- a/src/hardware/parport/directlpt.cpp
+++ b/src/hardware/parport/directlpt.cpp
@@ -18,17 +18,8 @@
 
 #include "dosbox.h"
 
-#if C_DIRECTLPT
-/*
-  For MinGW and MinGW-w64, _M_IX86 and _M_X64 are not defined by the compiler,
-  but in a header file, which has to be (indirectly) included, usually through a
-  C (not C++) standard header file. For MinGW it is sdkddkver.h and for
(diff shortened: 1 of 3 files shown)


dbjh:
Although this fix is in line with what is done elsewhere in the code and I am generally of the opinion that code should be as independent on the build system as possible, it is reasonable to ask: Why is C_DIRECTLPT non-zero for Visual Studio ARM32 and Visual Studio ARM64?

Issue #3819, opened 2022-11-10
Unable to compile in alpine since v0.83.22


bestpika:
### Question

I have used alpine to compile dosbox-x.
However, since v0.83.22, I can't compile it.
Before the compile failed I saw the following message in the shell.
```shell
dos_programs.cpp: In member function 'virtual void SETCOLOR::Run()':
dos_programs.cpp:7614:22: warning: unused variable 'colornul' [-Wunused-variable]
 7614 |                 bool colornul = IS_VGA_ARCH && (altBGR1[i].red > 4 || altBGR1[i].green > 4 || altBGR1[i].blue > 4) && rgbcolors[i].red < 5 && rgbcolors[i].green < 5 && rgbcolors[i].blue < 5;
      |                      ^~~~~~~~
dos_programs.cpp:7645:22: warning: unused variable 'colornul' [-Wunused-variable]
 7645 |                 bool colornul = IS_VGA_ARCH && (altBGR1[i].red > 4 || altBGR1[i].green > 4 || altBGR1[i].blue > 4) && rgbcolors[i].red < 5 && rgbcolors[i].green < 5 && rgbcolors[i].blue < 5;
      |                      ^~~~~~~~
dos_programs.cpp:7659:18: warning: unused variable 'colornul' [-Wunused-variable]
 7659 |             bool colornul = IS_VGA_ARCH && (altBGR1[i].red > 4 || altBGR1[i].green > 4 || altBGR1[i].blue > 4) && rgbcolors[i].red < 5 && rgbcolors[i].green < 5 && rgbcolors[i].blue < 5;
      |                  ^~~~~~~~
dos_programs.cpp: In function 'void Add_VFiles(bool)':
dos_programs.cpp:8258:5: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
 8258 |     if (IS_DOSV)
      |     ^~
dos_programs.cpp:8261:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
 8261 |  VFILE_RegisterBuiltinFileBlob(bfb_EDLIN_EXE, "/DOS/");
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../libs/libchdr/chd.h:49,
                 from cdrom.h:38,
                 from dos_programs.cpp:44:
../libs/libchdr/coretypes.h: At global scope:
[...]

PR #4114, opened 2023-03-29
Start using nullptr instead of NULL or 0

Allofich:

"nullptr" warnings in static analyzer.

No.

No.

## Additional information

I've been using the static analyzer SonarLint to tidy up the DOSBox-X code in PRs. Many of the warnings issued for every file I look at are suggestions to use `nullptr` instead of `NULL` or `0`. Here is the CPP Core Guidelines' take on the issue: https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#es47-use-nullptr-rather-than-0-or-null

If nothing else, it would be nice for me to have those warnings taken care of so they don't get in the way of noticing others.

For this PR I converted a few files, not all of them. If this change is OK, I will continue to convert over files as I look at them to fix warnings.

### DIFF
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -67,7 +67,7 @@ class DOS_DTA;
 
 class DOS_File {
 public:
-    DOS_File() :flags(0) { name = 0; attr = 0; date = 0; drive = 0; refCtr = 0; open = false; time = 0; hdrive = 0xff; newtime = false; };
+    DOS_File() :flags(0) { name = nullptr; attr = 0; date = 0; drive = 0; refCtr = 0; open = false; time = 0; hdrive = 0xff; newtime = false; };
 	DOS_File(const DOS_File& orig);
 	DOS_File & operator= (const DOS_File & orig);
 	virtual ~DOS_File(){ delete [] name;};
@@ -92,7 +92,7 @@ public:
 	virtual void 	LoadState( std::istream& stream, bool pop );
     virtual void    Flush(void) { }
 
-	char* name = NULL;
+	char* name = nullptr;
 	uint8_t drive = 0;
 	uint32_t flags;
 	bool open;
@@ -279,7 +279,7 @@ private:
 
 	CFileInfo*	dirBase;
 	char		dirPath				[CROSS_LEN] = {};
-	DOS_Drive*	drive = NULL;
+	DOS_Drive*	drive = nullptr;
 	char		basePath			[CROSS_LEN] = {};
 	bool		dirFirstTime = false;
 	TDirSort	sortDirType;
@@ -329,7 +329,7 @@ public:
 	virtual Bits UnMount(void)=0;
 
 	/* these 4 may only be used by DOS_Drive_Cache because they have special calling conventions */
-	virtual void *opendir(const char *dir) { (void)dir; return NULL; };
+	virtual void *opendir(const char *dir) { (void)dir; return nullptr; };
 	virtual void closedir(void *handle) { (void)handle; };
     virtual bool read_directory_first(void *handle, char* entry_name, char* entry_sname, bool& is_directory) { (void)handle; (void)entry_name; (void)entry_sname; (void)is_directory; return false; };
     virtual bool read_directory_next(void *handle, char* entry_name, char* entry_sname, bool& is_directory) { (void)handle; (void)entry_name; (void)entry_sname; (void)is_directory; return false; };
--- a/src/dos/cdrom_image.cpp
+++ b/src/dos/cdrom_image.cpp
@@ -117,7 +117,7 @@ CDROM_Interface_Image::BinaryFile::BinaryFile(const char *filename, bool &error)
         typedef wchar_t host_cnv_char_t;
         host_cnv_char_t *CodePageGuestToHost(const char *s);
         const host_cnv_char_t* host_name = CodePageGuestToHost(filename);
-        if (host_name != NULL) {
+        if (host_name != nullptr) {
             file = new ifstream(host_name, ios::in | ios::binary);
             error = file->fail();
         }
@@ -304,7 +304,7 @@ int64_t CDROM_Interface_Image::AudioFile::getLength()
 CDROM_Interface_Image::CHDFile::CHDFile(const char* filename, bool& error)
     :TrackFile(RAW_SECTOR_SIZE) // CDAudioCallBack needs 2352
 {
-    error = chd_open(filename, CHD_OPEN_READ, NULL, &this->chd) != CHDERR_NONE;
+    error = chd_open(filename, CHD_OPEN_READ, nullptr, &this->chd) != CHDERR_NONE;
     if (!error) {
         this->header           = chd_get_header(this->chd);
         this->hunk_buffer      = new uint8_t[this->header->hunkbytes];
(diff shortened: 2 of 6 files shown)


joncampbell123:
> On a related note, there are also warnings in SonarLint suggesting to replace C-style arrays with "std::string", "std::array" or "std::vector", to avoid using "new" and "delete", etc.
> 
> If it's desired to keep DOSBox-X using C-style code (although C++ features are already in use in DOSBox-X code), perhaps that's a reason to not use `nullptr` or features such as the above. I would still prefer to use them, though, to satisfy the warnings and possibly correct bugs (this PR is only for `nullptr`).

Was this program written by a C++ purist with a fear of pointers?

What's wrong with new and delete, or even malloc/free for that matter? Try to code without those and you will end up with a program that is slow, extremely inflexible, and may possibly require more RAM than it should. If you replace everything with the STL you can actually make things more complicated to develop and maintain and open yourself up to mysterious bugs because all you've really done is hide the memory management from yourself and now you have to worry about things like pointer and iterator invalidation. As a good example consider why the menu API in DOSBox-X is so heavily designed around handle values (that are really just indexes into the vector) rather than just pointing at menus. Because as useful as std::vector is, when you add to it, it might reallocate the array (copying it to reallocate, no less!) and invalidate your pointer or iterator.

I don't really care whether NULL or nullptr is involved in the code. I do think though that if anything relies on assigning "0" to a pointer, then yes, that needs to be assigned NULL or nullptr to make it clear you are assigning a value to a pointer, not an integer.

Allofich:
> I don't really care whether NULL or nullptr is involved in the code. I do think though that if anything relies on assigning "0" to a pointer, then yes, that needs to be assigned NULL or nullptr to make it clear you are assigning a value to a pointer, not an integer.

For this, the conversion to nullptr seems like a good thing. For example you can see that the first change in the diff view is for the assignment to `name`, which is a `char*`, which was being given 0. This was flagged for conversion to nullptr, the same as the `NULL` usages were.

About new and delete and all that, certainly I will defer to whatever you think is best for DOSBox-X. The static analyzer program (SonicLint) seems to be based around the CPP Core Guidelines. As far as I know, every warning it gives is tied to one of those guidelines.

joncampbell123:
> > I don't really care whether NULL or nullptr is involved in the code. I do think though that if anything relies on assigning "0" to a pointer, then yes, that needs to be assigned NULL or nullptr to make it clear you are assigning a value to a pointer, not an integer.
> 
> For this, the conversion to nullptr seems like a good thing. For example you can see that the first change in the diff view is for the assignment to `name`, which is a `char*`, which was being given 0. This was flagged for conversion to nullptr, the same as the `NULL` usages were, so addressing `nullptr` warnings should take care of these cases.
> 
> If you would prefer I just replace those `0` cases with `NULL` rather than using `nullptr`, I can do that. `nullptr` is just easier because I can just open a file and check for whether the warning for it appears or not.
> 
> About new and delete and all that, certainly I will defer to whatever you think is best. The static analyzer program (SonicLint) seems to be based around the CPP Core Guidelines. As far as I know, every warning it gives is tied to one of those guidelines.

The CPP Core Guidelines have some good ideas but there's a lot of unnecessary nitpicking in there too and this idea that wrapping things in safe abstractions will solve programming problems. Just know what you're doing, how it works, limitations and capabilities of your code, and you won't have any problems. Even with malloc/free. I don't have a problem with careful use of the STL and maybe even a template or two but you need to watch out with that because those features of C++ can get you into trouble if you overdo them even a little.

I'm going to link to those guidelines. [https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines) If there is a reasonable improvement that can be made with them, run it by me and I'll consider it. I might apply a few reasonable ones as well.

maron2000:
There are NULL values that certain OS won't accept such as the one in the following link.
It may be worth considering if fixing those will not hurt the OS already working.
https://github.com/joncampbell123/dosbox-x/issues/3819#issuecomment-1315033434

Allofich:
Seems like in those cases NULL is being used where 0 would be appropriate, as the type is an integer or enum. A couple are conditioned to use 0 with FreeBSD or OpenBSD and NULL otherwise, I don't know why.

maron2000:
Sorry, my information above was somewhat old.
Some of the codes has already been fixed when intensive patches were applied to make BSD buildable and not breaking other OS builds.
3f396b3

Allofich:
I added fixes for the remaining NULL usages you referenced and will put up a PR shortly. I'm just waiting for the CI builds to run to be sure that they all compile.

Allofich:
> I'm going to link to those guidelines. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines If there is a reasonable improvement that can be made with them, run it by me and I'll consider it. I might apply a few reasonable ones as well.

@joncampbell123, just to clarify, are you waiting on me to take action regarding the above? I haven't read the guidelines, rather I have been just using the static analyzer, which gives its own explanations of the guidelines and links back to them for the issues it flags.

So far I have only made PRs with what seemed to me to be uncontroversial changes, like removing redundant parentheses, unreachable statements, duplicate switch cases, initializing variables and preventing out-of-range memory accesses, removing redundant casts (casting an unsigned int to an unsigned int, etc.), removing unused parameters, using emplace_back instead of push_back (supposed to prevent unneeded object copying) and preferring in-class initialization over constructor list initialization, and list initialization over initialization in the body to avoid excess operations (initialization in the body is actually initialization + an assignment, according to the warning explanation).

Another potentially controversial warning I was looking at was marking functions as const. This is advised by the analyzer/CPP guidelines, but the existing codebase doesn't use const with functions, only with variables, so I have not sent any PRs that make functions const. The analyzer flags member functions that do not modify their object as recommended to be const, so if you approve of that I (or you) can start converting those.

joncampbell123:
> > I'm going to link to those guidelines. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines If there is a reasonable improvement that can be made with them, run it by me and I'll consider it. I might apply a few reasonable ones as well.
> 
> @joncampbell123, just to clarify, are you waiting on me to take action regarding the above? I haven't read the guidelines, rather I have been just using the static analyzer, which gives its own explanations of the guidelines and links back to them for the issues it flags.
> 
> So far I have only made PRs with what seemed to me to be uncontroversial changes, like removing redundant parentheses, unreachable statements, duplicate switch cases, initializing variables and preventing out-of-range memory accesses, removing redundant casts (casting an unsigned int to an unsigned int, etc.), removing unused parameters, using emplace_back instead of push_back (supposed to prevent unneeded object copying) and preferring in-class initialization over constructor list initialization, and list initialization over initialization in the body to avoid excess operations (initialization in the body is actually initialization + an assignment, according to the warning explanation).
> 
> Another potentially controversial warning I was looking at was marking functions as const. This is advised by the analyzer/CPP guidelines, but the existing codebase doesn't use const with functions, only with variables, so I have not sent any PRs that make functions const. The analyzer flags member functions that do not modify their object as recommended to be const, so if you approve of that I (or you) can start converting those.

Not at all. This pull request has become a discussion, apparently :)

If a function and/or the parameters can be marked const without breaking anything, that's fine. I do that in my professional work too. It also helps to mark C++ member functions that do not change state as const because then that allows use of that function even if the C++ instance is const.

PR #4182, opened 2023-04-15
Cleanup Bash scripts.

rderooy:
This makes them pass shellcheck. Except for a few unused export warnings.

I also wonder if the majority of these files should be in the root of the project, or should be moved to "build-scripts" (or should still even exist).


Torinde:
macOS artifact builds not available after this PR?

rderooy:
I think they were already failing before this PR.

rderooy:
Sorry was looking at the wrong place.

Found the issue, will submit a patch this evening.

PR #4207, opened 2023-04-25
Automatically mount directories 'drive[c-y]' found in DOSBox-X executable directory

aybe:

Nothing.

Looks for existing directories named `drivec` and up to `drivey` in DOSBox-X's executable directory.

If these do exist they will be mounted as local drives.

Basically, this allows a user to simply create such directories and they will be immediately mounted without further action.

Probably none.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -114,6 +114,8 @@ Next:
     DOSBox-X is never going to compile for a C standard that old.
     Furthermore this check is preventing compilation on ARM-based
     Macbooks for some reason, though not Intel based Macbooks.
+  - Automatically mount directories found in DOSBox-X's executable
+    directory with naming scheme 'drive[drive letter]'. (aybe)
 
 2023.03.31
   - "mount -t overlay" will now print a message on the console to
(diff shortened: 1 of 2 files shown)


rderooy:
I think dosbox-staging uses a "drive_c" naming convention.

Torinde:
I don't know about "drive_c", but in the Staging folder I see the following sub-folders: `resources\drives\y\`
[Staging Getting Started guide](https://www.johnnovak.net/dosbox-getting-started/getting-started/beneath-a-steel-sky/) shows `drives\c\` and also interestingly `\cd\` where .ISO CD-ROM images are supposed to go?

In any case, yes, common convention between -X and Staging would be good, but it seems they're already different? Maybe then support both conventions...

Automatically mounting \drived\ or \cd\ contents (just files there OR the "first" .ISO found inside) as CD-ROM device at letter D or whichever is the first empty letter?

aybe:
Been thinking about this and I think best is to just abandon, for simplicity and turns out there's already a solution out there, Z:\.

The software already mounts Z:\ and chdir to it, can't be simpler than that.

Issues with doing so are:

- if the user wants to remove that, it will have to go to the directory and move the files, really boring
- the case of A:\, B:\ and CD-ROMs are unsalvageable without implementing some weird convention 

Overall, the simpler the better in my opinion.

Issue #2157, opened 2021-01-05
MOUNT -freesize option is ignored


dajhorn:
**Describe the bug**

`MOUNT -freesize` is ignored for most `freesizecap` values.

**To Reproduce**

Invoke dosbox-x from the command prompt:

1. `dosbox-x.exe -set freesizecap=cap`
1. `dosbox-x.exe -set freesizecap=false`
1. `dosbox-x.exe -set freesizecap=fixed`
1. `dosbox-x.exe -set freesizecap=relative`
1. `dosbox-x.exe -set freesizecap=true`
 
And for each invocation, mount a folder from a host drive that has a large amount of free space:

```
Z:\> MOUNT C D:\Test -freesize 128
Z:\> DIR C:
```

**Expected Behavior**

DOS reports 128 megabytes free in the C drive.

**Actual Behavior**

DOS reports 4 terabytes free in the C drive, which confuses installation programs:

![freesizecap=cap](https://user-images.githubusercontent.com/494022/103592268-9aa31b00-4ec0-11eb-98cf-f580a40dceb5.png)

![freesizecap=relative](https://user-images.githubusercontent.com/494022/103592553-519f9680-4ec1-11eb-9ecf-bb44b61a40ac.png)

![freesizecap=true](https://user-images.githubusercontent.com/494022/103592565-582e0e00-4ec1-11eb-9636-cf1635c5110f.png)

![freesizecap=false](https://user-images.githubusercontent.com/494022/103592567-59f7d180-4ec1-11eb-86e1-039780eed697.png)

True and False behavior is backwards.

Only `freesizecap=fixed` works as expected:

![freesizecap=fixed](https://user-images.githubusercontent.com/494022/103592367-e5249780-4ec0-11eb-952e-99b831911223.png)

This behavior persists if `freesizecap` is instead set in the `dosbox-x.conf` file.

**Environment**
- Windows 10 Version 20H2 Build 19042.685
- Default configuration.

![DOSBox-X 0 83 9 About](https://user-images.githubusercontent.com/494022/103592828-1487d400-4ec2-11eb-96bd-2bb7fe3d78ed.png)

Wengier:
@dajhorn Thanks for checking this feature.

Despite its appearance, there are actually only 3 possible options for the freesizecap config option:

1. freesizecap=cap, which is functionally equivalent to freesizecap=true for backward compatibility purpose.
[...]

PR #4261, opened 2023-05-11
Maxpat78 patch 2 [wip]

### LINKED ISSUE #4258: Free size limit 2TB not to overflow, 2GB limit to be increased
> 2. when free size is above 2TB, then is reported the difference (e.g. some counter reaches a limit and overflows?)
> 3. when cap is used and cap is above 1919 - only 1919 MB are reported

That is more complicated, seems a design choice. 

GetDiskFreeSpaceEx Win32 API would avoid doing calculations and return directly byte sizes... but in _64-bit_ form: `localDrive::AllocationInfo` wants 16-bit answers, in pure ol' DOS fashion... perhaps something like `fatDrive::AllocationInfo32` should be implemented also in `localdrive::`, but probably your multi-TB drives would need 64-bit integers even in such case.

_Originally posted by @maxpat78 in https://github.com/joncampbell123/dosbox-x/issues/4004#issuecomment-1541717976_
            
For item2:
If there is a limitation due to being "too big for DOS", then it should be one of those?
image

Seeing that DOSbox-X shell shows properly 8GB files (beyond the 4GB max for FAT32), I assume it's possible to show volume sizes beyond 2TB as well. Please consider that it's feasible to have a volume with dozens of TB free space (a few high capacity HDDs in NAS).

In any case if actual free space is above the limit, DOSbox should report the limit instead of overflowing (e.g. 2TB+1byte available should not be reported as 1 byte free).

For item3:
when using a freesize cap I get up to a maximum of 1919 MB. Values above give wrong result of 1919MB:
image

maxpat78:
This patch provides a new GetDiskSize64 member in DOS_Drive, and implements it in localDrive class so that internal shell's DIR command can know proper disk size of a (network) drive >2TB.

Fixes #4258 

DOS_Drive::GetDiskSize64 
localDrive::GetDiskSize64 

## Additional information

### DIFF
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -318,6 +318,7 @@ public:
 	virtual bool AllocationInfo32(uint32_t * _bytes_sector,uint32_t * _sectors_cluster,uint32_t * _total_clusters,uint32_t * _free_clusters) { (void)_bytes_sector; (void)_sectors_cluster; (void)_total_clusters; (void)_free_clusters; return false; }
 	virtual bool FileExists(const char* name)=0;
 	virtual bool FileStat(const char* name, FileStat_Block * const stat_block)=0;
+   	virtual bool GetDiskSize64(uint64_t * _disk_size64) { (void) _disk_size64; return false; }
 	virtual uint8_t GetMediaByte(void)=0;
 	virtual void SetDir(const char* path) { strcpy(curdir,path); };
 //	virtual void EmptyCache(void) { dirCache.EmptyCache(); };
--- a/src/dos/drive_fat.cpp
+++ b/src/dos/drive_fat.cpp
@@ -1016,25 +1016,52 @@ bool fatDrive::getDirClustNum(const char *dir, uint32_t *clustNum, bool parDir) 
 	return true;
 }
 
+/*
+   The following sector# to CHS-sector# conversion is unclear to me, since
+   the sector nature would be RELATIVE (but here... to what???), while we
+   need an ABSOLUTE (disk-relative) sector number to pass to loadedDisk
+   read/write functions, which calculate:
+	  sectnum = ( (cylinder * heads + head) * sectors ) + sector - 1L;
+   where heads and sectors are from *DISK* geometry, not *BPB*.
+   
+   However, if CHS is expressed in terms of the loadedDisk geometry
+   (instead of fatDrive's, which can differ), VHD access works fine, and
+   RAW images keep working.  2023.05.11 - maxpat78 */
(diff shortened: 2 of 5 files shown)


Torinde:
I can't test now, but will do later.

I assume it solves the 2TB overflow, but doesn't solve the "when freesize cap is used and cap is above 1919 - only 1919 MB are reported", correct?

maxpat78:
> I can't test now, but will do later.
> 
> I assume it solves the 2TB overflow, but doesn't solve the "when freesize cap is used and cap is above 1919 - only 1919 MB are reported", correct?

Try it. I didn't test cap nor side effects.

joncampbell123:
> > I can't test now, but will do later.
> > I assume it solves the 2TB overflow, but doesn't solve the "when freesize cap is used and cap is above 1919 - only 1919 MB are reported", correct?
> 
> Try it. I didn't test cap nor side effects.

Both of you: Let me know if it works, then I'll merge it.

Torinde:
1. I get "2 491,17 G Bytes free" - so that works!
2. But the freesize cap is ignored - wrong behavior seems similar to the issue I mentioned in my OP : #2157
3. When booting MS-DOS 7.10 with `convert fat free space = 1147483` (expecting ~1TB) it seems that it overflows:
image

joncampbell123:
Should I merge this or hold it until after release at the end of the month?

Torinde:
IMHO - Hold. Hopefully maxpat78 can look at it after he's done with the more important VHD improvements.

My results:
Item1 is good (improvement). Item3 is unchanged.
Item2 is regression (in Master builds the cap works at least till 1919MB, but here it's ignored?)

maxpat78:
Mmmm... time, yes, to setup a Debian VM and test effects of deeper changes, to accomodate all runtime scenarios!

Torinde:
@maxpat78, any tests I can run to help?

maxpat78:
@Torinde Perhaps I'll find some time to work on this later, I don't think tests can help at this point (I'll have to carry on some tests under Linux, one time or another, to let the fix work outside Windows).
However, if I recall well, your problem originates from some hard-coded value put there to let the full code work (DOSBox-X often set things instead of measure real capacities, i.e. in formatting related interrupts).

Lartenik:
Я сделал перевод en_US  на русский язык ru_RU ,  есть возможно сть адаптировать этот файл для корректной работы программы ?
[ru_RU.txt](https://github.com/joncampbell123/dosbox-x/files/14048994/ru_RU.txt)

Issue #4249, opened 2023-05-07
Break key is Scroll lock in Opera games (Goody, Living, Mission)


Torinde:
### Question

Opera software games are supposed to use the `Break` key to return to the title screen (when you are playing). But both in DOSbox-X and Staging - `Pause/Break` doesn't do anything (even with AltGr+Pause, Alt+Pause, [Ctrl+NumLock](https://en.wikipedia.org/wiki/Break_key#Modern_keyboards))
![image](https://user-images.githubusercontent.com/97228894/236670089-fb36399b-f83d-4f9c-8085-c94cfe011320.png)
![image](https://user-images.githubusercontent.com/97228894/236670176-29cbb30d-b083-4f4f-95eb-2cf25e790098.png)

With `Scroll Lock` it works!

Are these two key mixed up on the DOSbox side - or the games' title screens are wrong? (as e.g. Breack is misspelled in Goody)

Could it be that Spanish keyboard has those scancodes different from other keyboards? 

@aybe @FeralChild64

### Have you checked that no similar question(s) exist?

- [X] I have searched and didn't find any similar question.

### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

maron2000:
In general, there are no dedicated key for Break key but, CTRL+Scroll Lock and CTRL+Pause.
So pressing Scroll Lock as Break makes sense to some extent.
The behavior of actual machine shouldd be checked before suspecting this is a bug.

Torinde:
When I say "With `Scroll Lock` it works!" - I mean with Scroll lock ONLY. Not a CTRL+/ALT+ combination. That's why I think either keys are swapped or game instructions are wrong.

I'll try running those on a physical system...

maron2000:
Yes I perfectly understood your explanation.
I mentioned `pressing Scroll Lock`, not CTRL+Scroll Lock.
You can additionally try `CTRL+Pause` which is the first thing you should try if you have Pause/Break key.

Torinde:
Ran Goody on modern laptop booting FreeDOS - Scroll lock works, Pause/Break doesn't work.

So, conclusion is Opera software problem, not DOSbox problem?

Issue #3583, opened 2022-06-19
lDDebug PRA (int 22h) hook (self parented termination) does not work properly


ecm-pushbx:
### Describe the bug

As I noted in https://github.com/joncampbell123/dosbox-x/issues/3582

> Note that the debugger installs itself as a self-parented shell process, that is it modifies the parent field and the parent return address (interrupt 22h) field of its own PSP. This is intended so that a Control-C caught by DOS will abort the process, but that this will in turn not free the file handles or memory blocks of the process and instead return [to the debug22 entrypoint](https://hg.pushbx.org/ecm/ldebug/file/b1d1b70a60b8/source/debug.asm#l1286) of the debugger.

Attempting to run interrupt 21h service 4Ch (return code 00h) in lDDebug's process context does not work, it crashes a bit and albeit returning to one of the applications, it leaves the system in an unusable state. I test this by running lDDebug within lDebug, then [breaking upwards](https://pushbx.org/ecm/doc/ldebug.htm#cmdbu) in lDDebug, then from lDebug [use the QA command](https://pushbx.org/ecm/doc/ldebug.htm#cmdqa), which runs interrupt 21h service 4Ch in the process context of lDDebug.

### Steps to reproduce the behaviour

1. Download https://pushbx.org/ecm/download/ldebug.zip and https://pushbx.org/ecm/download/lddebug.zip (these will move to https://pushbx.org/ecm/download/old/ldebug/20220604.zip and https://pushbx.org/ecm/download/old/lddebug/20220604.zip once I have updated the application)
2. Extract ldebug.zip bin/ldebugu.com and lddebug.zip bin/lddebugu.com to ~/dosbox directory
3. Run DOSBox-X
4. Run `keyb gr`
5. Run `mount c ~/dosbox`
6. Run `c:`
7. Enter the command `ldebugu.com /c='g;qa' lddebugu.com /c='bu;bu'`

### Expected behavior

Output should look like this: (obtained from running the command in a dosemu2 process)

```
&g
~&bu
Breaking to next instance.
Unexpected breakpoint interrupt
AX=000D BX=0002 CX=0002 DX=73AD SP=C0EC BP=000A SI=B515 DI=BCBE
DS=20CA ES=20CA SS=20CA CS=2CD9 IP=8131 NV UP EI PL ZR NA PE NC
2CD9:8131 C3                retn
&qa
~&bu
Breaking to next instance.
Unexpected breakpoint interrupt
[...]

Issue #3582, opened 2022-06-19
Control-C acts strange in my application


ecm-pushbx:
### Describe the bug

Control-C behaves oddly when I enter it to the command-line interface of my application (DOS debugger). This is true, though differing in the details, both when using interrupt 21h service 0Ah to read input as well as when the debugger is instructed [to use its `rawinput` function](https://hg.pushbx.org/ecm/ldebug/file/b1d1b70a60b8/source/lineio.asm#l2331) (by setting DCO flag 800h), which in turn [uses `getc` to read keypress by keypress](https://hg.pushbx.org/ecm/ldebug/file/b1d1b70a60b8/source/lineio.asm#l1198) which [calls interrupt 21h service 08h to get a keypress](https://hg.pushbx.org/ecm/ldebug/file/b1d1b70a60b8/source/lineio.asm#l1370).

Note that the debugger installs itself as a self-parented shell process, that is it modifies the parent field and the parent return address (interrupt 22h) field of its own PSP. This is intended so that a Control-C caught by DOS will abort the process, but that this will in turn not free the file handles or memory blocks of the process and instead return [to the `debug22` entrypoint](https://hg.pushbx.org/ecm/ldebug/file/b1d1b70a60b8/source/debug.asm#l1286) of the debugger.

I also noticed that when I download `lddebug.zip` and extract `bin/lddebugu.com` and run `ldebugu.com lddebugu.com` then it appears that the interrupt 23h handler of the debuggable debugger points at 072Fh:0000h, which holds a single `int 20h` instruction, but this address is never actually called. (Setting a permanent breakpoint on it in the outer debugger, then entering Control-C to the inner debugger, never breaks at that breakpoint.)

When setting the DCO flag 8 (use `rawinput` with ROM-BIOS input through interrupt 16h) then no unexpected results occur.

### Steps to reproduce the behaviour

1. Download https://pushbx.org/ecm/download/ldebug.zip (this will move to https://pushbx.org/ecm/download/old/ldebug/20220604.zip once I have updated my applications)
2. Extract bin/ldebugu.com to ~/dosbox directory
3. Run DOSBox-X
4. Run `keyb gr`
5. Run `mount c ~/dosbox`
6. Run `c:`
7. Run `ldebugu.com`
8. Input Control-C
9. Input `a` then Control-C
10. Input Enter
11. Input `r dco or= 800` then Enter
12. Input Control-C
13. Input Enter
14. Input Enter
[...]

PR #4270, opened 2023-05-15
Fix break keys (Windows SDL1/SDL2)

maron2000:
This PR fixes Ctrl+Pause and Ctrl+Scroll Lock to work as Break key.
Tested on Windows VS builds SDL1/SDL2.

### DIFF
--- a/include/keyboard.h
+++ b/include/keyboard.h
@@ -20,23 +20,23 @@
 #define DOSBOX_KEYBOARD_H
 
 enum KBD_KEYS {
-	KBD_NONE,
-	KBD_1,	KBD_2,	KBD_3,	KBD_4,	KBD_5,	KBD_6,	KBD_7,	KBD_8,	KBD_9,	KBD_0,		
-	KBD_q,	KBD_w,	KBD_e,	KBD_r,	KBD_t,	KBD_y,	KBD_u,	KBD_i,	KBD_o,	KBD_p,	
-	KBD_a,	KBD_s,	KBD_d,	KBD_f,	KBD_g,	KBD_h,	KBD_j,	KBD_k,	KBD_l,	KBD_z,
-	KBD_x,	KBD_c,	KBD_v,	KBD_b,	KBD_n,	KBD_m,	
-	KBD_f1,	KBD_f2,	KBD_f3,	KBD_f4,	KBD_f5,	KBD_f6,	KBD_f7,	KBD_f8,	KBD_f9,	KBD_f10,KBD_f11,KBD_f12,
-	
-	/*Now the weirder keys */
+    KBD_NONE,
+    KBD_1, KBD_2, KBD_3, KBD_4, KBD_5, KBD_6, KBD_7, KBD_8, KBD_9, KBD_0,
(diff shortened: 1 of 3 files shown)


Torinde:
Per Wikipedia I see the following combinations:
- [SystemRequest](https://en.wikipedia.org/wiki/System_request) = Alt + PrintScreen
- [Break](https://en.wikipedia.org/wiki/Break_key#Modern_keyboards) = Ctrl + ScrollLock
- Pause = Ctrl + NumLock
Pressing the dedicated `Pause` key on 101/102-key keyboards sends the same [scancodes](https://en.wikipedia.org/wiki/Scancode) as pressing `Ctrl`, then `NumLock`, then releasing them in the reverse order would do; additionally, an E1hex prefix is sent
The `Pause` key is different from all other keys in that it sends no scancodes at all on release in PS/2 modes 1 or 2, so it is impossible to determine whether this key is being held down with older devices. In PS/2 mode 3 or USB HID mode, there is a release scancode

I don't see there Ctrl + Pause combination.

joncampbell123:
> Per Wikipedia I see the following combinations:
> 
>     * [SystemRequest](https://en.wikipedia.org/wiki/System_request) = Alt + PrintScreen
> 
>     * [Break](https://en.wikipedia.org/wiki/Break_key#Modern_keyboards) = Ctrl + ScrollLock
> 
>     * Pause = Ctrl + NumLock
>       Pressing the dedicated `Pause` key on 101/102-key keyboards sends the same [scancodes](https://en.wikipedia.org/wiki/Scancode) as pressing `Ctrl`, then `NumLock`, then releasing them in the reverse order would do; additionally, an E1hex prefix is sent
>       The `Pause` key is different from all other keys in that it sends no scancodes at all on release in PS/2 modes 1 or 2, so it is impossible to determine whether this key is being held down with older devices. In PS/2 mode 3 or USB HID mode, there is a release scancode
> 
> 
> I don't see there Ctrl + Pause combination.

Yes, but the OS is not the one sending CTRL + scroll lock, the emulator is responsible for that. Therefore there is no reason to make a special case for CTRL + scroll lock at that code.

Pause becomes CTRL+Break if you hold down CTRL. 
image

joncampbell123:
I appreciate your effort but I don't think this is the right way to go about it.

If Windows builds are having difficulty allowing CTRL+Break and Pause to work, that should be addressed, certainly.

maron2000:
Ok, I'm not strongly sticking to my fix, but sure it has to be fixed.

Torinde:
@maron2000, can you please share where the problem is, how it can be tested?
I myself thought there is problem with Break key, but it transpired the games claiming to use it actually use Scroll Lock :) #4249

Then I see two other issues, but I'm not sure they're related to what you mean: #3582, #3583

Torinde:
When using MORE command - pressing Ctrl+NumLock (equivalent of Pause?) stops it and prints ^C (behavior of Break?), so is Wikipedia is wrong or DOSbox or myself?

image

Ctrl+ScrLock (equivalent of Break), ScrLock, Ctrl, Alt, Shift, Menu, PrtScr, NumLock - don't do anything. Pause - does Pause (e.g. you need to press twice some letter to proceed instead of once).

Not sure if related to the keyboard emulation or to implementation of MORE #4292

PR #4273, opened 2023-05-16
Enhanced Dynamic and Differencing VHD support

maxpat78:
DOSBox-X is enhanced with the ability to create Dynamic and Differencing VHD images (and to convert raw hd images into Fixed VHD) to play with IMGMOUNT and share data with Windows easily.

- extends imageDiskVHD in bios_vhd.cpp to create Fixed, Dynamic and Differencing VHD on the fly;
- adds a VHDMAKE.COM virtual program to access new functionalities;
- patches IMGMAKE to create both Dynamic and Fixed VHD, already partitioned and formatted.

New IMGMAKE syntax:
  `IMGMAKE new_dynamic.vhd -t vhd -size XX` creates a Dynamic VHD.
  `IMGMAKE new_fixed.vhd -t hd -size XX` creates a Fixed VHD.

### DIFF
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -299,9 +299,9 @@ public:
 	static VHDTypes GetVHDType(const char* fileName);
 	VHDTypes GetVHDType(void) const;
 	virtual ~imageDiskVHD();
-	uint32_t mk_crc(uint8_t* s, uint32_t size);
-	void mk_chs(uint64_t size, char* chs);
-	uint32_t CreateDynamic(char* filename, uint64_t size);
+	static uint32_t mk_crc(uint8_t* s, uint32_t size);
+	static void mk_chs(uint64_t size, char* chs);
+	static uint32_t CreateDynamic(char* filename, uint64_t size);
 
 private:
 	struct Geometry {
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -278,6 +278,8 @@ public:
 		INVALID_MATCH = 4,
 		INVALID_DATE = 5,
 		UNSUPPORTED_WRITE = 6,
+        UNSUPPORTED_SIZE = 7,
+        ERROR_WRITING = 8,
 		PARENT_ERROR = 0x10,
 		ERROR_OPENING_PARENT = 0x11,
 		PARENT_INVALID_DATA = 0x12,
@@ -298,9 +300,12 @@ public:
 	static ErrorCodes Open(const char* fileName, const bool readOnly, imageDisk** disk);
 	static VHDTypes GetVHDType(const char* fileName);
 	VHDTypes GetVHDType(void) const;
-	virtual ~imageDiskVHD();
-	static uint32_t CreateDynamic(const char* filename, uint64_t size);
+    static uint32_t CreateFixed(const char* filename, uint64_t size);
+    static uint32_t ConvertFixed(const char* filename);
+    static uint32_t CreateDynamic(const char* filename, uint64_t size);
 	static uint32_t CreateDifferencing(const char* filename, const char* basename);
+    static void PseudoCHSFromSize(uint64_t size, uint32_t* c, uint32_t* h, uint32_t* s);
+    virtual ~imageDiskVHD();
 
 private:
 	struct Geometry {
@@ -363,7 +368,6 @@ private:
 	virtual bool loadBlock(const uint32_t blockNumber);
 	static bool convert_UTF16_for_fopen(std::string &string, const void* data, const uint32_t dataLength);
     static uint32_t mk_crc(uint8_t* s, uint32_t size);
-    static void mk_chs(uint64_t size, char* chs);
 
     imageDisk* parentDisk = NULL;
 	uint64_t footerPosition = 0;
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -299,6 +299,9 @@ public:
 	static VHDTypes GetVHDType(const char* fileName);
 	VHDTypes GetVHDType(void) const;
 	virtual ~imageDiskVHD();
+	uint32_t mk_crc(uint8_t* s, uint32_t size);
+	void mk_chs(uint64_t size, char* chs);
+	uint32_t CreateDynamic(char* filename, uint64_t size);
 
 private:
 	struct Geometry {
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -299,9 +299,8 @@ public:
 	static VHDTypes GetVHDType(const char* fileName);
 	VHDTypes GetVHDType(void) const;
 	virtual ~imageDiskVHD();
-	static uint32_t mk_crc(uint8_t* s, uint32_t size);
-	static void mk_chs(uint64_t size, char* chs);
-	static uint32_t CreateDynamic(char* filename, uint64_t size);
+	static uint32_t CreateDynamic(const char* filename, uint64_t size);
+	static uint32_t CreateDifferencing(const char* filename, const char* basename);
 
 private:
 	struct Geometry {
@@ -363,6 +362,8 @@ private:
 	static ErrorCodes Open(const char* fileName, const bool readOnly, imageDisk** disk, const uint8_t* matchUniqueId);
 	virtual bool loadBlock(const uint32_t blockNumber);
 	static bool convert_UTF16_for_fopen(std::string &string, const void* data, const uint32_t dataLength);
+    static uint32_t mk_crc(uint8_t* s, uint32_t size);
+    static void mk_chs(uint64_t size, char* chs);
 
     imageDisk* parentDisk = NULL;
 	uint64_t footerPosition = 0;
--- a/src/dos/dos_programs.cpp
+++ b/src/dos/dos_programs.cpp
@@ -7729,6 +7729,82 @@ static void TITLE_ProgramStart(Program * * make) {
     *make=new TITLE;
 }
 
+class VHDMAKE : public Program {
+public:
+    void Run(void);
+private:
+    uint64_t ssizetou64(const char* s_size);
+	void PrintUsage() {
+        constexpr const char *msg =
+           "Creates a blank Dynamic VHD image.\n\n"
+           "VHDMAKE filename nnn[BKMGT]\n\n"
+           "  filename   Specifies the name of the new VHD image file.\n"
+           "  nnn        Specifies the virtual disk size.\n";
+        WriteOut(msg);
+	}
+};
+
+// Converts a string disk size with unit into a 64-bit unsigned integer
+uint64_t VHDMAKE::ssizetou64(const char* s_size) {
+    char* sizes = "BKMGT";
+    char* sd_size = strdup(s_size);
+    char* last = sd_size + strlen(s_size) - 1;
+    char* c;
+    uint64_t size;
+
+    if((c = strchr(sizes, toupper(*last)))) {
+        *last = 0;
+        size = atoll(sd_size);
+        size <<= ((c - sizes) * 10);
+    }
+    else {
+        size = atoll(sd_size);
+    }
+    free(sd_size);
+    return size;
+}
+
+void VHDMAKE::Run()
+{
+	// Hack To allow long commandlines
+	ChangeToLongCmd();
+
+	// Usage
+	if (cmd->FindExist("-?", false) || cmd->FindExist("/?", false) || cmd->GetCount() < 2) {
+		PrintUsage();
+		return;
+	}
+
+    char filename[256], size[16];
+    cmd->FindCommand(1, temp_line);
+    safe_strcpy(filename, temp_line.c_str());
+
+    cmd->FindCommand(2, temp_line);
+    safe_strcpy(size, temp_line.c_str());
+
+    uint32_t ret = imageDiskVHD::CreateDynamic(filename, ssizetou64(size));
+
+    switch(ret) {
+    case 2:
+        WriteOut("Can't create a VHD < 5 MB !\n");
+        break;
+    case 3:
+        WriteOut("Can't create a VHD > 2040 GB !\n");
+        break;
+    case 4:
+        WriteOut("Error, could not open %s for writing", filename);
+        break;
+    default:
+        WriteOut("New blank VHD image succesfully created. You can mount it with \033[34;1mIMGMOUNT\033[0m.\n");
+        break;
+    }
+}
+
+
+static void VHDMAKE_ProgramStart(Program * * make) {
+    *make=new VHDMAKE;
+}
+
 class COLOR : public Program {
 public:
     void Run(void);
@@ -8540,6 +8616,7 @@ void Add_VFiles(bool usecp) {
 
     PROGRAMS_MakeFile("COLOR.COM",COLOR_ProgramStart,"/BIN/");
     PROGRAMS_MakeFile("TITLE.COM",TITLE_ProgramStart,"/BIN/");
+    PROGRAMS_MakeFile("VHDMAKE.COM",VHDMAKE_ProgramStart,"/BIN/");
     PROGRAMS_MakeFile("LS.COM",LS_ProgramStart,"/BIN/");
     PROGRAMS_MakeFile("ADDKEY.COM",ADDKEY_ProgramStart,"/BIN/");
     PROGRAMS_MakeFile("CFGTOOL.COM",CFGTOOL_ProgramStart,"/SYSTEM/");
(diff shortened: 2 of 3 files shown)


maxpat78:
@Torinde VHDX is out of scope, it would require at least twice the coding found in that clever bios_vhd.cpp, and be _many_ times harder (see partial Python implementation in my FATtools project). Moreover, I don't think old DOS emulation requires multi-TB disks provided by VHDX.

Differencing VHD creation was just added, and a game setup: OS (Father VHD) -> Game (Son VHD) -> Game states/saves (Nephew VHD) should be already possible, and easily implementable (I've tried a couple ol' Win installs myself to comprove proper functionality).

Actually, new VHD stuff needs to be urgently coordinated with IMGMAKE (DosBox-X does not like very much un-partitioned or un-formatted drives; FreeDOS FORMAT is not so mature to handle formatting of such drives inside the internal shell, you have to boot a "real" DOS session); or I'll try to implement simple MBR-partitioning and FAT-formatting services in VHDMAKE.

Torinde:
Agree that having more formatting capabilities/options would be good, I also had difficulties with "fresh" un-initialized/un-partitioned/un-formatted images... of course you can format them outside DOSbox-X, but a native "known good" solution would help diagnosing some fringe configurations.

VHDX 
- less than 2TB images can be created as VHDX as well (and have advantages over VHD, supposedly)? So it's more about "recent host" compatibility rather than "multi-TB in DOS/Win9x"
- GPT support in Win9x via the 3rd party [GPTTSD](https://msfn.org/board/topic/184646-terabyte-pack-2/): https://github.com/LordOfMice/Tools
- GPT is [officially supported](https://learn.microsoft.com/en-us/previous-versions/windows/hardware/design/dn640535(v=vs.85)?redirectedfrom=MSDN#can-the-32--and-64-bit-versions-of-windows-server-2003-read-write-and-boot-from-gpt-disks) by Win 2003 SP1 (the server edition of XP and it seems XP may get bootable in DOSbox-X)

Differencing - sorry, now I notice you mention it's supported. I wonder how/if you made the UI/commands for branching/merging (haven't tested your build yet)

It seems there are 3 failing builds - HX DOS / MinGW32 lowend / XP?

maxpat78:
> VHDX
> 
> * less than 2TB images can be created as VHDX as well (and have advantages over VHD, supposedly)? So it's more about "recent host" compatibility rather than "multi-TB in DOS/Win9x"

The only "advantage" I know about is metadata journaling (some internal structures are protected by a journal against write errors/system failures - VHD does not check anymore neither timestamps to enforce parent-child relationship.....)

> * GPT support in Win9x via the 3rd party [GPTTSD](https://msfn.org/board/topic/184646-terabyte-pack-2/): https://github.com/LordOfMice/Tools

Ok, but... 

> Differencing - sorry, now I notice you mention it's supported. I wonder how/if you made the UI/commands for branching/merging (haven't tested your build yet)

What do you mean "branching"?
Merging is a thing could come later...

> 
> It seems there are 3 failing builds - HX DOS / MinGW32 lowend / XP?

MinGW32_lowend_CI_build (SDL2) and MinGW32_lowend_CI_build seem unrelated to my changes?
```
138: cpp.exe: error: C: No such file or directory
139: cpp.exe: warning: '-x c' after last input file has no effect
140: cpp.exe: fatal error: no input files
```

HX_DOS_CI_build, instead (I'll try to figure out where time is declared there):
```
bios_vhd.cpp:600:12: error: 'time' was not declared in this scope
```

Torinde:
all builds seem OK (at least are green)

Branching - I meant the "linking" of Father/Son images - I saw now how you do it for Father/Son (parent.vhd new.vhd). But is it recursive, can I add 3rd level of branches, e.g. "-l os.vhd game.vhd savestate1.vhd"?

And what do I put in IMGMOUNT then?

Also, I suppose after making multiple of those branches - user may get confused about the relationships. Is the information contained in the child images themselves? Can I run something like "VHDMAKE -map parent.vhd" and get the tree of all subsequent branches? Or somebody (user, VHDMAKE) should keep track of that somewhere (napkin, log file)?

What I'm getting into is something akin the Snapshot browser in Virtualbox (splitting, merging, tree/map, user descriptions).

Anyway, I think what you did so far is great! As you say - merging can be added later. "Show tree" capability would be nice, but even without it differencing support in VHDMAKE is quite useful!


About the partitioning/formatting - would be good to have the option the created VHD files to be "ready for use", e.g. like from [IMGMAKE](https://dosbox-x.com/wiki/Guide%3AManaging-image-files-in-DOSBox%E2%80%90X#_vhd_support): `The advantage of using IMGMAKE is that the created image file will be partitioned and formatted and as such is ready to be mounted. The disadvantage is that only fixed size VHD images can be created this way.`

```
VHDMAKE [-f] new.vhd nnn[BKMGT]
VHDMAKE [-f] -l parent.vhd new.vhd

  -f         Force overwriting a file
  -l         Links to a parent VHD image
  filename   Specifies the name of the new VHD image file
  nnn        Specifies the disk size (eventually with size unit)
             (B)ytes is implicit
  parent.vhd Specifies the base VHD to link the new differencing VHD to
  new.vhd    Specifies the new differencing VHD to make

```

maxpat78:
@Torinde with `VHDMAKE -l parent.vhd child.vhd` you add one link to the chain at a time, so if you want a "nephew" or 3rd grade link you have to do `VHDMAKE -l child.vhd nephew.vhd` (or, better, use suggestive names like child-1.vhd, child-2.vhd and so on).

Then you'll issue `IMGMOUNT 2 nephew.vhd` to mount the raw tri-image chain of the example.

Please notice that my last commit patches IMGMAKE to directly emit a (partitioned and formatted) Dynamic image, so VHDMAKE will last for chain linking matter only.

Notice also (but this is explicited in VHD spec) that once you've created a chain, you *never* have to mount and write to intermediate images, or the chain will get corrupted from that point onward (there's little check on this: DosBox-X, and Windows 11 itself, actually look only that a child GUID matches the parent's GUID; no check on internal or external timestamps is performed).

Perhaps I'll implement merging operation that, substantially, equals to copying all blocks from the last linked VHD into a fresh VHD image.

Torinde:
Thank you, that made it more clear for me!

Also you had good workaround for the relationship tracking - instead of napkin to use the filenames :)

About never mounting the parent image - yes, I get that. I'm not sure right now, but I think Virtualbox manages that by creating copies of the parents where needed. Anyway, that's something the user can manage externally as well.

Finally, just for my curiosity - how many levels deep can the chain go? Three or more?

maxpat78:
> Finally, just for my curiosity - how many levels deep can the chain go? Three or more?
There is no limit, as far as I can remember - no internal counters or something like that.
In short, the answer is: more than a normal user could require.

Actually, I'm trying to resolve another big issue: internal VHD CHS geometry and FAT BPB geometry often disagree, so IMGMOUNT refuses to mount a VHD image created by IMGMAKE (and properly recognized by Windows 1x, 7zFM or VirtualBox).

maxpat78:
Commit [19ee8af ](https://github.com/joncampbell123/dosbox-x/pull/4273/commits/19ee8aff321e3f84dd9456ebccd8721a140a7af0) mitigates an annoying issue with `IMGMOUNT` and VHD images.

Since DosBox-X reads VHD geometry from VHD headers instead of FAT BPB, at some point mounting many valid VHD images (i.e., correctly handled by Windows, Virtualbox or 7zFM) leads to a "Sanity checks failed" in `fatDrive`, line 1899, where the test `(BPB.v.BPB_SecPerTrk > cylsector && !IS_PC98_ARCH))` very often fails.

My latest patch forces `IMGMAKE` to use internal VHD geometry while creating the partition table and the boot sector, so at least Dynamic VHDs created by DOSBox-X can get mounted.

Torinde:
> Please notice that my last commit patches IMGMAKE to directly emit a (partitioned and formatted) Dynamic image, so VHDMAKE will last for chain linking matter only.

I couldn't do that - IMGMAKE VHD is corrupt and also fixed size (not dynamic)
image
image

I think I downloaded your latest artifact earlier today, but maybe I made a mistake? DOSbox-X.exe size is 20 240 896 bytes (MinGW64 SDL1)
VHDMAKE created image can be mounted with "IMGMOUNT 2", FDISK creates partition (from DOSbox shell), after restart "IMGMOUNT C" gives error: "Cannot create drive from file". Same happened when instead of DOSbox FDISK I used my host Windows Disk management.
Maybe some mistake on my side - I need to test more.

A side note: I prefer dynamic images, but probably it's good to keep the option to create fixed size as well (e.g. at least in one of IMGMAKE and/or VHDMAKE)

WOW... while writing the above it seems you made improvements that may solve it? :)

maxpat78:
@Torinde you used the wrong syntax and created a non-VHD image. Try:
`imgmake test.vhd -t vhd -size 20 -force`

where `-t vhd` is the new image format (Dynamic VHD), `-size 20` (without unit specification: the default IMGMAKE behaviour) the disk size, and `-force` to overwrite your pre-existing (bad) image file.

maxpat78:
> A side note: I prefer dynamic images, but probably it's good to keep the option to create fixed size as well (e.g. at least in one of IMGMAKE and/or VHDMAKE)

It was on my list to add a "conversion" tool in `VHDMAKE` (since a Fixed VHD is simply a raw image with a 512-byte VHD Footer sector appended).

Torinde:
Yes, with this latest commit and `-t vhd` IMGMAKE created a dynamic image that was immediately mountable with IMGMOUNT - no need to partition/format!

Where do you think `-t vhd` should be mentioned? In IMGMAKE /? output and/or in the [Wiki](https://dosbox-x.com/wiki/Guide%3AManaging-image-files-in-DOSBox%E2%80%90X#_vhd_support)?

For the previously existing fixed VHD creation capability - it uses the regular "-t hd" and is triggered by the .vhd extension, correct?
```
Or to create it with IMGMAKE, simply ensure the filename extension is .vhd. e.g.,
IMGMAKE -t hd -size 40 hdd.vhd
```
I didn't have success with that, but I assume that's not related to this PR anyway.

For the parent/child case - what about VHDMAKE making the parent.vhd read-only when creating a child.vhd?

maxpat78:
@Torinde help updates and internationalization will come later.
Actually, fixed VHD creation is disabled in IMGMAKE or elsewhere, I'll see that later.

> For the parent/child case - what about VHDMAKE making the parent.vhd read-only when creating a child.vhd?

This isn't in the official specs, but could do a valid trick. Also (re-)enabling timestamps check could be a good idea.

Torinde:
Great!

My cursory testing discovered only a few cosmetic issues:

**VHDMAKE - the "<3M!" message doesn't appear** (in the source I didn't see what the difference is from the >2040G! message, so why it's not working ??)
image
Unrelated trivia: Ctrl+Tab gives list of files and directories - similar to `LS` command, but not the same (pressed that between the two VHDMAKE commands in the screenshot). Where can I read more about this function?

**IMGMAKE doesn't give <3M! or >2040G! messages**, but just outputs regular /? text

IMGMAKE -t hd works, IMGMOUNT mounts directly, no need to partition/format.

**IMGMAKE -t vhd gives duplicated messages:**
image
Only one file is created. IMGMOUNT mounts directly, no need to partition/format.
Using -t hd gives only one message. 

image
VHDMAKE without `-l` doesn't work (per the help, so that's OK) - just mentioning that when second string is not a valid size, but a filename (especially with .vhd extension) - maybe the differencing linked image can be created even without `-l`.

Differencing images are created and mountable both for fixed and dynamic size parents.

IMGMAKE can make a -t vhd -nofs dynamic image, it then behaves like the image.vhd -t hd fixed one: FDISK partitions, but IMGMOUNT doesn't mount (e.g. as expected, but annoying)

joncampbell123:
Should I merge this? Is it ready?

maxpat78:
> Should I merge this? Is it ready?

I'm refining some things, it is useful and useable already but perhaps you'd better wait a bit more. Your decision.

maxpat78:
@joncampbell123 OK, it's ready to be merged.

Here a batch file showing a sample of the new capabilities:
```
@echo off
cls
vhdmake a.vhd 2m
vhdmake a.vhd 2041g

echo Fixed VHD...
imgmake fixed.vhd -t hd -size 80 -force 
imgmount k fixed.vhd
imgmount -u k

echo ...and its Differencing VHD
vhdmake -l fixed.vhd delta.vhd -f
imgmount k delta.vhd
imgmount -u k

echo Dynamic VHD...
imgmake dyna.vhd -t vhd -size 540 -force
imgmount k dyna.vhd
imgmount -u k

echo ...and its Differencing VHD
vhdmake -l dyna.vhd dyna_delta.vhd -f
imgmount k dyna_delta.vhd
imgmount -u k

echo Raw HD image conversion...
imgmake static.raw -t hd -size 32 -force
vhdmake -c -f static.raw static.vhd
imgmount k static.vhd
imgmount -u k
```

maxpat78:
@joncampbell123 it's ready!

Here a batch file showing new functionalities:
```
@echo off
cls
vhdmake a.vhd 2m
vhdmake a.vhd 2041g

echo Fixed VHD...
imgmake fixed.vhd -t hd -size 80 -force 
imgmount k fixed.vhd
imgmount -u k

echo ...and its Differencing VHD
vhdmake -l fixed.vhd delta.vhd -f
imgmount k delta.vhd
imgmount -u k

echo Dynamic VHD...
imgmake dyna.vhd -t vhd -size 540 -force
imgmount k dyna.vhd
imgmount -u k

echo ...and its Differencing VHD
vhdmake -l dyna.vhd dyna_delta.vhd -f
imgmount k dyna_delta.vhd
imgmount -u k

echo Raw HD image conversion...
imgmake static.raw -t hd -size 32 -force
vhdmake -c -f static.raw static.vhd
imgmount k static.vhd
imgmount -u k
```

Torinde:
I think it can be merged as well.

Minor items for afterwards, whenever/if you have time:
1. IMGMAKE help to be changed, for example:   `-size: Size of a custom hard disk image in MB. For VHD between 3 and 2088967.` (that'll take care/make use of the "unexpected help output" when selecting too small/big size)
2. merging capability for child+parent
3. making the parent.vhd read-only when creating a child.vhd
4. (re-)enabling timestamps check - is there a scenario when user may want timestamps NOT to be checked?
5. not sure if it's a problem, but IMGMAKE from 2088968 to 2097151 behaves differently ("Could not open image file") than 2097152 onwards (shows help):
image

maxpat78:
@Torinde @joncampbell123 commit https://github.com/joncampbell123/dosbox-x/pull/4273/commits/de36972f9d3225d3bf363c2315155d9b72d4c306 fixes IMGMAKE weird behaviour with wrong sizes and helps about vhd size limits (2088960 MB is the upper vhd limit).

Please note that remains a weird corner-case with _fixed_ VHDs, more annoying to fix due to the fact that `-size` parameter is checked before calculating file `extension`.

maxpat78:
@Torinde
> merging capability for child+parent

Later, perhaps

>making the parent.vhd read-only when creating a child.vhd
>(re-)enabling timestamps check - is there a scenario when user may want timestamps NOT to be checked?

Windows (11) does not. Reasons should be evaluated.
Making parent RO annoys a user that wants mount and use the parent discarding the child, for example.
Time stamp checking is more useful (if extended to `fstat` ed container results, not only the _internal_ ts), but adds complexity (internal ts should be UTC always) and perplexity (copying/moving container(s), or simply accessing them, could lead to ts alterations, but not relevant to the mounter's perspective).

Torinde:
Agree the Read-only/timestamp check may have disadvantages - is "make an option" an option? not sure if worth it?

IMGMAKE help output - now the main parameters are not visible on the last screen:
image

Before, one more line was visible. I would suggest - instead of separate line about VHD size restrictions - let's use the empty space on the "-size" line to mention the VHD restrictions, e.g.
`  -size: Size of a custom hard disk image in MB. For VHD is from 3 to 2088960.`
That seems to fit on the screen... Maybe M/MB will fit also, but I don't think "2040G" will fit.

Original IMGMAKE output - where main parameters are visible on last page:
image

rderooy:
Trying to keep the options on a 80x25 screen gets increasingly difficult, and even harder when you look at translations. With -staging work was done to automatically use `more` to page through them if the text will not fit.

Torinde:
Merging and other improvements: #4308

Torinde:
In addition to the GPT support in Win2003 and 3rd party support in Win9x, there is also a [LBA 33-bit enhancement for DOS](https://www.bttr-software.de/forum/forum_entry.php?id=17268) that would allow 2x2TB partitions (thus requiring VHDX).

PR #4298, opened 2023-05-28
Added info about the internal DEBUGBOX command to the debugger readme.

yuv422:
I was struggling to find info on how to start a program and break on the entry point.
I was looking in `README.debugger` but saw no mention of how to do this.

Just adding info about the `DEBUGBOX` command to help others that might also be searching
for this helpful command.

:)

### DIFF
--- a/README.debugger
+++ b/README.debugger
@@ -78,6 +78,17 @@ view will show 'pf' to indicate this.
   pf = segment:offset or linear address is paged out or
        not present according to page tables.
 
+Debugging a program and breaking on start
+-----------------------------------------
+
+Dosbox-x includes a built-in command to launch a program and
+break at the entry point.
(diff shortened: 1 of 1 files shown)


README.debugger
```diff
@@ -78,6 +78,17 @@ view will show 'pf' to indicate this.
   pf = segment:offset or linear address is paged out or
        not present according to page tables.
 
+Debugging a program and breaking on start
+-----------------------------------------
+
+Dosbox-x includes a built-in command to launch a program and
```
Allofich, line 84: The official name of the program is (as I understand) "DOSBox-X", so it would be good to match the capitalization.

yuv422: Good point. Fixed :)

PR #4333, opened 2023-06-13
Set usescancodes=true when non-US keyboards are detected (Linux/MacOS builds)

maron2000:
With this PR, If `usescancodes=auto` (default), it is set to true when using non-US keyboards in Linux SDL1 builds.
This behavior is consistent with what is mentioned in `dosbox-x-reference.conf`, and the same is already set for Windows build in PR #4037.


Torinde:
Is that needed also for macOS and HX DOS?

maron2000:
Well certainly yes, I don't know how to do it for those OSs.
However it is not a must but is better to be fixed.

PR #4352, opened 2023-06-26
Build nightly setup file for Windows

maron2000:
Current CI workflow for Windows setup can only build release versions only.
This PR enables building nightly builds as well.
Visual Studio (Intel, ARM) and MinGW 32-bit and 64-bit builds are packed into a single setup file.

### DIFF
--- a/.github/workflows/windows-installers.yml
+++ b/.github/workflows/windows-installers.yml
@@ -2,97 +2,368 @@ name: Windows installers
 
 on:
   push:
+  pull_request:
   workflow_dispatch:
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


maron2000:
Sorry for the errors. 
It's now ready to be merged.
We have to see if uploading release version works as expected.

Torinde:
Great!
Can you add inside also the MinGW-lowend build? Release 32-bit installer has it inside.

I'll try testing in various Windows hosts later...

maron2000:
MinGw  lowend builds are already included in the XP installer.

PR #4451, opened 2023-09-05
HX-DOS fix pre-upload testing failure

maron2000:
HX-DOS builds tests itself by running it on VS builds before Release/Development builds are uploaded.
Upon building release versions, test may fail because VS release builds are yet to be built and missing, therefore wait for sometime to finish building it.


Torinde:
Let's see how it goes. Last run seems to be ~50mins, so I would suggest to put a delay of 1 hour, just to be sure?
image

maron2000:
Actual required time to build 32bit VS builds is around 25min. but sometimes jobs get queued for sometime.
You need 15-20min. for the HX-DOS build to finish, and wait additional 40 min. which means around 55-60min., so it shouldn't be so short.
If we often see VS builds to take more than an hour, we may omit the testing since it is tested every time it is committed during the development stage.

PR #4458, opened 2023-09-10
INSTALL.md - MinGW-lowend include debugger as of #4456

Torinde:


maron2000:
I think this should be fixed not now but in the next release, since it is now only included in the nightly builds.

Torinde:
Agree, changed title to DRAFT.

Torinde:
Now that the next version was released, this PR can be merged.

PR #4491, opened 2023-09-25
Small fixes and enhancements in VHD code

maxpat78:
- ConvertFixed in bios_vhd.cpp can now do better partition detection
- enforced checks in VHDMAKE -convert
- cosmetic changes in VHDInfo struct
- updated help strings (VHDMAKE)

### DIFF
--- a/include/bios_disk.h
+++ b/include/bios_disk.h
@@ -322,15 +322,15 @@ public:
         bool IsValid();
         void SetDefaults();
     };
-    typedef struct _VHDInfo {
+    struct VHDInfo {
         uint32_t allocatedBlocks;
         uint32_t totalBlocks;
         uint32_t blockSize;
         float vhdSizeMB;
         uint32_t vhdType;
-        _VHDInfo *parentInfo = NULL;
+        VHDInfo *parentInfo = NULL;
         std::string diskname;
-    } VHDInfo;
+    };
     VHDTypes vhdType = VHD_TYPE_NONE;
 	virtual uint8_t Read_AbsoluteSector(uint32_t sectnum, void * data);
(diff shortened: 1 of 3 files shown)


maron2000:
Great work!

BTW, I see the warning below, but I think contributors should make the code compatible with c++11.
You can find such comment in configure.ac, and maybe somewhere else as well.
@joncampbell123 do you have any further comments?

```
dos_programs.cpp:7806:19: warning: ISO C++11 does not allow conversion from string literal to 'char *' [-Wwritable-strings]
    char* sizes = "BKMGT";
                  ^
```

joncampbell123:
> Great work!
> 
> BTW, I see the warning below, but I think contributors should make the code compatible with c++11. You can find such comment in configure.ac, and maybe somewhere else as well. @joncampbell123 do you have any further comments?
> 
> ```
> dos_programs.cpp:7806:19: warning: ISO C++11 does not allow conversion from string literal to 'char *' [-Wwritable-strings]
>     char* sizes = "BKMGT";
>                   ^
> ```

It might be possible to silence that warning by changing it to:

    const char* sizes = "BKMGT";

maron2000:
Just tried on MinGW build but seems good.

Current code
before

After changing to `const char* sizes = "BKMGT";`, no warning regarding line 7806
after

maxpat78:
@maron2000 @joncampbell123 I've noticed the "Not everyone wants to put underscores" patch... I suppose it originated from my access/_access and strdup/_strdup change in code to satisfy VC++? What is the more C++-in-DOSBox-X conformant way to write that?

maron2000:
@maxpat78  The fix by @joncampbell123 was required to build successfully on Linux and mac, while your original code failed.
https://github.com/joncampbell123/dosbox-x/actions/runs/6301953916

PR #4498, opened 2023-09-29
Add warnings to IMGMOUNT command

maron2000:
This PR warns users if trying to mount an unformatted drive or image files with unsupported extensions.

imgmount_warnings


rderooy:
I did not check the code, but did you test this with PC booter games that do not have a FAT file system? Or is this only meant to work on HDD images?

maron2000:
This PR is merely a warning and will never reject mounting due to whether it is FAT formatted or not.
The unformatted warning will show only when you IMGMOUNT with drive letters, and if the game is not FAT formatted, I believe you have to mount it with drive numbers.
I can try if you can raise a paricular game in mind.

Issue #3577, opened 2022-06-16
DOSbox running under Win9x, WinNT, Win3x


Torinde:
### Question

Per the [DOSBox-X website ](https://dosbox-x.com)it requires as a minimum for 32-bit WinXP and for 64-bit Vista. Here is what happened when I tried 0.84.0 on earlier OS versions (hosted in VirtualBox, all systems have normal sound in Windows; DLLs are listed in the order of appearing and the last one shown is the one I couldn't solve):
- Windows XP x64: DOSBox-X win32 VS SDL1 works (also #2235)
- ReactOS 32-bit: DOSBox-X works, but sound very low (also #1269)
- Windows 2000: 
-- DOSBox-X mingw-lowend works, but no sound (also #582)
-- DOSBox 0.74-3 works with sound
- Win NT4 SP6a: 
-- DOSBox-X error messages - first about IMM32.DLL (resolved by placing it in the DOSbox-X folder), then user32.dll (couldn't resolve)
-- DOSBox 0.74-3 error message about shell32.dll, works after installing IE4SP2, but sound mediocre and delayed
- Win NT3.51:
-- DOSBox-X: error messages DSound.dll, MSVCRT.dll, IMM32.dll, user32.dll
-- DOSBox 0.74-3 error messages: MSVCRT.dll, OPENGL32.DLL, SHELL32.DLL (IE4SP2 / IE5 for Win3.x/NT3.x is 16-bit and doesn't have Active desktop shell32.dll)
- Win ME:
-- DOSBox-X: error message KERNEL32.DLL
-- DOSBox 0.74-3 works with sound
- Win 98 First edition:
-- DOSBox-X: error message KERNEL32.DLL, wasn't resolved by [KernelEx](http://kernelex.sourceforge.net), although maybe there is a way?
-- DOSBox 0.74-3 works, but no sound
- Win 95 RTM + Winsock2 + OpenGL95 + IE4SP2 + DirectX:
-- DOSBox-X: error message IPHLPAPI.DLL, DHCPCSVC.DLL, NTDLL.DLL (resolved by replacing the one in WINDOWS\SYSTEM), KERNEL32.DLL
-- DOSBox 0.74-3 works, but sound very choppy
- Win NT3.5, NT3.1 - not tested after stuck at NT3.51
- Win 3.1x with Win32S (and WinG, Video for Windows, etc.) has an early version of SHELL32.DLL - not tested after stuck at NT3.51
[...]

PR #4612, opened 2023-11-18
Enables Win9x support on a Pentium3 PC

crazii:
## Enables Win9x support on a Pentium3 PC.

- Add _WIN32_WINDOWS macro to detect win9x target platform
- Add OpenCOW (static lib) to support Unicode functions
  Unicode functions (e.g. GetOpenFileNameW to choose dosbox-x.conf) which is not originally supported on Win9x, it just fails silently with no error code. (similar found in code: FindFirstFileW/FindNextFileW, GetFileAttributeExW, InsertMenuItemW)
  For MSVCRT it is the same, e.g. wstat/wopen won't work. mounted drive in DOSBox-X will fail getting file information and result in empty mounted disks.
  The OpenCOW lib will translate *W calls to *A calls: adjust wide strings to ANSI string with current system code page and call the corresponding ANSI function, and make the win9x build compatible with Unicode functions.
- Use 32bit file offsets
  Win9x uses FAT32 which doesn't need 64bit file offset, so it is not supported.
- New build script: build-mingw-lowend9x (SDL1), no SDL2 support.

## Additional information
A custom toolchain that targeted to pentium3 is needed (https://github.com/crazii/MINGW-toolchains-w9x/releases/tag/v1.0), 
otherwise the built binary can not run.
  Tested working on a real Pentium3 Win98SE laptop.


src/gui/render.cpp
crazii, line 1: Changing from __SSE__  to __SSE3__ should be incorrect, at least it should be __SSE2__. if uses __SSE__ the dynamic branch of SSE2 is OK but it still got compiled, when targeting to pentium3 it failed to compile.
add target specific _WIN32_WINDOWS should be OK.

Torinde:
Related: #3577
- will this build also run on NT4 or NT3.x?
- I assume it won't be able to run under Win3.1x/Win32S? That would've also allowed to run under OS/2 or ArcaOS...

crazii:
NT4 has similar API set to Win9x, so at least It's possible to make it run, with a few changes if needed. But I didn't test it yet, I don't have legacy PCs that run NT4.
But NT3.x and Win3.x don't.

Also I need correct my previous statement about the custom toolchain, it is essential because the original MSYS2 mingw32-crt startup routine will call some inline NT functions and corrupt the memory. Maybe an older mingw32 will help. What edition of Mingw32 is used to build DOSBox-X lowend exactly?

And I've found some new problems and still been testing, for sometimes the build exe fails with page error on start, and it always does. but another build will always run. This only happens on my Penntium3 laptop, but works well in VMWare Player (with modern CPU features, which indicate the problem lays on instruction sets). - So it is not ready for PR yet.

I don't know about Github Actions, but I can try to make one when the build is stable.

Torinde:
> call some inline NT functions and corrupt the memory. Maybe an older mingw32 will help.

I don't know what's used for the MinGW-lowend build, but there is also a HX DOS build, which I think doesn't use any NT functions.

>  I don't have legacy PCs that run NT4.

I can test in VirtualBox - when you have an executable ready.

crazii:
the executable is here: https://github.com/crazii/dosbox-x-w9x/releases/tag/beta

joncampbell123:
Ready for me to merge your pull request or should I wait for some testing to complete first?

Torinde:
I did some tests - results are in #3577, basically the 9x build works under 98/Me/2000/11 and ReactOS.

Doesn't work under 95 or NT4 (so I haven't tested NT3.x and Win32S) - can that be corrected?

Would be nice to default into video **output surface** and to get the **debugger** enabled.

Besides Github **Actions**, would be good if it gets an **Installer** (the NSIS used for XP/Vista installers supports [down to Win95/NT](https://nsis.sourceforge.io/Features)) and added to the DOSbox-X **Release** scripts

joncampbell123:
> I did some tests - results are in #3577, basically the 9x build works under 98/Me/2000/11 and ReactOS.
> 
> Doesn't work under 95 or NT4 (so I haven't tested NT3.x and Win32S) - can that be corrected?
> 
> Would be nice to default into video **output surface** and to get the **debugger** enabled.
> 
> Besides Github **Actions**, would be good if it gets an **Installer** (the NSIS used for XP/Vista installers supports [down to Win95/NT](https://nsis.sourceforge.io/Features)) and added to the DOSbox-X **Release** scripts

I suppose an #ifdef could be added to default to surface output if compiling the lowend version for systems below XP. Not sure if pdcurses can run on an OS that low. Is the missing component that prevents Windows 95/NT4 MSVCRT.DLL? That seems to be a common thing with MinGW compiled binaries in that they require MSVCRT.DLL which is provided by default on Windows 98/2000 on up.

Another idea is that if pdcurses can't compile for Windows that old, perhaps the SDL interface itself could provide the debugger interface. In the same way that when you switch to the mapper interface the SDL window changes to an 8bpp mode to render the mapper buttons and such. Then the debugger interface could call some code that does very minimal ncurses emulation on the SDL window when active.

Torinde:
Win 95 error messages are for IPHLPAPI.DLL (resolved by taking it from WinME), DHCPCSVC.DLL (resolved by taking it from WinME), NTDLL.DLL (resolved by replacing the one in WINDOWS\SYSTEM with one from WinME), unresolved for KERNEL32.DLL (Windows can't boot if the file is replaced with the one from WinME)

Win NT4 first error is about IMM32.DLL (resolved by placing it in the DOSbox-X folder), then user32.dll (couldn't resolve)

Error message sometimes mention specific function inside the DLL that's required. I will make screenshots of those later.

Torinde:
NSIS installer - from the link above (which I think is the official/regular one used already by DOSBox-X) it seems to already support Win95/NT (maybe even NT3.x, because number isn't mentioned):
image

Torinde:
Win 95 RTM + Winsock2 + OpenGL95 + IE4SP2 + DirectX:
image

image

image

image


Win NT4 SP6a
image

image

crazii:
> Ready for me to merge your pull request or should I wait for some testing to complete first?

No yet, it needs more tests.

crazii:
> I haven't tested with Microsoft Layer for Unicode, but it's not expected to help since the 9x build already has OpenCOW, correct?

Yes. Thanks for the testing, I will try to fix it on win95 in VM first. Can you at someone who's familiar with the toochains used to build the lowend binary? I could check if that helps.

Torinde:
@joncampbell123 @maron2000, can you please tell:

> What edition of Mingw32 is used to build DOSBox-X lowend exactly?

joncampbell123:
> @joncampbell123 @maron2000, can you please tell:
> 
> > What edition of Mingw32 is used to build DOSBox-X lowend exactly?

This would be the original 32-bit MinGW (not the MinGW x64 which seems to support only Windows Vista and up).
The original MinGW 32-bit compiler, last I checked sometime around 2017, is still able to target as low as Windows 98 (Windows 95 if you install MSVCRT.DLL).

joncampbell123:
You should be able to comment out an #ifdef to disable the IMM32.DLL dependency. Some contributors have added IME support especially where it concerns Japanese/Chinese keyboard input and/or PC-98 mode.

You could also comment out the "multiple monitor" support code to eliminate the USER32.DLL dependency error. DOSBox-X is only trying to use that to determine your monitor's DPI and which monitor it's on. To my recollection, the multiple monitor API didn't appear in Windows 9x until Windows 98 when it was one of the Big New Features added by Microsoft. Not sure about when Windows NT added it (maybe it was added in Windows 2000?).

joncampbell123:
Come to think of it, I already have plenty of LoadLibrary and GetProcAddress() for more recent Windows APIs like the one in Windows 7 to add buttons and define the portion of your window shown when you hover over the app icon on the taskbar, perhaps I should LoadLibrary/GetProcAddress IME support and multiple monitor support too when targeting Windows.

crazii:
OK, I was using the customized 32bit toolchain of MINGW-w64, I'll try to build a new binary with the legacy toolchain from SF.

dbjh:
> OK, I was using the customized 32bit toolchain of MINGW-w64, I'll try to build a new binary with the legacy toolchain from SF.

You can use that, or a slightly more recent version from the official site https://mingw.osdn.io/. However, for me the official site is unreachable at the moment. As an alternative you can also use https://osdn.net/projects/mingw/.

joncampbell123:
> > OK, I was using the customized 32bit toolchain of MINGW-w64, I'll try to build a new binary with the legacy toolchain from SF.
> 
> You can use that, or a slightly more recent version from the official site https://mingw.osdn.io/. However, for me the official site is unreachable at the moment. As an alternative you can also use https://osdn.net/projects/mingw/.

It seems like the original MinGW site has gotten more and more inaccessible since about 2017 when I noticed the GUI to install it's packages stopped working.

maron2000:
FYI, the mingw32 environment used to build the x86 lowend builds is found [here](https://github.com/joncampbell123/dosbox-x/blob/master/build-scripts/mingw/lowend-bin/i686-7.3.0-release-posix-dwarf-rt_v5-rev0%2Bnasm.7z)

crazii:
Thank  you guys, I've finished the build (with extra modifications of the source code) using toolchains from both SF and OSDN, tested good in win95, 98 and NT4.
Now I'm gonna try the mingw32 that maron2000 mentioned.

crazii:
I ran a simple & fast test on the toolchain used to build lowend target (it seems a i686-w64-mingw32 for ming-w64), under win98 there's a missing function AddVectoredExceptionHandler in kernel32.dll, which is XP(or maybe 2k) only.
That seems to be  fixed in latest mingw32 which uses LoadLibrary last time I checked, when customizing the i686-w64-mingw32 toolchain.
Conclusion: no luck.

For now the legacy MinGW from SF/OSDN is a sound choice. The built binary runs on win98 p3 real pc or win98/95osr2.5/nt4sp1 in VirtualBox. But I have to use Pentium instruction instead of p3, for win95 will fail on a invalid instruction of SSE1 (win98 works fine), didn't figure out why, but it shouldn't happen - Virtualbox won't limit such CPU caps on guest OS.

Torinde:
Some info on [SSE in Win95](https://github.com/JHRobotics/mesa9x).

Can you make an executable with the latest changes?

joncampbell123:
> I ran a simple & fast test on the toolchain used to build lowend target (it seems a i686-w64-mingw32 for ming-w64), under win98 there's a missing function AddVectoredExceptionHandler in kernel32.dll, which is XP(or maybe 2k) only. That seems to be fixed in latest mingw32 which uses LoadLibrary last time I checked, when customizing the i686-w64-mingw32 toolchain. Conclusion: no luck.
> 
> For now the legacy MinGW from SF/OSDN is a sound choice. The built binary runs on win98 p3 real pc or win98/95osr2.5/nt4sp1 in VirtualBox. But I have to use Pentium instruction instead of p3, for win95 will fail on a invalid instruction of SSE1 (win98 works fine), didn't figure out why, but it shouldn't happen - Virtualbox won't limit such CPU caps on guest OS.

Windows did not support SSE extensions until Windows 98 SE. Any lower version of Windows does not enable the SSE extensions and applications will crash if they try to use SSE extensions.

joncampbell123:
Here's an assembly language example of enabling SSE extensions on a processor that supports it:

https://github.com/joncampbell123/doslib/blob/master/hw/cpu/cpusse.c#L505

See also:

https://wiki.osdev.org/CPU_Registers_x86#CR4

Note that enabling FXSAVE/FXRSTOR also enables SSE instructions.

maron2000:
Refer to #4409 to enable debugger functions on mingw builds.

crazii:
> Refer to #4409 to enable debugger functions on mingw builds.

Thanks! I'll check that out and see what happens.

crazii:
The debugger window now can be shown. didn't test much, and there's some questions that I'm not quite sure about:

After libncurses is installed, configure cannot find it, it seems MinGW have a wide-char version - libncursesw, So I remove libncurses and use libpdcurses instead.

There's some API used by google test (gtest) that not available on win9x (OpenThread), or not on win95 only(InterlockedCompareExchange), after removing those calls with #ifdef, dosbox-x can run but will crash on quit, in kernel32.dll (after a few times it won't start, and then a popup about error in kernel32.dll and system freeze).  I don't think it is the removed code that causes the problem, since the gtest code is not executed. or there are static objects in gtest code that cause problem. didn't investigate much so I just remove the whole gtest thing using #ifdef, then the problem is gone.
Is google test used now? if yes, does the test run on a daily basis? is it needed for lowend builds?

Torinde:
Tests with the second beta - #3577 

Win 95 **RTM** + Winsock2 + **OpenGL95** + IE4SP2 + DirectX - works
Win NT4 - works
Win NT3.51 - error about IMM32.DLL (passed by providing the DLL), USER32.DLL (entry point MonitorFromWindow)

I'll test Win32s next.

joncampbell123:
> Tests with the second beta - #3577
> 
> Win 95 **RTM** + Winsock2 + **OpenGL95** + IE4SP2 + DirectX - works Win NT4 - works Win NT3.51 - error about IMM32.DLL (passed by providing the DLL), USER32.DLL (entry point MonitorFromWindow)
> 
> I'll test Win32s next.

There are parts of DOSBox-X (I think most of it in the SDL library) that use multithreading, which is not supported in Win32s. It helps to think of Win32s as a thin Win32 to Win16 thunk layer for Windows 3.1. Also Windows 3.1 uses cooperative multitasking *including* Win32s and if DOSBox-X gets stuck everything else stops responding as well. Fair warning.

joncampbell123:
I know this from long experience writing the stuff in DOSLIB and from an aborted 32-bit games jam entry I started but never finished beyond the "I got video to work and sprites with a transparency mask to blit using images I loaded using a custom PNG reader I wrote for DOS" phase. The Win32 build was written explicitly to be able to run under Windows 95 and Win32s, that's how I know about the SetDIBitsToDevice limitation where negative biHeight values for top down (instead of bottom up) bitmaps are not supported. This was one of the few video support routines that was able to compile for DOS using VESA BIOS extensions and for anything else (like Mac OS) by simply linking to SDL2.

joncampbell123:
Also from experience: There are plenty of functions that have to be there for Win32 apps to work at all in Windows 3.1 that are really just stubs returning a fixed error code. CreateDIBSection() for example.

crazii:
What's the problem of the Linux/MacOS CI build check? I've no idea what happens, probably due to some changes on the configure.ac?

joncampbell123:
> What's the problem of the Linux/MacOS CI build check? I've no idea what happens, probably due to some changes on the configure.ac?

It can't find the ".in" file :laughing: because it forgot that "configure" is part of it.

crazii:
> > What's the problem of the Linux/MacOS CI build check? I've no idea what happens, probably due to some changes on the configure.ac?
> 
> It can't find the ".in" file 😆 because it forgot that "configure" is part of it.

Yes I can see that in the check log, only a .in, but why?
Is this related to my changes?

crazii:
The source code itself is ready. There's my questions about CI workflow

I checked the yml for build lowend, it runs under msys2, while the legacy MinGW runs on msys1, I don't know whether it works if the shell is changed to MinGW or msys, and the package installing part, should a pre-made zip be used (like mingw lowend zip) or install packages on the fly?  I should test it on my fork to make sure everything works. But I don't know how.

Should I enable the workflow on my fork and add a new one to test? The original workflows are disable by github, and I don't know the consequences if it is enabled.

Anyone can help with this?

maron2000:
You can take a look at `.github/workflows/mingw32.yml`
Lowend builds now replace the mingw environment.
You can follow that or make a workflow to tweak the latest mingw as you already did in order to build your own executables.

dbjh:
> What's the problem of the Linux/MacOS CI build check? I've no idea what happens, probably due to some changes on the configure.ac?

You converted configure.ac to DOS text format (\r\n). If you convert it back to UNIX format (\n) the problem will be solved.

Torinde:
WfW 3.11 + Win32s 1.30c: goes trough IMM32.DLL, MSVCRT.DLL, OPENGL32.DLL, GLU32.DLL, DCIMAN32.DLL, OLE32.DLL and then reaches unresolvable 'system component out of date' and w32scomb.dll (see at https://github.com/joncampbell123/dosbox-x/issues/3577#issuecomment-1837296222)

So, it seems NT 3.51 is much closer to working (IMM32.DLL, then USER32.DLL). Would you be able to adapt the 9x/NT4 to build to work on NT 3.x?

dbjh:
> > > What's the problem of the Linux/MacOS CI build check? I've no idea what happens, probably due to some changes on the configure.ac?
> > 
> > 
> > You converted configure.ac to DOS text format (\r\n). If you convert it back to UNIX format (\n) the problem will be solved.
> 
> Thanks. I didn't remember doing that. Doesn't git always use LF in remote? I've seen articles about that, there's a auto CRLF option for windows local clones.

Could you create a pull request with the fix? It appears no one else will do that and the issue breaks the build on at least Linux and macOS (for everyone, not just for CI).

crazii:
> > > > What's the problem of the Linux/MacOS CI build check? I've no idea what happens, probably due to some changes on the configure.ac?
> > > 
> > > 
> > > You converted configure.ac to DOS text format (\r\n). If you convert it back to UNIX format (\n) the problem will be solved.
> > 
> > 
> > Thanks. I didn't remember doing that. Doesn't git always use LF in remote? I've seen articles about that, there's a auto CRLF option for windows local clones.
> 
> Could you create a pull request with the fix? It appears no one else will do that and the issue breaks the build on at least Linux and macOS (for everyone, not just for CI).

Sorry for the inconvenience. https://github.com/joncampbell123/dosbox-x/pull/4651

crazii:
> WfW 3.11 + Win32s 1.30c: goes trough IMM32.DLL, MSVCRT.DLL, OPENGL32.DLL, GLU32.DLL, DCIMAN32.DLL, OLE32.DLL and then reaches unresolvable 'system component out of date' and w32scomb.dll (see at [#3577 (comment)](https://github.com/joncampbell123/dosbox-x/issues/3577#issuecomment-1837296222))
> 
> So, it seems NT 3.51 is much closer to working (IMM32.DLL, then USER32.DLL). Would you be able to adapt the 9x/NT4 to build to work on NT 3.x?

Not for now probably. I wonder how many users in real world runs DOSBox-X under a pre 2k era machine. Me for instance, I've a win98 laptop with HW SBPro and soft FM emulation by the vxd driver, but a few games have problem with the FM emulation and they have no sound, and after they exit, the system get muted. After test them on DOSBox in win98 I got a satisfied result, then I want to see what happens if with the OPL hardware passthrough of DOSBox-X? But it fails to run. That's my initial intention of the fork. Those games runs with sound in DOSBox-X with sound in hardware OPL mode, but after exit DOSBox-X, the system still got muted. it's the bad driver of the host.
DOSBox-X is more powerful than DOSBox especially running win9x guests, but on a Win9x system, there's no need of that, and most games runs OK natively (except for some early DOS games). and the need of running DOSBox-X on NT3.51 is little.

crazii:
OK, after reading the yml carefully, I get that the setup step uses a bash shell which is separated from msys2 shell, so msys2 has wget but bash env hasn't.

dbjh:
> > WfW 3.11 + Win32s 1.30c: goes trough IMM32.DLL, MSVCRT.DLL, OPENGL32.DLL, GLU32.DLL, DCIMAN32.DLL, OLE32.DLL and then reaches unresolvable 'system component out of date' and w32scomb.dll (see at [#3577 (comment)](https://github.com/joncampbell123/dosbox-x/issues/3577#issuecomment-1837296222))
> > So, it seems NT 3.51 is much closer to working (IMM32.DLL, then USER32.DLL). Would you be able to adapt the 9x/NT4 to build to work on NT 3.x?
> 
> Not for now probably. I wonder how many users in real world runs DOSBox-X under a pre 2k era machine. Me for instance, I've a win98 laptop with HW SBPro and soft FM emulation by the vxd driver, but a few games have problem with the FM emulation and they have no sound, and after they exit, the system get muted. After test them on DOSBox in win98 I got a satisfied result, then I want to see what happens if with the OPL hardware passthrough of DOSBox-X? But it fails to run. That's my initial intention of the fork. Those games runs with sound in DOSBox-X with sound in hardware OPL mode, but after exit DOSBox-X, the system still got muted. it's the bad driver of the host. DOSBox-X is more powerful than DOSBox especially running win9x guests, but on a Win9x system, there's no need of that, and most games runs OK natively (except for some early DOS games). and the need of running DOSBox-X on NT3.51 is little.

The OPL pass-through code is not enabled by default, but your use-case may be a reason to change that. In order to enable the code change line 6 of src/hardware/hardopl.h as indicated in the comment (https://github.com/joncampbell123/dosbox-x/blob/master/src/hardware/hardopl.h#L6C12-L6C12).

crazii:
Yes, I checked the source that it is not enabled, I've enabled it on my local builds (didn't want to commit to the repo).
Funny that the hardware OPL passthrough has worse quality than emulated by DOSBox-X, since the "hardware" OPL on host is a poor emulation by the driver, worse that DBOPL emulation. But still way better than the ESS allegro/maestro or some wavetable emulation.

joncampbell123:
> WfW 3.11 + Win32s 1.30c: goes trough IMM32.DLL, MSVCRT.DLL, OPENGL32.DLL, GLU32.DLL, DCIMAN32.DLL, OLE32.DLL and then reaches unresolvable 'system component out of date' and w32scomb.dll (see at [#3577 (comment)](https://github.com/joncampbell123/dosbox-x/issues/3577#issuecomment-1837296222))
> 
> So, it seems NT 3.51 is much closer to working (IMM32.DLL, then USER32.DLL). Would you be able to adapt the 9x/NT4 to build to work on NT 3.x?

Here's an article from IBM who, unlike Microsoft, doesn't throw away their old Knowledge Base articles:

https://web.archive.org/web/20231203070203/https://www.ibm.com/support/pages/win32s-errors

Apparently something to do with OLE32.DLL?

crazii:
> > I wonder how many users in real world runs DOSBox-X under a pre 2k era machine.
> > the need of running DOSBox-X on NT3.51 is little.
> 
> There's a community of NT3.51 enthusiasts, for example [Running NT 3.51 on real new hardware (Ryzen 3900X)](http://www.win3x.org/win3board/viewtopic.php?t=27197&sid=4884df3157479ee7fbf647e4a273678b), [Gaming on Windows NT 3.51](http://www.win3x.org/win3board/viewtopic.php?t=19219) and [VOGONS](https://www.vogons.org/viewtopic.php?f=5&t=84474)
> 
> DOSBox 0.65 is reported as the last version working under NT3.51, so those users would benefit greatly if latest DOSbox-X lowend9x can be made to run there.
> 
> (I also wonder if NewShell patches will help or allow NT4 Internet Explorer to be installed, which may also help)

OK, cool.
I can figure out a time to test it on NT3.51, when I've finished my work at hand, it might take long, please wait :)

Torinde:
WinG is supported in NT3.5, so a build with the above SDL1.2 for Win32s/Win3.1x would cover also NT3.5 and most probably the rest of the NT versions other than NT3.1

PR #4623, opened 2023-11-21
Website - XP build works under ReactOS

Torinde:
See https://github.com/joncampbell123/dosbox-x/issues/3577#issuecomment-1821801793

### DIFF
--- a/_includes/version.html
+++ b/_includes/version.html
@@ -22,7 +22,7 @@
 <h4>Windows version:</h4>
 <h5 class="subtitle-mute">(Windows XP and later versions supported)</h5>
 <h5>
-  32/64-bit Setup {% include version_file.html version=version name="winXP_setup_file" display_name=" XP+ " %}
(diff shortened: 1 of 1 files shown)


maron2000:
It's ok to test and report that it works on ReactOS, but I oppose to mention that it is as if it is officially supported.

joncampbell123:
> It's ok to test and report that it works on ReactOS, but I oppose to mention that it is as if it is officially supported.

You could list it as supported, but experimental, with no warranty, etc.

maron2000:
Thanks. I'll leave the decisions to @Torinde and @joncampbell123 .
Sorry for bothering.

Torinde:
OK, maybe a compromise would be to mention ReactOS as experimental/unofficial on the XP line of INSTALL.md? ReactOS users probably are savvy enough to try themselves the XP version and/or to read the Install instructions.

I made another PR for INSTALL.md, so @joncampbell123 can choose whether to merge this PR, the other PR or both.

We can also add a <abbr title="Experimental support">tooltil</abbr> with "abbr" stating "Experimental support" on the website front page - should I try making such PR?

PR #4632, opened 2023-11-26
replace sOffset (use offsetof)

jg1uaa:
Building DOSBox-X on OpenBSD/amd64 (clang-13), I see many -Wnull-pointer-subtracion warnings like this:

```
../../include/dos_inc.h:542:35: warning: performing pointer subtraction with a null pointer may have undefined behavior [-Wnull-pointer-subtraction]
        void SetBootDrive(uint8_t drv) { sSave(sDIB,bootDrive,drv); }
                                         ^~~~~~~~~~~~~~~~~~~~~~~~~
../../include/dos_inc.h:395:61: note: expanded from macro 'sSave'
#define sSave(s,m,val) SaveIt(sizeof(((s *)&pt)->m),(PhysPt)sOffset(s,m),val)
                                                            ^~~~~~~~~~~~
../../include/dos_inc.h:393:46: note: expanded from macro 'sOffset'
#define sOffset(s,m) ((char*)&(((s*)NULL)->m)-(char*)NULL)
                                             ^~~~~~~~~~~~
```
I think sOffset() in dos_inc.h can be replaced with offsetof() and it can reduce this warning. Is there any historical reason to keep this code?

(in OpenBSD's stddef.h, offsetof() is defined like this)
```
#if __GNUC_PREREQ__(4, 0)
#define offsetof(type, member)  __builtin_offsetof(type, member)
#else
#define offsetof(type, member)  ((size_t)(&((type *)0)->member))
#endif

```

### DIFF
--- a/include/dos_inc.h
+++ b/include/dos_inc.h
@@ -390,7 +390,7 @@ static INLINE uint16_t DOS_PackDate(uint16_t year,uint16_t mon,uint16_t day) {
 
 
 /* Remains some classes used to access certain things */
-#define sOffset(s,m) ((char*)&(((s*)NULL)->m)-(char*)NULL)
+#define sOffset(s,m) offsetof(s,m)
 #define sGet(s,m) GetIt(sizeof(((s *)&pt)->m),(PhysPt)sOffset(s,m))
 #define sSave(s,m,val) SaveIt(sizeof(((s *)&pt)->m),(PhysPt)sOffset(s,m),val)
 


maron2000:
I didn't try but one of the concern is that whether the code is C++11 compatible.
The following is sited from the contribution guide.
```
DOSBox-X is still intended to support compilation as C++11. If you add new code please make sure it works as C++11. One way to do so is to run ./configure with --enable-force-cxx11 option added, and build the code.
```

jg1uaa:
I checked this change does not make error to compile with gcc-8.4.0 and clang-13 on OpenBSD,
with or without configure --enable-force-cxx11.

at least, dos_inc.h has already included <stddef.h> to use offsetof().
https://github.com/joncampbell123/dosbox-x/blob/master/include/dos_inc.h#L28

joncampbell123:
offsetof() is way older than C++11, it shouldn't be a problem.

https://en.cppreference.com/w/cpp/types/offsetof

Issue #4644, opened 2023-12-01
Variable in DOS_Int21_71a6() dereferenced past its size


SmileTheory:
https://github.com/joncampbell123/dosbox-x/blob/f8e923696c29760aae974e9444229ed210d97cb9/src/dos/dos.cpp#L5357C204-L5357C204

On this line, `handle` is `uint8_t`, but the `sprintf()` call casts its address to a `char *` and attempts to pull four characters out of it.

Spotted this while looking at https://github.com/joncampbell123/dosbox-x/issues/4538 , don't know if it is the error, but it is at least an error.

joncampbell123:
Good catch. Perhaps it's a programmer mistake considering all the other parameters are (char*)&something.

SmileTheory:
https://github.com/joncampbell123/dosbox-x/commit/505231b81b6c43681190ec0e82b1bc7c412818e9 fixes this bug, but in investigating it I found another one, for which I have submitted a pull request in https://github.com/joncampbell123/dosbox-x/pull/4647 .

Issue #4538, opened 2023-10-16
SH.EXE hangs DOSBox-X 0.84.0


eltoneo:
### Describe the bug

Hi,
'BSH205BBR4.ZIP's SH.EXE' hangs DOSBox-X 0.84.0! All newer DOSBox-X version's have serious 'install' problems with macOS Monterey.
https://www.delorie.com/pub/djgpp/deleted/v2gnu/

CRASH REPORT:
DosBox quit unexpectedly.
Click “Send to Apple” to submit the report to Apple. This information is collected anonymously.
Problem Details and System Configuration
Process: DosBox [3832]
Path: /Applications/dosbox-x.app/Contents/Mac0S/DosBox
Identifier: com. dosbox—x
Version: 0.84.0 (0.84.0)
Code Type: X86-64 (Native)
Parent Process: launchd [1]

Date/Time: 2023-10-15 19:01:10.0844 -0500
OS version: macOS 12.6.8 (21G725)

Thread 0 Crashed:: Dispatch queue: com.apple.main—thread
0 libsystem_kernel.dylib: 0x7ff80e294Ffe __pthread_kill + 10
1 libsystem_pthread.dylib: 0x7ff80e2cbiff pthread_kill + 263
2 libsystem_c.dylib: 0x7ff80e216dbe __abort + 139
3 libsystem_c.dylib: 0x7ff8@e202b8c __stack_chk_fail + 100
4 DosBox: 0x10846f765 DOS_Int21_71a6(char const*, char const*) + 1125
5 DosBox: 0x10847af68 DOS_21Handler() + 22184
6 DosBox: 0x1083afd7c Normal_Loop() + 428
7 DosBox: 0x1@83afec6 DOSBOX_RunMachine() + 38
8 DosBox: 0x1089f6a53 CALLBACK_RunRealInt(unsigned char) + 83
9 DosBox: 0x108603394 DOS_Shell::Execute(charx, char const*) + 2436
10 DosBox: 0x1085efa51 DOS_Shell: :DoCommand(char*x) + 1713
11 DosBox: 0x1@85c8cfd DOS_Shell::ParseLine(charx) + 2029

### Steps to reproduce the behaviour

Install DJGPP using their zip picker and then install UNIX app sh.exe and its dependencies.

### Expected behavior

DOSBox-X aborts

### What operating system(s) this bug have occurred on?

macos Monterey

### What version(s) of DOSBox-X have this bug?

[...]

PR #4645, opened 2023-12-02
Fix int 21h LFN function

nanshiki:
Fixed a bug in which 2 extra bytes were copied to the buffer in int 21h ax=714eh,714fh.
Fixed wrong value of date/time etc. in file information of int 21h ax=714eh,714fh,71a6h.
Fixed date/time conversion of int 21h ax=71a7h to be correct.


maron2000:
According to the build error, we need to add `#include <time.h>` ?

nanshiki:
#4648
I have included #include <time.h> for now.
I just now noticed #4538 #4644. The timing of the fixes seem to coincide by accident.

PR #4648, opened 2023-12-02
add #include <time.h>

nanshiki:
It did not cause an error in our environment, but I included #include<time.h> anyway.


maron2000:
Yes, it's weird, but seems that old build tools (VS for XP and MinGW lowend) fires an error for some reason.
Anyway, placing the header should work.

PR #4651, opened 2023-12-03
fix CRLF for configure.ac

crazii:
Sorry that the configure.ac is messed up on previous PR.


crazii:
I thought that the first CRLF commit is already merged, so I pushed the CI commit...:D

The CI workflow is tested OK here: https://github.com/crazii/dosbox-x-w9x/actions/runs/7075053040
and the built binary is tested working on win95/98/NT4

PR #4652, opened 2023-12-04
Mention W9x builds in devel-build.html

maron2000:
We now have W9x builds by CI workflow, so mention that in development builds webpage.
Also put the Windows installers before portable builds, since installers are recommended over portable builds.


maron2000:
Added NT4+ as per your suggestion.
I assume release builds will be available, so we can modify the webpage after it is released.

PR #4677, opened 2023-12-16
Fixed to set the drawing bit of the GDC status.

### LINKED ISSUE #4564: Crash at running Meisou Toshi  in PC-98 mode (it works fine on version 0.84.3)
### Describe the bug

I tried to run the Chinese version of Meisou Toshi by using DOSBox-X, but couldn't load the main menu properly, while using Anex86 or Neko Project II Kai (RetroArch Core) it worked well, I also tested the Japanese version and it has the same problem.
zrEg929
zvsgqqs
4bgWci9

Files here:
https://drive.google.com/file/d/1WpleAEQ0veAIPqqbVm2EnH2t71ODiXsb/view?usp=sharing

Unzip Meisou.zip in the root directory, run `dosbox-x -fastlaunch -conf Meisou\JA.conf` or `dosbox-x -fastlaunch -conf Meisou\ZH.conf`

### Expected behavior

_No response_

Windows 10 21H2 Build 19044.3208

2023.10.06 (64-bit SDL1)

### Used configuration

```ini
[dosbox]
language=ja_JP
machine=pc98

[pc98]
pc-98 anex86 font="Meisou\Chinese Simplified FONT.bmp"

[autoexec]
imgmount C "Meisou\[ZH]Meisou.hdi"
C:
cd \
call AUTOEXEC.BAT
exit
```

### Output log

_No response_

### Additional information

_No response_

nanshiki:
Fix #4564
In case of SDL2, the bit is set for the time corresponding to the number of dots drawn, and in case of SDL1, the bit is set for the number of times specified in the status read.
Also, even if EGC is not enabled for port output, if GRCG and RMW are enabled, it is necessary to make GDC's dot drawing equivalent to EGC.

### DIFF
--- a/include/pc98_gdc.h
+++ b/include/pc98_gdc.h
@@ -174,6 +174,14 @@ struct PC98_GDC_state {
     bool                    doublescan;                 /* 200-line as 400-line */
 
     bool                    dbg_ev_partition;
+    uint8_t                 drawing_status;
+#if defined(C_SDL2)
+    int                     dot_count;
+    uint64_t                drawing_start;
+    uint64_t                drawing_time;
+#else
+    int                     drawing_count;
+#endif
 };
 
 typedef union pc98_tile             egc_quad[4];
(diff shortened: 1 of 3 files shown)


joncampbell123:
Please use PIC_FullIndex(), which is tied to emulator time. Your implementation will produce inconsistent results if emulation is not 100% (cycle count too high or user is fast forwarding).

Issue #4680, opened 2023-12-16
PC-98 Mouse emulation issues


sdsnatcher:
### Describe the bug

When the Falcom Character CG app is run, two different types of issues happen, depending on the mouse driver used.

Case-1: If dosbox-x  boots its own internal HDD and uses its own mouse driver:

- The mouse coordinates are shown correctly at the status line
- But the mouse pointer is invisible
- Mouse button clicks are not registered at all

Case-2: If dosbox-x boots from the floppy disk image and uses the MS-DOS 3.10 and supplied MOUSE.SYS driver:

- The mouse coordinates don't work correctly. The mouse pointer just stays stationary.
- The mouse pointer is now visible
- Mouse button clicks are registered

Note: I tested the case-2 with the Neko Project II emulator, and the mouse worked just fine.

[Falcom Character CG [PC98].zip](https://github.com/joncampbell123/dosbox-x/files/13693703/Falcom.Character.CG.PC98.zip)

### Steps to reproduce the behaviour

Case-1: Internal mouse driver
1. Boot dosbox-x in PC-98 mode without any floppy disks
2. Insert FCCG_SYS.FDI on drive A:
3. On the DOS prompt, type the following commands:
```
A:
FPG
```

Case-2: MS-DOS 3.10 and its MOUSE.SYS driver
1. Boot dosbox-x in PC-98 mode
2. Select the menu option Drive->A->Boot from disk image 
3. Select the FCCG_SYS.FDI disk image
4. Press ENTER when the date and time is requested
5. On the DOS prompt, type the following command:
```
FPG
```

### Expected behavior

The mouse should work correctly on both cases: 

- The mouse pointer should be displayed
- The mouse coordinates should move accordingly
- The mouse button clicks should be correctly registered

### What operating system(s) this bug have occurred on?

macOS 12.7.1

### What version(s) of DOSBox-X have this bug?

https://github.com/joncampbell123/dosbox-x/actions/runs/7191033384

### Used configuration

```ini
Default configuration file, with only machine=pc98 changed.
```

### Have you checked that no similar bug report(s) exist?

- [X] I have searched and didn't find any similar bug report.

### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

maron2000:
At least booted MS-DOS seems to work.
Are you sure that you enabled mouse capture?
[...]

PR #4697, opened 2023-12-21
Fixed PC-98 mouse cursor movement

### LINKED ISSUE #4585: [PC-98] Mouse pointer cannot leave the left upper quater of screen in Macross: Skull Leader Complete Pack (translated by Nebulous Translations)
### Describe the bug

It means that player can't even to make h(is/er) choice in the title menu and start the game.

1. Select Drive->C->Boot from disk image and specify the game .hdi image.
2. Capture the mouse.
3. Skip the beginning cutscenes.
4. Try to point a mouse to Start menu entry.

### Expected behavior

_No response_

Windows 10 22H2

v2023.10.06, MinGW SDL2, 64-bit

### Used configuration

_No response_

### Output log

_No response_

### Additional information

[Used DOSBOX-X config file.](https://www.mirari.fr/6n9h)

nanshiki:
Fix #4585
It seems that the Macross Skull Leader mouse driver does not reset the counter, so if you limit it to -128 to 127, you cannot move it to the whole thing.
Normal mouse drivers reset the counter after reading the value of the counter, so there should be no problem without value limitation.

### DIFF
--- a/src/hardware/keyboard.cpp
+++ b/src/hardware/keyboard.cpp
@@ -2252,14 +2252,8 @@ uint8_t p7fd9_8255_mouse_latch = 0;
 uint8_t p7fd8_8255_mouse_int_enable = 0;
 
 void pc98_mouse_movement_apply(int x,int y) {
-    x += p7fd9_8255_mouse_x;
-    if (x < -128) x = -128;
-    if (x > 127) x = 127;
-    y += p7fd9_8255_mouse_y;
-    if (y < -128) y = -128;
-    if (y > 127) y = 127;
-    p7fd9_8255_mouse_x = (int8_t)x;
-    p7fd9_8255_mouse_y = (int8_t)y;
+    p7fd9_8255_mouse_x += x;
+    p7fd9_8255_mouse_y += y;
 }
 
 unsigned int pc98_mouse_rate_hz = 120;


joncampbell123:
The idea is sound but the fix is in the wrong place. The patch in question changes how the emulated bus mouse hardware reports movement to the guest. The proper place to read and clear the counters is in the src/ints/mouse.cpp INT 33h emulation.

joncampbell123:
To remain valid data the bus mouse, just like the PS/2 mouse, has to limit the counter value within the limits of an 8-bit signed integer, which is why those range checks exist.

joncampbell123:
If this is a bootable disk (guest OS), then perhaps the mouse driver is failing to clear the counters through 8255 port C in the way DOSBox-X expects it to (src/hardware/keyboard.cpp line 2419) class PC98_Mouse_8255 instance pc98_mouse_8255 (line 2464) via I/O port 7FD9h-7FDFh odd (line 2467).

joncampbell123:
If the PC-98 bus mouse hardware (real hardware) does not limit the counters to signed 8-bit but allows them to overflow, I'd like to know. Should I pull out some old PC-9821 laptops and test that?

nanshiki:
Macross, but int 33h is not used.
As to why this is the case, there is a compatibility issue between the NEC version of MOUSE.SYS and the Microsoft version of MOUSE.COM.
It is unknown which one is installed by the user, so commercially available games for PC-98x1 do not use int 33h, but instead use a proprietary mouse driver or handle the port internally in the program.
The compatibility is related to #4680, but the biggest problem is that the button states that can be obtained with int 33h ax=3 are different.

The counter value is described in the PC9801 Programmer's Bible.
mouse
The red box part says, "The data width of the mouse counter is 8 bits, so the counter value loops in increments of 256"
In other words, the sign is not supposed to be checked.
I input the sample program that followed and tried to run it on an actual PC-9821, but it seems that the counter is reset after reading the value from 0x7fd9 as the operation.
I have left the program here for you to try.
[test.zip](https://github.com/joncampbell123/dosbox-x/files/13747992/test.zip)
The specifications indicate that the counter is cleared when Bit7 of 0x7fdd is operated, but it seems that the counter is reset even without operating Bit7.
When 8255 was installed in early PC-9801, counter reset was necessary with 0x7fdd Bit7, but when the peripherals were converted to a single chip, it is thought that the counter is reset automatically after reading out the value.
I don't have an early PC-9801 here, so I can't confirm the difference in operation.
Macross was released in 1994, at the end of the PC-98x1 series, so it may be built on the assumption that it will automatically reset after reading the value.

joncampbell123:
I see.

I was wondering why some PC-98 games seemed to have cursor confinement issues.

joncampbell123:
By the way, when they went single chip is that also when PC speaker behavior changed?

There's a known issue where the PC-98 port of Thexder relies on PIT 1 timer behavior of older PC-9801 systems and will hang when your game character dies because of it. It relies on PIT 1 timer cycling at all times and PC speaker output is gated, which is not true on PC-9821 systems where there's no PC speaker output gate and the bit to turn on/off the beep turns on/off the PIT 1 timer counter instead.

joncampbell123:
I think what I'll do is merge your change in and then have it reset on read as you describe unless the machine type is pc9801, and I'll have an option to control it as well. If a game relies on the old behavior, you can set up DOSBox-X to behave like it.

nanshiki:
Sorry. The guess part above is a big mistake.
I was running the mouse driver embedded in the actual machine. The counter is reset in the interrupt of the mouse driver.
If you run test.exe without the mouse driver on the actual machine, the counter will reset to 0 when it exceeds 255 in the positive direction, and to 255 when it is less than 0 in the negative direction.

joncampbell123:
Ah, that does indeed solve the issue I noted with some games where the cursor is confined to some corner of the screen.

nanshiki:
Therefore, I think it is unnecessary to reset the counter by reading the counter, but only by bit7 operation as it is now.
Since the actual machine seems to be operating without a sign, I think the correction in this pull request is sufficient.

joncampbell123:
@nanshiki Would you like to note your change in the CHANGELOG? If you submit a pull request mentioning this change I'll merge it in as well.

nanshiki:
I will send out another pull request and CHANGELOG will be added afterwards.

Also, the "Also newer hardware is said to auto-reset the counter when reading, instead of on bit 7." in the NOTE is a mistake on our part. is a misunderstanding on our part, and in fact the automatic counter resetting does not occur after a readout, so please remove it.

nanshiki:
The counter was being reset because the mouse driver was installed in config.sys, and the counter was reset by changing Bit 7 in the interrupt process.
If the mouse driver is removed from config.sys, it will not be reset unless Bit 7 is changed on the program side.
I was able to confirm that even on the actual PC-9821, if the mouse driver was removed from config.sys, the counter would not be reset unless Bit 7 was changed, so all of the speculation I wrote is incorrect.
In other words, the behavior of the mouse port has not changed with the late model PC-98x1.

PR #4784, opened 2024-01-26
Fix COPY/XCOPY command fails to recognize 2GB+ free disk space

### LINKED ISSUE #4779: Unable to copy files larger than 2gb into a created FAT32 disk image
### Describe the bug

I'm trying to create a 16gb FAT32 formatted HDD, and copy a file that is ~ 2.5gb into that mounted drive, however I get an insufficient disk space error. I start with a clean version of DOSBox-X (v2023.10.6) running on Mac M1 and setting the reported DOS version to 7.1. I have tried with both the default version and the SDL2 version. Despite the disk having enough space, XCOPY will fail to copy the large file and reports and error:

`Insufficient disk space in destination path - C:\PERFSOL\DATA\PS.KDT`

I run the following commands for my particular use case: 

```
IMGMAKE hdd.img -t hd -size 16384 -fat 32
IMGMOUNT C hdd.img
MOUNT D: /Users/cj.moore/dos
XCOPY D: C: /S /E
```

Where within `/Users/cj.moore/dos` there is the 2.5 gb file. I expect any random 2.5 or larger file would fail in the same way. 

To verify on your own, we can create a 2.5 gb dummy file and try to copy it into the image and fail in the same way with the following steps: 

1. Create the target holding drive and dummy file and then move the dummy.img to some place in the host file system where it can be mounted. 
```
IMGMAKE target_hdd.img -t hd -size 16384 -fat 32
IMGMAKE dummy.img -t hd -size 2459 -fat 32
```

2. Mount the target disk image, and host file system and try to copy the file
```
IMGMOUNT C target_hdd.img
MOUNT D /Users/cj.moore/dos
XCOPY D:/dummy.img C: 
```

See that XCOPY fails: 
image

### Expected behavior

I would expect that the large file would be correctly copied to the destination drive given that it has sufficient space. 
image

image

`fdisk` also shows that the C: drive has sufficient space:

image

Mac M1 - Ventura 13.5.1 (22G90)

2023.10.06

### Used configuration

```ini
I use the default config and only chhange the reported DOS version to 7.1.
```

### Output log

_No response_

### Additional information

I have searched existing bugs, and I thought perhaps this issue would have been resolved in: https://github.com/joncampbell123/dosbox-x/issues/1701 

maron2000:
COPY command checks free disk space before actually copying the file, but didn't consider 2GB+ free space in FAT32 images.
This PR fixes it to use a function to consider FAT32.

Fixes #4779 
I personally tested on VS x64 SDL1 build.


maron2000:
Added FAT32 support (detect free disk space) to FreeDOS XCOPY.

PR #4843, opened 2024-02-25
Adding support for ESS ESFM synthesis

Kagamiin:

None.

This PR introduces support for **ESFM** sound synthesis as featured in **ESS's AudioDrive ES1xxx** line of cards (ES1488, ES1688, ES1868 etc.) and supported by several DOS games using the Miles Sound System driver (Warcraft, Theme Hospital among others).

It does so via the [ESFMu emulator](https://github.com/Kagamiin/ESFMu), which we've been developing over the last year from reverse engineering of the card and available documentation on the web. The design of the emulator is based on Nuked OPL3. It's currently fully register-compatible with the real card and the sound output is audibly indistinguishable from the real thing. The emulator has already been battle-tested in other software (specifically, Furnace Tracker) and optimized by members of the community. We're looking for people to reverse-engineer the die scans of the chip so we can improve emulation on the audio side in the future.

ESFM works in two different modes: Emulation mode (OPL3 compatibility, which it boots into) and Native mode (full ESFM feature set). While the card is in Emulation mode, you can use Raw OPL capture just as if you were using any other card. But while the card is in Native mode, it's not possible to use Raw OPL capture, because the address layout of ESFM Native mode is completely incompatible with OPL3. The DRO capture format would have to be altered in order to accommodate for this, and I don't even know if that would even be desirable.

## Additional information

To setup ESFM support, in dosbox-x.conf, you must set both `sbtype=ess1488` and `oplemu=esfmu` (you might also need to set `oplmode=esfm`). If you set the sbtype but not the oplemu, DOSBox-X will fall back to OPL3 emulation. On the other hand, if you set the oplemu but not the sbtype, the card might not work in ESFM-enhanced games since the ESS-specific functions of the DSP interface will not be emulated.

The emulated ESFM card can also be used for playing normal OPL2/OPL3 games. ESFMu does accurately reproduce the differences in sound between a real OPL3 and a real ESFM, so you can appreciate those differences in soundtracks that showcase them.

I also tried to implement support for the ES1488* Windows driver so I could get ESFM MIDI to work under Windows, but I couldn't get the card to be properly recognized. Further debugging of the driver and/or investigation of the real card's behavior will be needed in order to get that to work. So for now I'm submitting this PR which already works for DOS games (auto-detection is still a bit flaky however - you do need to manually select enhanced ESFM mode in the game's audio setup).

_*I chose the ES1488 because it's the closest card with ESFM support to the ES688 which is currently implemented in DOSBox-X._


joncampbell123:
Awesome! Please make sure to note your changes in the CHANGELOG so users know this is an option, however experimental or incomplete!

joncampbell123:
Please make your next pull request a change to CHANGELOG crediting yourself with the addition of ESFM emulation, even if incomplete or experimental in any way.

maron2000:
You need to add the files to dosbox-x.vcxproj to tell Visual Studio to compile the files that you added.
Refer to #4236 #4707 for examples.

Kagamiin:
Yes, I already noticed the issue and I am doing that right now @maron2000. Thanks for the heads up anyway.

PR #4854, opened 2024-02-28
Adding ESS ES1688; removing ESS ES1488 (see description for reason)

Kagamiin:
This PR is adding the ESS AudioDrive ES1688 (`sbtype=ess1688`) sbtype.

This PR is also removing the ESS AudioDrive ES1488 (`sbtype=ess1488`), because as found out by further testing and implementation, it is unsupported by DOS games using the Miles Sound System and does not have ESFM MIDI support in its Windows driver, and therefore I see no use case for the ES1488 in DOSBox apart from custom software that's yet to be made in the future.

None.

This PR adds support for the ESS AudioDrive ES1688 (`ess1688`) sbtype, which unlike the `ess1488` sbtype (as found out by further testing and implementation) is actually supported by DOS games using the Miles Sound System and does have ESFM MIDI support in its Windows driver.

This PR is removing the ESS AudioDrive ES1488 (`ess1488`) sbtype. As I'm performing this removal 2 days after its addition, I'm hoping the impact will be minimal.

If there are any objections against this removal, however, I can simply revert commit d39ec9421, leaving both sbtypes to be used. However, I have updated the ES1488 implementation to better match the real hardware and give it compatibility with the Windows driver, which breaks compatibility with DOS games (the fact it was working before was, ironically, a bug).

## Additional information

The ES1688 contains a toggleable MPU-401 UART in the port range 3x0h-3x1h that I haven't implemented yet. By default, the ES1688 Windows driver tries to enable and test this UART, and if it doesn't work properly, it rejects the card. To get the card to work on Windows, you need to disable the MIDI UART.

In order to get the card to work under Windows: in the Device Manager, select the card, click on the Properties button, go onto the Resources tab, and in the `Setting based on:` spinbox, select `Basic configuration 2`. Check the list of resource settings above - the port range 3x0h-3x1h should have vanished. Adjust the port ranges to match your DOSBox settings and click OK, reboot the machine if needed, and the card should work.

Demonstration video of the ESS AudioDrive ES1688 working under Windows 95:

https://github.com/joncampbell123/dosbox-x/assets/102362203/14443fa0-748a-4640-97e7-9ae0116fd9cf

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -173,9 +173,8 @@ Next:
   - Added support for XDF disk format (maron2000)
   - Added ESFM emulation (ESFMu v1.2 core) (Kagamiin)
   - Added support for ESS "ESFM" sound synthesis (oplemu=esfmu, oplmode=esfm) (Kagamiin)
-  - Added experimental/WIP support for the ESS ES1488 sblaster type (sbtype=ess1488) (Kagamiin)
-  - Delete obsolete information regarding "INTRO SPECIAL" option (maron2000)
   - Added experimental/WIP support for the ESS ES1688 sblaster type (sbtype=ess1688) (Kagamiin)
+  - Delete obsolete information regarding "INTRO SPECIAL" option (maron2000)
    
 2023.10.06
   - Add "VRD" debugger command to force redraw of the VGA screen. (joncampbell123)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -175,6 +175,7 @@ Next:
   - Added support for ESS "ESFM" sound synthesis (oplemu=esfmu, oplmode=esfm) (Kagamiin)
   - Added experimental/WIP support for the ESS ES1488 sblaster type (sbtype=ess1488) (Kagamiin)
   - Delete obsolete information regarding "INTRO SPECIAL" option (maron2000)
+  - Added experimental/WIP support for the ESS ES1688 sblaster type (sbtype=ess1688) (Kagamiin)
    
 2023.10.06
   - Add "VRD" debugger command to force redraw of the VGA screen. (joncampbell123)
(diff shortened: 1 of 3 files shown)


bocke:
I think this is wrong. ES1488 does support ESFM and Windows driver does come with ESFM driver. 

ES488 is the variant without ESFM. ES1488 is the variant with ESFM. 

C
[1488_v1087.zip](https://github.com/user-attachments/files/31397456/1488_v1087.zip)

As in point, from the attached driver:

```inf
[ESS1481.CopyList]
es1488.vxd,,
es1488.drv,,
essfm.drv,,
essmport.drv,,
```

bocke:
While what I wrote is true, if you tested for OPL2 compatibility then you are right: It implements OPL2 via ESFM in software. Windows driver implements OPL2 as compatiblity layer against raw ESFM (undocumented native format). Later cards implement OPL3 and seems like they have hardware OPL3 compatiblity.

But I can't claim that is completely true. I got the information from analyzing the drivers and AI dissasembly, pending more checking.

Anyway, ES1488 is a mono card. While 1688 is a stereo card, so it's even better to support 1688 for better experience. ES1488 emulation is only interesting to people developping for specific retro systems (and using DOSBox-x for testing).

Kagamiin:
@bocke That is actually interesting and I'd like to know more. At the time I was implementing these changes, not much was known about ES1488's ESFM implementation - we knew it had some form of OPL2/3 clone but we didn't have any evidence that it was compatible with later ESFM implementations.

If the card is indeed mono and the drivers are mostly OPL2-compatible, it could be that the ES1488 has some early variant of ESFM that would be interesting to reverse-engineer from a preservation standpoint.

PR #4864, opened 2024-03-02
gh-pages: Update for 2024.03.01

Dobby233Liu:
- Update gems
- devel-build: Generate table in Liquid, and other code cleanup
- Fix \<article> misuse
- DOSBox-X logo is now clickable; adjust other designs
- Pages CMS integration (because it's funny)
- Add 2024.03.01, add special case for missing general Windows setup file
- Update wiki against latest dosbox-x.wiki
- Eilminate wiki footer
- Fake SPA feature for wiki, basically the nav page's URL hash contains the current page now


rderooy:
@Dobby233Liu there are a lot of pending wiki updates already in https://github.com/rderooy/dosbox-x-wiki

joncampbell123:
Ready?

Dobby233Liu:
Gonna test fake SPA a little more

joncampbell123:
> In the past the Windows links were:
> 
> 32/64-bit Setup `ReactOS/XP+` | `Vista+`
> 
> This PR above resulted in: image
> 
> Where the 32/64-bit Setup leads to the XP installer. And on the other line is the 9x/NT4 portable
> 
> So, basically only the "special" versions are shown, while the regular Vista+ portables (or installer, which unfortunately isn't in the release) aren't linked at all?
> 
> @joncampbell123, can you please add the Vista+ installer manually to the release files?

I normally do not make the installers myself? How do I do that?

maron2000:
PR#4869 will hopefully upload the installer for the next coming release. 
github sometimes slightly changes the build environment that prevents building successfully that worked before.

Meanwhile the build identical to the 2024.03.01 release is found here.
https://github.com/joncampbell123/dosbox-x/actions/runs/8118164086

It is built with the code with the same commit hash as the release.

PR #4871, opened 2024-03-05
gh-pages: Fix wiki index.html, and other misc. changes

Dobby233Liu:
See commit messages

Fix wiki/index.html not loading anything if the hash is set to anything else than Home due to incorrect sandbox configuration
image
And more after I realized Guide: is being treated as a protocol for some reason

Even better header wrapping for thin screens: it utilizes flex-wrap, and the tagline is hidden if the screen is tight enough

Two files in the images folder go

## Additional information

~~I may put up a PR to fix another thing related to the wiki later this week~~ Did it right away


Dobby233Liu:
Gosh darn it there's another bug

Issue #4475, opened 2023-09-19
Crash on a loongarch64 CPU (Loongson 3A5000M).


KatyushaScarlet:
### Describe the bug

* Run from the shell
```
LOG: Early LOG Init complete
LOG: DOSBox-X's working directory: /home/katyusha/program/dosbox-x/dosbox-x/src/dosbox-x-dosbox-x-v2023.09.01/src
LOG: Logging init: beginning logging proper. This is the end of the early init logging
LOG: Logging: No logfile was given. All further logging will be discarded.
LOG: DOSBox-X version 2023.09.01 (Linux SDL1)
Warning: XMODIFIERS is unspecified
LOG: Current X11 keyboard layout (token) is: 'us'
LOG: Host keyboard layout is now us (US English)
LOG: Mapper keyboard layout is now us (US English)
LOG: SDL1 auto-detected desktop as 1920 x 1080
LOG: The default output for the video system: surface
LOG: Configured windowposition: -
LOG: X11 main window is 640 x 400 maximized=0
LOG: X11 extension XRANDR is available
LOG: XRandR CRTC 0: pos=(0,0) size=(1920,1080) outputs=1
LOG: Our window lies on this CRTC display (window pos=(0,0) size=(640,400) match=(320,200)).
LOG:   Goes to output 0: name='eDP-1' size_mm=(300 x 170)
LOG: Screen report: Method 'XRandR' (1920.000 x 1080.000 pixels) at (0.000 x 0.000) (300.000 x 170.000 mm) (11.811 x 6.693 in) (162.560 x 161.365 DPI)
LOG: Current X11 keyboard layout (token) is: 'us'
LOG: SDL_mapper: US keyboard detected, set usescancodes=false
LOG: ISA BCLK: 8333333.333Hz (25000000/3)
LOG: monopal: green, 
LOG: Active save slot: 1 [Empty]
LOG: USING AVI+ZMBV
LOG: Max 3670016 sz 16384
LOG: Final 16384
LOG: SDLNet_TCP_Open: Couldn't connect to remote host
LOG: TiMidity: can't open control connection (host=127.0.0.1, port=7777)
LOG: MIDI:fluidsynth: SoundFont not specified, and no system SoundFont found
LOG: MIDI:synth: Specify .SF2 sound font file with midiconfig=
LOG: MIDI:Opened device:none
LOG: Pentium CMPXCHG8B emulation is enabled
[...]

PR #4888, opened 2024-03-10
Disable FP exceptions in a portable way

xry111:
As we are requiring ISO C++11 where `std::feholdexcept` is available for disabling FP exceptions, use it instead of platform-specific hacks.

Prompted by #4475.  It's really not a good idea to expand the `#if` line forever when a portable interface is available.


donmor:
图片
Failed to build with this patch as there are more references to `FPU_SyncCW` in this file.

PR #4906, opened 2024-03-17
PC-98 int 18h ah=47h,48h,49h supported

nanshiki:
PC-98 int 18h function added.
ah=47h Draw line or box
ah=48h Draw arc
ah=49h Draw text


vasco9244:
Thank you!
I verified the operation using PC-98 software called RECOMPOSER.
Until this feature was implemented, staff notation was not displayed.
This feature has been implemented and staff notation can now be displayed.
rc25gs_001

PR #4909, opened 2024-03-18
Add breakpoint type freeze memory 

Enmet:
When reversing games, a practical feature is to be able to freeze a certain value in able to identify it or make certain progress in the game easier. For instance, if there are multiple moving objects on screen, freezing the address of one of them will help reveal which one is which.

This simple PR adds what is essentially a copy of the memory breakpoint, but instead of breaking, it resets the value at the address and thus in practice freezes it in place. Because it is of the breakpoint class, it can be listed and deleted like any other breakpoint type. Only a single byte is frozen at the specified address.

No.

## Additional information
Reasoning behind name and location in help listing: Although it is technically a type of breakpoint, it behaves differently and it is for this reason I believe the shorthand "FM" makes sense. In the help list it is listed below "SMV" (set memory virtual).

### DIFF
--- a/src/debug/debug.cpp
+++ b/src/debug/debug.cpp
@@ -515,7 +515,7 @@ std::vector<CDebugVar*> CDebugVar::varList;
 bool mustCompleteInstruction = false;
 bool skipFirstInstruction = false;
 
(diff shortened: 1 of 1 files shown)


joncampbell123:
Ah, so you can Game Genie yourself some infinite lives, right? :)

Merged.

PR #4961, opened 2024-04-13
Detects possible FAT geometry in fixed VHDs too

### LINKED ISSUE #4958: DBX not mounting Windows created VHDs of 8,16,24,32,40, and 48MB?
### Describe the bug

Title. If I create VHDs with sizes of either 8,16,24,32,40, or 48MB, DBX will not mount the VHD and gives an error of "Cannot create drive from file."

1. Go to Windows Disk Management > Action > Create VHD
2. Create a VHD with a size of 8,16,24,32,40, or 48MB, set to VHD and Fixed Size
3. Initialize the drive as MBR
4. Create a new simple volume, assign a drive letter, and do a quick format
5. Unmount the drive from Windows
6. Go to dosbox-x and type "imgmount [driveletter] path/to/my.vhd -t hdd"

DBX will echo a message of "Cannot create drive from file."

### Expected behavior

The expected behavior is that DBX mounts the VHD to your chosen drive letter.

WIndows 10 Pro 22H2

2024.03.01

### Used configuration

_No response_

### Output log

_No response_

### Additional information

_No response_

maxpat78:
Since MBR and FAT CHS geometries may differ even with Windows 11 created VHDs, we attempt to read both of them.

Fixes #4958

### DIFF
--- a/src/ints/bios_vhd.cpp
+++ b/src/ints/bios_vhd.cpp
@@ -127,13 +127,20 @@ imageDiskVHD::ErrorCodes imageDiskVHD::Open(const char* fileName, const bool rea
         uint8_t buf[512];
         if(fseeko64(file, 0, SEEK_SET)) { fclose(file); return INVALID_DATA; }
         if(fread(&buf, sizeof(uint8_t), 512, file) != 512) { fclose(file); return INVALID_DATA; }
+        //detect real DOS geometry (MBR, BPB or default X-255-63)
         imageDiskVHD::scanMBR(buf, sizes, footer.currentSize);
         vhd->cylinders = sizes[3];
         vhd->heads = sizes[2];
         vhd->sectors = sizes[1];
         vhd->fixedDisk = new imageDisk(file, fileName, vhd->cylinders, vhd->heads, vhd->sectors, 512, true);
+        //now possible FAT geometry can be detected (in Windows 11, those two can differ!)
+        vhd->DetectGeometry(sizes);
+        vhd->cylinders = sizes[3];
+        vhd->heads = sizes[2];
+        vhd->sectors = sizes[1];
+        vhd->fixedDisk = new imageDisk(file, fileName, vhd->cylinders, vhd->heads, vhd->sectors, 512, true);
         *disk = vhd;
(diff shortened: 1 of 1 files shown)


src/ints/bios_vhd.cpp
```diff
@@ -127,13 +127,20 @@ imageDiskVHD::ErrorCodes imageDiskVHD::Open(const char* fileName, const bool rea
...
         vhd->cylinders = sizes[3];
         vhd->heads = sizes[2];
         vhd->sectors = sizes[1];
         vhd->fixedDisk = new imageDisk(file, fileName, vhd->cylinders, vhd->heads, vhd->sectors, 512, true);
+        //now possible FAT geometry can be detected (in Windows 11, those two can differ!)
+        vhd->DetectGeometry(sizes);
+        vhd->cylinders = sizes[3];
+        vhd->heads = sizes[2];
+        vhd->sectors = sizes[1];
+        vhd->fixedDisk = new imageDisk(file, fileName, vhd->cylinders, vhd->heads, vhd->sectors, 512, true);
```
maron2000, line 141: Does calling `new imageDisk()`  twice safely deletes the previous one?
I think better to check sizes[] and DetectGeometry before `new imageDisk`

BridgeHeadland:
Is there really any difference between creating a VHD file in Windows 10 or 11, even in DOSBox-X?

src/ints/bios_vhd.cpp
maxpat78: Obviously, it does not (deleting caused an obscure exception).

DetectGeometry actually requires a valid disk image internally to properly read VHD sectors, that's the reason why I avoided that call in previous commits (having tought at least Windows 1x set MBR CHS == BPB CHS).

Perhaps I'll find another way, if you haven't suggestions.

maxpat78:
> Is there really any difference between creating a VHD file in Windows 10 or 11, even with DOSBox-X?

Yes, if you partition or format, different softwares do different calculations of geometry recorded in MBR and boot sector.
It happens that Windows 11 stores a X-255-63 geometry in boot sector, and another one in MBR.

PR #4973, opened 2024-04-19
Fix reported size issues in builtin DIR command

### LINKED ISSUE #4004: Free space shown is from drive where DOSbox-X executable is, not from the mounted drive
### Describe the bug

- DOS version set to 7.1. freesizecap=false.
- SMB share mounted as drive letter in DOSbox-X. 
- Using DIR command in the DOSbox-X shell.
- free space result is wrong: 156.98GB
image
- correct free space (from Windows CMD.exe) is ~2TB:
image
- the 156.98GB is coming from the drive where `dosbox-x.exe` is located:
image
- if `dosbox-x.exe` is run from the SMB share, then DIR gives another wrong result: 544.14GB
image

mount SMB share as as drive letter in DOSbox-X. Execute DIR.

### Expected behavior

Free space from the correct location to be shown. If there is a limitation due to being "too big for DOS", then it should be one of those?
image

Seeing that DOSbox-X shell shows properly 8GB files (beyond the 4GB max for FAT32), I assume it's possible to show volume sizes beyond 2TB as well. Please consider that it's feasible to have a volume with dozens of TB free space (a few high capacity HDDs in NAS).

Windows 10 22H2

SDL1 MinGW64 2022-11-20

### Used configuration

_No response_

### Output log

_No response_

### Additional information

Not sure if related to #2157

### LINKED ISSUE #4258: Free size limit 2TB not to overflow, 2GB limit to be increased

maxpat78:
- when a SMB share is mounted to a drive letter, the correct free space is shown
- `-freesize` switch is honored
- also, it purges the 1919 MB freesize cap limit

(tested WIN32 version only, with an 8TB VHDX drive)

Closes #4004
Closes #4258

### DIFF
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -33,6 +33,7 @@
 #define DOS_PATHLENGTH 255u
 #define DOS_TEMPSIZE 1024u
 #define DOSERR_FUNCTION_NUMBER_INVALID 1
+
 void DOS_SetError(uint16_t code);
 
 enum {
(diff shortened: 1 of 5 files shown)


maxpat78:
@Torinde The free size is limited in other ways, as you can see:
image

Perhaps I have to remove all that code?

For the SMB share, I created a network share on the same 8 TB VHDX I've mounted for the other test.

maxpat78:
Further notes for possible code reorganization:
dos_programs.cpp: 1170ss (MOUNT -freesize)
The intent is to create a "fake" FAT disk geometry and translate the free space in terms of it (it is expressed in clusters): at this time, no information is known about the real emulated disk!

localDrive::localDrive constructor takes these fake params, passed as uint16_t and uint8_t. Max possible size is 65535 clusters and, we know, DOS knows very little about 4Kn sectors or big (>32K) clusters...

In drives.h, struct allocation (member of localDrive) should perhaps be enhanced to 64-bit, to track modern disk quantities. Appropriate members (AllocationInfo, AllocationInfo32) should translate those sizes as needed by "real" DOS apps.

maxpat78:
@Torinde I made a newer commit https://github.com/maxpat78/dosbox-x/tree/cmd_dir_patch with artifacts to allow `-freesize X` with X>2048.
Can you test them, extensively and intensively?
Since I changed some uint_t* types, I would be certain that no side effects are introduced.

PR #5048, opened 2024-06-13
Add build instructions for Raspberry PI

maron2000:
Add build instructions for Raspberry PI, which is proven to work by the official Raspberry PI website.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -16,7 +16,7 @@ The four major operating systems and platforms of DOSBox-X are:
 
 1. Windows 11, 10, 8, 7, Vista and XP for 32-bit and 64-bit x86/x64 and ARM
 
-2. Linux (with X11) 64-bit x86/x64, and on a Raspberry Pi 3/4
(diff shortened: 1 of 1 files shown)


maron2000:
It's due to security reason that we cannot build it. See the build error.

joncampbell123:
> Why dropping Vista installer support when adding those Pi instructions, how are those related? Or it's just a coincidence in time?
> 
> Is it really needed to drop Vista from the regular installers (I assume that doesn't affect the XP installers)?

I agree. If your pull is about build instructions for the Raspberry Pi, then it should only add build instructions for the Raspberry Pi. Drop the Windows installer commit and I'll merge the Pi instructions.

maron2000:
As per your request, I reverted the commit. Sorry for the confusion.

maron2000:
@Torinde
Let's close this PR by just adding the build instructions for Raspberry PI, as @joncampbell123 requested not to mix several fixes in a single PR.
The website should be updated not now but after the next release.

PR #5059, opened 2024-06-19
Add MEM command for PC-98 mode

maron2000:
Add `MEM` command for PC-98 mode, imported from a forked version 1.11p1 [lpproj/mem.nec](https://github.com/lpproj/mem.nec/).

Tested with VS x64 SDL2 version.


joncampbell123:
Is MEM actually supposed to do something?
command_000

joncampbell123:
Alright. I'm going to open an issue on this because the version check is very specific which is inappropriate for DOSBox-X.

The IBM PC MEM command isn't this picky. The PC-98 version shouldn't be either.

command_001

PR #5090, opened 2024-07-03
Bump SDL2 libary to version 2.30.4

maron2000:
SDL2 library was updated to 2.30.4. https://github.com/libsdl-org/SDL/releases
This PR updates the in-tree SDL2 library.

Tested with VS x86, x64 on Windows 10 Pro 22H2, and MacOS (Ventura).

DOSBox-X custom IME input support working fine.
image


joncampbell123:
I'm for this, even though I recall this might undo one tiny change added by @Wengier some time ago.

maron2000:
Thank you for your comments.
I forgot to reapply PR #5007, so I will do that later on.

PR #5093, opened 2024-07-03
Corrected the version info for the DOSBox-X installers.

JoeBroesele:
Corrected the version info for the DOSBox-X installers (2024.03.1 -> 2024.07.1).


joncampbell123:
I'll merge this and then fix the update-version-number script which does not attempt to update those *.iss files (thus this issue).

PR #5094, opened 2024-07-03
Fix WIndows installer version number

maron2000:
This PR fixes that the version number of Windows installers (Standard and XP) have not been updated for the 2024.07.01 release and after.

Also, some unused files related to Windows installers are removed.


maron2000:
Closed since this is a duplicate of #5093.

PR #5099, opened 2024-07-04
Add "[dos] automount drive directories" option

aybe:

```
[dos]
automount drive directories=true
```

This will auto-mount existing directories, from `DriveC` to `DriveY`.

Somewhat handy but on the other hand it might be troublesome, default value has been set to `false`.

Not sure if it's worth it in the end... feel free to merge... or not.


joncampbell123:
Please add an #ifdef to the code to limit the feature to Windows only (#ifdef WIN32), such a feature is useless on any other OS like Mac OS and Linux.

aybe:
Done.

PR #5104, opened 2024-07-07
libchdr update to latest

treloret:
Updated libchdr and zlib associated with it

chd has added zstd support, for faster decompression with decent compression ratio

Add support for the latest chd's including zstd compression

zlib has been upgraded to new version 1.3.1, this might cause issues with outdated code for zip handling

## Additional information

All assembler based optimizations have been merged out of the lzma handling compared to the main libchdr branche.
Because of my lack of knowledge, handling of chds made with createdvd isn't supported (they use .iso files as a basis), all the required code is there, they use different codecs for that cdlz/cdzl/cdzs/cdfl are for createcd, but for createdvd those are respectively lzma/zlib/zstd/flac


treloret:
Added missing vcxproj files

treloret:
And another missing file in the mac workflow run

PR #5105, opened 2024-07-08
HX DOS build breaks because of reference to _fseeki64/_ftelli64, remove duplicate dr_flac.h

treloret:
Further split the win portion of the #define structure for fseek/ftell, and removed the duplicate dr_flac.h

HX build doesn't work anymore since my additions yesterday.

No

dr_flac.h is now the latest version, according to the changelog there shouldn't be any breaking changes with the previous version

## Additional information

No real means to test if it works, besides letting the HX DOS build run its flow, HX seems to be mingw, but if has mingw64 defined, this error will still occur, if that happens, then I'll have to find another way to identify that the 64-variants of fseek/ftell aren't to be used


treloret:
Silly typo that doesn't break on a linux build, hopefully this fixes all.

maron2000:
Is the duplicated dr_flac.h removed?
I don't see it in the `Files changed` tab.

treloret:
> Is the duplicated dr_flac.h removed? I don't see it in the `Files changed` tab.

Clearly I'm a novice with git, because somehow those changes were reverted, I'll see that it gets in.

treloret:
I tried to keep the changes to libchdr to a minimum, besides the assembly removal and some void* casts, and the breaking fseek/ftell macros, so that any future changes to libchdr should be easier, that includes the reversal of some removed lzmaenc.c functions. Was glorious to see Bleifuss starting up and playable. Zstd deprecated/legacy could probably be removed, and I believe I'm breaking some FOSS code by not including the license files for lzma/zstd, should probably add that.

PR #5107, opened 2024-07-09
Upgrade to VS2022

aybe:

Upgrade project to VS2022.

Also, enabled edit and continue for Windows/Debug builds as it's really a time saver.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -439,3 +439,5 @@ contrib/macos/dosbox-x.plist
 /dosbox-x.conf
 /src/dosbox-x.conf
 
+# visual studio edit and continue temp files
+/vs/enc_temp_folder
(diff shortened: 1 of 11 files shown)


joncampbell123:
Anyone here OK with moving on from VS2019?

maron2000:
We can add a CI workflow that the current code is still buildable on VS2019, just like the C++11 check on Linux builds.
Currently, nightly builds are built on VS2017 and 2019.

maron2000:
Well the only problem I can think of is that if you upgrade to VS2022, VS2019 users can no longer build without reverting the build tool settings, and slight incompatibilities exist between the two tools.

Other than that, I personally feel that we can go ahead to upgrade the build tool.

Nightly builds can flexibly change the build tools.
Already switching between v141(2017) and v142(2019) is done to build XP builds, and switching from v143(2022) will not be a problem.

aybe:
After switching branches, I noticed this:

image

I thought, okay, let's try that instead...

Well... they are making it really hard to get it these days:

image

After searching... winget can get it to install and show up in VS installer:

https://winstall.app/apps/Microsoft.VisualStudio.2019.BuildTools

But it will only work once you do that:

image

So yeah, it's possible to stay on 2019 while using 2022 but it's tedious to do so.

aybe:
Just made a simpler PR for ENC on 2019 instead of that one. #5110

Ideally, 2022 switch would be better, but if not ready then it can be postponed.

PR #5108, opened 2024-07-09
Windows 3.1 style for configuration tool

aybe:

Trying to apply pixel-perfect Windows 3.11 look'n'feel to CFGTOOL...

Buttons are up as well as their focus style.

Ideally, 
 - extract/group/fix theme constants
 - sort messy help pages alphabetically
 - implement mouse wheel
 - simplify some aspects

Before:

image

After:

image

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,6 +21,7 @@ NEXT
   - Updated SDL2 library to 2.30.4 (maron2000)
   - Fixed TTF mode didn't change to graphic mode when machine = Hercules
     (maron2000)
+  - CFGTOOL: more Windows 3.1 styling, general UX/UI improvements (aybe).
 
 2024.07.01
   - Correct Hercules InColor memory emulation. Read and write planar
--- a/src/gui/sdl_gui.cpp
+++ b/src/gui/sdl_gui.cpp
@@ -715,7 +715,7 @@ public:
         sx += 4;
         ex -= 4;
 
-        d.setColor(GUI::Color::Shadow3D);
+        d.setColor(GUI::CurrentTheme.Shadow3D);
         d.drawDotLine(sx,y,ex,y);
     }
 };
--- a/src/gui/sdl_gui.cpp
+++ b/src/gui/sdl_gui.cpp
@@ -3137,7 +3137,6 @@ public:
         bar->addItem(2,mainMenu.get_item("help_about").get_text().c_str());
         bar->addActionHandler(this);
 
-        new GUI::Label(this, 10, 30, MSG_Get("CONFIGURE_GROUP"));
         advopt = new GUI::Checkbox(this, 340, 30, MSG_Get("SHOW_ADVOPT"));
         Section_prop * section=static_cast<Section_prop *>(control->GetSection("dosbox"));
         advopt->setChecked(section->Get_bool("show advanced options"));
--- a/src/gui/sdl_gui.cpp
+++ b/src/gui/sdl_gui.cpp
@@ -1453,8 +1453,22 @@ public:
         b = new GUI::Button(this, button_row_cx + (button_w + button_pad_w), button_row_y, MSG_Get("OK"), button_w);
 
         int i = 0, j = 0;
-        Property *prop;
-        while ((prop = section->Get_prop(i))) {
+        Property *property;
+        std::vector<Property*> properties;
+        auto propertyIndex = 0;
+        
+        while ((property = section->Get_prop(propertyIndex++)))
+        {
+            properties.push_back(property);
+        }
+
+        std::sort(properties.begin(), properties.end(),[](const Property* a, const Property* b)
+        {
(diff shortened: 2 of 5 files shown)


aybe:
By the way, not sure about that one nor when it started to happen:

https://github.com/joncampbell123/dosbox-x/pull/5108/commits/f78010ce9aed89ce806a54c8459a8a852a944505

joncampbell123:
Nice work! Here's an example of an actual Windows 3.1 application using the "3D" common controls DLL for color reference for future development.
krnl386_000

maron2000:
I wonder if we can make the buttons slightly thinner so that the `Save` and `Close` buttons don't get cut off.
The screenshot below is shown when window size is set to default.

image

joncampbell123:
Either the configuration menu should enforce a minimum of 800x600, or perhaps all the buttons could be placed inside a scrollable region.

maron2000:
Seems to set `gridbtnheight = 24` would be fine.

However on Linux, you can use the space used for the menu bar, so you have enough space with the default window size without changing the settings.
So maybe make it conditional for Windows.

image

aybe:
> Nice work! Here's an example of an actual Windows 3.1 application using the "3D" common controls DLL for color reference for future development. krnl386_000

I've got a VM up and been using Paint.NET to grab every detail as I can!

aybe:
> Either the configuration menu should enforce a minimum of 800x600, or perhaps all the buttons could be placed inside a scrollable region.

1st thing I guess would be to remove the header "Choose a settings...", it takes place and is useless.

Then to see where the checkbox could go, maybe it can work in menu or in other windows.

If so, then we gain 1 line immediately.

Also maybe, window can become 720 px wide with maybe slightly smaller labels, e.g. IDE \#1, LPT, COM.

If all these work we might be able to have 5 columns and would gain lot of space for eventual future sections.

aybe:
> I wonder if we can make the buttons slightly thinner so that the `Save` and `Close` buttons don't get cut off. The screenshot below is shown when window size is set to default.
> 
> image



> Seems to set `gridbtnheight = 24` would be fine.
> 
> However on Linux, you can use the space used for the menu bar, so you have enough space with the default window size without changing the settings. So maybe make it conditional for Windows.
> 
> image

Problem is, if below 26 then focus while pressed is cropped... And if you adjust label positions it looks really squished.

Then you have margins between buttons that must be kept else it looks horrible.

Noticed that, not sure why on default config with TTF save/close are not cropped.

maron2000:
Hmm, so it's kind of complicated.
One idea is to remove the 'Choose  a setting...' line, and move the checkbox next to the 'Save' & 'Close' buttons.
Or simply resize the window while diaplaying the configuration tools.

aybe:
Before: 

image

After: 

image

Before: 

image

After: 

image

Summary:
 - made a Win31 pixel perfect button, colors, focus
 - overhauled main window so it fits by default, it's exactly its size now
 - added `Theme` class and upgraded all `Color::` to it
 - adjusted all controls so they are aligned on the right

I think that's enough for the PR...

**Reference for future improvements:**

 - remove useless "Configuration for", "Click help button" at top/bottom of each window
 - implement mouse wheel
 - sort properties by name
 - thoroughly apply win31 theme
   - maybe keep older gray since it's less annoying
 - rename/nice sort main window sections
 - IMFC is empty, could be removed so it's even neater
 - dark theme?
 - ...

aybe:
**Final version:**

image

image

image

Fixed layout, sorted help/properties, revamped main window/labels.

Mouse wheel does not work, it needs window to be clicked at least once for it to work, not sure why:

e5058a800c59de6d2685d073f14449fb8541c632

Dark theme:

Not for now, not easy at all to come up with the right colors...

Full Windows 3.1 theme:

Not for now, I did however extract all themes from CONTROL.INI, will need quite some refactoring.

https://github.com/joncampbell123/dosbox-x/commit/d628570f81a1c09ef9d9f5f7dad0d69912e5021b

---

Beside mouse wheel issue, things look good, I guess it's enough for this PR.

maron2000 (approved):
The test builds are built successfully with the fixes.

joncampbell123:
Ready?

maron2000:
Thank you for the fix. I'm happy that the cropped window is fixed as well as the improved look & feel.

aybe:
Nice!

Just two things I'm not happy with:

This is really a dirty fix I did by just swapping the operands, it happens during ENC obviously so I am wondering whether it's simply due to it or not, searched online but didn't find much about it:

f78010ce9aed89ce806a54c8459a8a852a944505

Seems like the GUI would need good refactoring to implement mouse wheel without clicking it first, correct?

(btw, just noticed that one can flick scroll, obviously, it's been clicked at this point so that doesn't help much)

e5058a800c59de6d2685d073f14449fb8541c632

PR #5110, opened 2024-07-09
Enable edit and continue for all but ARM64 (not available).

aybe:
As an alternative to #5107.

### DIFF
--- a/vs/dosbox-x.vcxproj
+++ b/vs/dosbox-x.vcxproj
@@ -282,18 +282,21 @@
     <IntDir>$(SolutionDir)..\obj\$(ProjectName)\$(Platform)\$(Configuration)\</IntDir>
     <IncludePath>$(IncludePath);$(DXSDK_DIR)Include</IncludePath>
     <LibraryPath>$(LibraryPath);$(DXSDK_DIR)Lib</LibraryPath>
+    <LinkIncremental>true</LinkIncremental>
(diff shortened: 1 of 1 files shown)


maron2000:
All Visual Studio builds are failing.
You may want to take a look.

maron2000:
Maybe the following line you additionally added from the previous PR (#5107) is making the flaw?
```
<LinkIncremental>true</LinkIncremental>
```

vs/dosbox-x.vcxproj
```diff
@@ -726,6 +732,7 @@ for /d %%i in ("$(SolutionDir)\..\contrib\translations\*") do copy "%%i\*.lng" "
       <WarningLevel>Level3</WarningLevel>
       <EnableEnhancedInstructionSet>StreamingSIMDExtensions</EnableEnhancedInstructionSet>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
+      <DebugInformationFormat>EditAndContinue</DebugInformationFormat>
```
maron2000, line 735: Shouldn't use `EditAndContinue` on **Release** builds since it conflicts with
```
<WholeProgramOptimization>true</WholeProgramOptimization>
```
There are couple of more places as well.

aybe: Reverted in #5112.

PR #5123, opened 2024-07-14
CFGTOOL: Windows themes

aybe:
## Additional information

Finally figured out CONTROL.INI nonsense, OLE colors in wrong order! 😫

For what's been upgraded, I simplified drawing code drastically by using fill rects...

I also added a visualizer for `GUI::String`... it took me ages to notice chars are 4 bytes... 💀🤣

image

Border is 1:1, menu almost done, the rest will be very long... (next time)

The original Windows 3.1 themes:

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -22,7 +22,7 @@ NEXT
   - Fixed TTF mode didn't change to graphic mode when machine = Hercules
     (maron2000)
   - Configuration tool: (aybe)
-    - more Windows 3.1 styling: buttons, colors, focus, layout
+    - more Windows 3.1 look'n'feel (themes w/ host dark mode detection)
     - enhance main window layout, it is now visible in its entirety
     - fix layout issues, off by one in rectangle drawing functions
     - fix 'list iterators incompatible' when pressing the Tab key
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -1422,6 +1422,11 @@ void DOSBOX_SetupConfigSections(void) {
     const char* quit_settings[] = { "true", "false", "1", "0", "auto", "autofile", nullptr };
     const char* autofix_settings[] = { "true", "false", "1", "0", "both", "a20fix", "loadfix", "none", nullptr };
     const char* color_themes[] = { "default", "black", "red", "green", "yellow", "blue", "magenta", "cyan", "white", nullptr };
+    const char* color_themes_config[] = {
+        "Windows Default", "Arizona", "Black Leather Jacket", "Bordeaux", "Cinnamon", "Designer", "Emerald City",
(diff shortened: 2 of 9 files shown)


aybe:
Not sure on how to fix these linux/mac errors...

aybe:
The original was based on 16-color mode, had to switch to 24-bit as color panel swatches were so confusing; then I realized that indeed it's different. It shouldn't be too hard to add it back.

Dark mode is hard, although you can easily come up with one, I found it more looking like high-contrast, it's not easy to come up with the right balance of colors.

Last things are on the list but before that, the rest of the interface must be ported to that new scheme first.

aybe:
There you go, all it needs now is a "theme guru" to craft nice ones.

image

image

image

joncampbell123:
Oooh, nice! You got the menu bar to look like Windows 3.1 too! :+1:

aybe:
> Thanks, @aybe!
> 
> Tangentially related - is there a setting to change the DOSbox menu look - graphical (as on these screenshots) vs text-based?

Obviously, LCD Dark looks terrible and one really needs a true dark skin... I suppose I'll try again at some point.

Just noticed a bug in the theme menu which gets overridden by section editor...

The banner you mean? [dosbox[ bannercolortheme

image



> Oooh, nice! You got the menu bar to look like Windows 3.1 too! 👍
> 
> That looks _way_ closer to what Windows 3.1 actually looks like! So now every window looks like WS_DIALOG, which is perfectly fine.

At the moment, borders, title bar, buttons, focus are 1:1, everything else has to be done.

Though nice, these old themes have a big problem, button color is always the same...

While fixing, I will DRY the code first like I did for the buttons, then adjust.

joncampbell123:
One thing you should consider is that the menu bar in Windows 3.1 has a line below it to visually separate it from the client area of the window. Can that be done?

aybe:
Yes I know, I just wanted to get theme in first to see how well it would work with actual code.

aybe:
1. comes from OS?
2. haven't seen it for ages
3. yes, it fixed some cases but broke others
4. rough guess, comes from fixed/sizable Win32 behavior

aybe:
It took me a while to figure out why menu height was -858,993,460... 💀

Windows 3.1 has no 3D window with a menu, I checked everywhere, you'll get stock thickness:

image

joncampbell123:
> It took me a while to figure out why menu height was -858,993,460... 💀
> 
> Windows 3.1 has no 3D window with a menu, I checked everywhere, you'll get stock thickness:
> 
> image
> 

That's how I remember it too.

aybe:
Two things,

First, it is impossible to redo that transparency effect for menu shadow:

image

Second, I don't know where to find these 3D checkboxes, indeed they're familiar but they're not in Windows 3.1:

image

Issue #5132, opened 2024-07-25
Linked VHDs are not working?


MrMEEE:
### Describe the bug

Hi

I simply can't get linked (parent/child) VHDs to work?

See below.. what am I doing wrong??

![Screenshot_20240726_004811](https://github.com/user-attachments/assets/bfebab73-b282-4c21-97f4-712193add042)
![Screenshot_20240726_004922](https://github.com/user-attachments/assets/ae850a41-3a2d-49a7-996f-c65558663db5)

Regards

### Steps to reproduce the behaviour

1. Create VHD file (imgmake vhdtest.vhd -t vhd -size 2000)
2. Mount the file (imgmount c vhdtest.vhd)
3. Verify that the filesystem works (dir c:)
4. Unmount the file (imgmount -u c)
5. Create linked image (vhdmake -l vhdtest.vhd vhdlinktest.vhd)
6. Try to mount the linked image as C:, which failes (imgmount c vhdlinktest.vhd)
7. Try to mount the linked image as primary disk, which fails (imgmount 2 vhdlinktest.vhd)

### Expected behavior

The linked image should be mountable

### What operating system(s) this bug have occurred on?

Ubuntu 24.04

### What version(s) of DOSBox-X have this bug?

DOSBox-X version 2024.07.01 Commit 6f1ab09

### Have you checked that no similar bug report(s) exist?

- [X] I have searched and didn't find any similar bug report.

### Code of Conduct & Contributing Guidelines

- [X] I agree to follow the code of conduct and the contributing guidelines.

maron2000:
I tried following the steps mentioned in PR #4308, but mentions that vhd file can't be opened for linking.
@maxpat78 Am I doing something wrong?

```
echo DYNAMIC
imgmake dyna.vhd -t vhd -size 540 -force
imgmount k dyna.vhd
imgmount -u k

echo its LINK (1st level)
vhdmake -l dyna.vhd dyna_delta.vhd -f
imgmount k dyna_delta.vhd
echo COGITO ERGO SVM >K:\A.TXT
imgmount -u k
```

![image](https://github.com/user-attachments/assets/fbb8ee78-8c96-49dc-b56f-059a4a0c41d4)

maron2000:
I see the error message needs to be corrected.
`dyna_delta.vhd` is not the **parent** VHD image. (or syntax `VHDMAKE [-f] -link parent.vhd new.vhd` is wrong?)

MrMEEE:
[...]

PR #5143, opened 2024-07-30
Changed Linux build to use 16-bit wide chars, to fix issue #5132

MrMEEE:

#5132

no

no

## Additional information
VHD Differencing is broken on Linux systems, this fixes the issue, by using the same size chars as on Windows and Macs

### DIFF
--- a/build
+++ b/build
@@ -9,6 +9,11 @@ if [ "$1" == "32" ]; then
     opt="--host=i686-pc-linux-gnu --target=i686-pc-linux-gnu"
 fi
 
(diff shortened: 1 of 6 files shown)


joncampbell123:
Eh.... if the issue is the VHD image code using wchar_t, forcing the whole project to use 16-bit wchar is kind of a sledgehammer approach that might cause other problems. Other libraries in Linux that use the normal wchar_t, for example.

I think a better approach is to modify the VHD image code to change wchar_t to uint16_t so the strings line up properly.

PR #5153, opened 2024-08-04
Calculates true parent's relative path in Differencing VHD locator

### LINKED ISSUE #5145: Linked VHDs in subfolders doesn't work
### Describe the bug

When trying to mount a linked VHD, located in a relativ path (subfolder), mounting fails..

Prior to testing this, https://github.com/joncampbell123/dosbox-x/pull/5143 must be applied to get linked VHDs to work on Linux

1. Create a subfolder outside of Dosbox (mkdir test)
2. Create a VHD in current dir or a subfolder (imgmake test.vhd -t vhd -size 100)
3. Create a VHD in the subfolder which is linked (vhdmake -l test.vhd test/subfolder_test.vhd)
4. Try to mount the VHD (imgmount test/subfolder_test.vhd)

### Expected behavior

VHD will be mounted

Ubuntu 24.04

2024.07.01

### Used configuration

_No response_

### Output log

```text
See screenshot
```

### Additional information

_No response_

maxpat78:
This patch is tested both in Windows 11 and Linux (Ubuntu, WSL). It refines the Differencing VHD handler in following ways:

- always tries to handle an absolute W2ku locator;
- more log messages to debug;
- 2nd locator offset is relative to 1st, a fixed position is no more assumed;
- calculates and stores the true parent's relative path in W2ru locator (parent's path has to be relative to child's path, not current working directory).

As a consequence, VHD sets created in Windows can be shared with Linux users and vice-versa (due to correct relative locator).

NOTE FOR THE WILLING PROGRAMMER: a `-relink` switch could be added to `VHDMAKE` allowing to move a parent VHD and manually relink its childs.

Closes #5145

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -139,8 +139,3 @@ contrib/macos/dosbox-x.plist
 # ignore default dosbox config files (used for manual testing)
 /dosbox-x.conf
 /src/dosbox-x.conf
-
-src/dos/dos_programs.7z
-src/lnkvhd.cpp
-src/mkvhd.cpp
-src/mkvhd.exe
--- a/.gitignore
+++ b/.gitignore
@@ -140,3 +140,7 @@ contrib/macos/dosbox-x.plist
 /dosbox-x.conf
 /src/dosbox-x.conf
 
+src/dos/dos_programs.7z
+src/lnkvhd.cpp
+src/mkvhd.cpp
+src/mkvhd.exe
--- a/src/ints/bios_vhd.cpp
+++ b/src/ints/bios_vhd.cpp
@@ -49,6 +49,51 @@
 *
 */
 
+// returns the path of "base" in a form relative to "child", where both exist
+char* calc_relative_path(const char* base, const char* child) {
+#ifndef WIN32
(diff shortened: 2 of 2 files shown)


maxpat78:
Test for Linux user:
- prepare a `subdir` and a `subdir/subsubdir` testing dirs in working directory;
- run following batch file in DosBox-X shell:
```
imgmake subdir/parent.vhd -t vhd -size 100 -force
vhdmake -f -l subdir/parent.vhd subdir/subsubdir/child.vhd 
imgmount d subdir/subsubdir/child.vhd
imgmount -u d
```

.gitignore
```diff
@@ -140,3 +140,7 @@ contrib/macos/dosbox-x.plist
 /dosbox-x.conf
 /src/dosbox-x.conf
 
+src/dos/dos_programs.7z
```
maron2000, line 143: Do we need to ignore these files?
There are no such file in the current code tree.

maxpat78: No, you don't. In fact those files came by mistake from another code tree of mine, but they hadn't to.

PR #5163, opened 2024-08-08
Added sharp-bilinear Direct3D shader 

RibShark:
This adds a sharp-bilinear shader for Direct3D.

None that I can see.

Yes, a new Direct3D shader (roughly equivalent to the "sharp" OpenGL shader).

No.

## Additional information

Requires #5161.

### DIFF
--- a/contrib/windows/shaders/sharp-bilinear-simple.fx
+++ b/contrib/windows/shaders/sharp-bilinear-simple.fx
@@ -1,0 +1,93 @@
+/*
+   Author: RibShark, rsn8887, (based on TheMaister)
+   License: Public domain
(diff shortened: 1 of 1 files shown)


BridgeHeadland:
As someone who likes sharp images, what is the difference between Direct 3D and OpenGL output like that otherwise?

RibShark:
> As someone who likes sharp images, what is the difference between Direct 3D and OpenGL output like that otherwise?

it should be mostly identical, but OpenGL has a bug (#3073) that prevents aspect correction when shaders are used, so I wanted to get this shader in D3D which isn't affected.

PR #5209, opened 2024-09-13
Fix codepage and language file switching issues

maron2000:
## What issue(s) does this PR address?
Some bugs in changing codepages were fixed to address the following two issues.
Keyboard layout is now unchanged while switching codepages.
Codepage is changed appropriately when loading language files using `CONFIG` command so that menu bar is not glitched.

Tested on VS and MinGW builds on Windows 10 Pro 22H2.

Fixes #5207
Fixes #5208 

Keyboard layout unchanged when switching codepages.
image

No menubar glitches when loading language files using `CONFIG` command. (normal and TTF output)
image

image

PR #5216, opened 2024-09-18
Fix codepage when loading a language file at launch

maron2000:
This PR fixes a bug in PR #5209 that codepage was wrong when loading a language file at launch.
Used command `dosbox-x -set language=ko_kr -set output=default`

Before fix
image

After fix
image

image

image


maron2000:
Some problems arised. Pending this PR.

PR #5231, opened 2024-10-02
Switch macOS x86 CI runner image to macos-13

maron2000:
macOS x86 CI runner is currently `macos-12` which is planned to retire on Dec.3rd 2024.
This PR switches to `macos-13`, the latest image for x86 macOS.

https://github.blog/changelog/2024-08-19-notice-of-upcoming-deprecations-and-breaking-changes-in-github-actions-runners/


maron2000:
I don't know github will ever supply a Intel macos-14 runner.
Sonoma and Sequoia are only available for Arm Macs only.

https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

PR #5263, opened 2024-10-30
Change order of menubar to cope with inaccessible items due to overlapped menus

### LINKED ISSUE #5249: SDL menu usability issues because menu size and scaling
### Describe the bug

Hello.

I noted issues with the SDL built-in menu to reach certain menus and items when the relative size of the menu and the DOSBox-X window size are very close. It's the same issue noted here #2877 in reality, but It could be trigger in bigger window sizes. 

[As explained here](https://github.com/joncampbell123/dosbox-x/issues/877#issuecomment-418227155), the menu first doubles the size to maintain readability in bigger resolution size, and that's fine. The jump happens in 1280 pixels horizontal (1280x800).

When the menu is double, and for example, the DOSBox-X window size is 1280x960 (bigger rounded 4:3 resolution to fit in a 1080p display) some items are inaccessible because the menu need to be rendered inside the window itself, as opposite when DOSBox-X on Windows is compiled to use its menus (There I have a 1280x960 DOSBox-X window and there is no issues with the menus).

To show the difference, there is the DOSBox-X with window size 1279x960:

Main Window at 1279x960 (tiny menu):
DOSBox-X_1279x960_Main

Config tool at 1279x960 (tiny window, bit still visible):
DOSBox-X_1279x960_ConfigTool

Drive menu at 1279x960 (tiny menu, but all options are reachable):
DOSBox-X_1279x960_DriveMenu

One pixel more, 1280x960, the menu doubles the size:

Main Window at 1280x960 (big menu):
DOSBox-X_1280x960_Main

Config tool at 1280x960 (big window, good visibility):
DOSBox-X_1280x960_ConfigTool

Drive menu at 1280x960 (big menu, but all options are **NOT** reachable. You can't select options from other drives than _A_!):
DOSBox-X_1280x960_DriveMenu

In video, to see it in action:
[DOSBox-X_1280x960_DriveHelpMenuIssue.webm](https://github.com/user-attachments/assets/54f596b1-178f-4266-abc7-a2467feb948e)

In #2877, the issue was using the _original_ window resolution (640x480) but without doubling the menu size, the unreachable menus are the same. 

1. Use DOSBox-X with built-in SDL menus (Flatpak Linux uses them).
2. Set window resolution to 1280x960 (SDL section, `windowresolution` in _dosbox-x.conf_).
3. Use It in window mode.
4. Try to reach all letters in the Drive menu or other ones with sub-menus. 

### Expected behavior

All items should be reachable.

KUbuntu 24.10, KDE Plasma 6.1.5, 1920x1080 (125% Scale)

DOSBox-X 2024.10.01 (Linux SDL2 64 bit). Flatpak version. Language es_ES.

### Used configuration

```ini
Only setting changed from default is:

[sdl]
(...)
windowresolution  = 1280x960
(...)
```

### Output log

_No response_

### Additional information

I know that have a bigger menu size is needed on bigger resolutions, but the jump at 1280 is a big one for using at that resolution or a little up. This was more or less talked here #877.

The problem here is not from the resolution itself, is how those menus work, as they overlap over the previous ones.

Some short or long solutions I could imagine could be:
- Have a between 1x - 2x menu size to use at a lower resolution (1.5x, but using the same menu with that multiplier for sure it gonna be look ugly).
- Redesign the menus, or at least, how they overlap between them, or how the sub-menus appear (click to appear sub-menu setting?).
- Set an option to force the menu scaling, setting _auto_ as default like it works, setting 1, 2, 3... as the multiply factor.

Thanks in advance!

maron2000:
Some items in the SDL drawn menu were not accessible due to the overlapping of items.
This PR swaps some orders of the item to make it accessible.

1. "Drive" menu and "Capture" menu are swapped
2.  Move "Save/Load options"
3.  Swap "DOS command"reference and "About DOSBox-X"

Fixes #5249

image
image
image

Useful for Japanese users as well.
image

### DIFF
--- a/src/gui/menu.cpp
+++ b/src/gui/menu.cpp
@@ -124,10 +124,10 @@ static const char *def_menu__toplevel[] =
     "VideoMenu",
     "SoundMenu",
     "DOSMenu",
+    "DriveMenu",
 #if !defined(C_EMSCRIPTEN)
     "CaptureMenu",
 #endif
-    "DriveMenu",
 #if C_DEBUG
     "DebugMenu",
 #endif
@@ -709,9 +709,9 @@ static const char *def_menu_capture[] =
     "mapper_capnetrf",
     "--",
 #endif
-    "saveoptionmenu",
     "mapper_savestate",
     "mapper_loadstate",
+    "saveoptionmenu",
     "saveslotmenu",
     "autosavecfg",
     "browsesavefile",
@@ -860,7 +860,7 @@ static const char* def_menu_help_debug[] =
 static const char *def_menu_help[] =
 {
     "help_intro",
-    "HelpCommandMenu",
+    "help_about",
 #if !defined(HX_DOS)
     "--",
     "help_homepage",
@@ -880,7 +880,7 @@ static const char *def_menu_help[] =
 #if C_PCAP || C_PRINTER && defined(WIN32) || !C_DEBUG && !defined(MACOSX) && !defined(LINUX) && !defined(HX_DOS) && !defined(C_EMSCRIPTEN)
     "--",
 #endif
-    "help_about",
+    "HelpCommandMenu",
     NULL
 };
 


maron2000:
You can also consider revising the translation to be a bit shorter.
I made the Japanese translation, consider overlapping and some times the translation may not be precise but to an extent that it is still understandable.

PR #5335, opened 2024-12-23
Don't switch to CP950 if zh_tw.lng is loaded when CP is 951

maron2000:
Language files for CP 950/951 (Traditional Chinese) can be shared between each other, but CHCP command sometimes didn't accept it.
This PR fixes such issue.

Fixes #5327 (PR tested in ticket)

PR #5339, opened 2024-12-23
Fix regression of garbled menubar in Traditional Chinese menu

maron2000:
PR #5335 had an regression that some items in the menubar may become garbled in a certain condition when Traditional Chinese language file was chosen.
This PR fixes such bug.

Steps to reproduce:
Launch DOSBox-X by the following command.
```
dosbox-x -set country=81,951 -set uao=true -set keyboardlayout=us -set language=zh_tw
```

Before fix
image

After fix
image


1abcd:
This is better than PR 5335. Thanks.

PR #5388, opened 2025-01-09
Fix Linux CI build

maron2000:
Github and/or Ubuntu dropped libpng as pre-installed library resulting in error of Linux CI builds. 
Add libpng before the build to avoid error.


joncampbell123:
You mean they dropped the libpng headers from install by default, right?

PNG is common enough in desktops that they would be nuts not to include the shared library.

maron2000:
Well, I don't know if only the header is missing or everything is missing.
But rediculously we have to apt-get libpng.

PR #5448, opened 2025-02-01
Revert commit 706fd0e to fix Windows build errors

maron2000:
Commit 706fd0e resulted in errors for all Windows related builds, so revert the deleted files.


joncampbell123:
I see. They didn't do anything on Linux. Sorry about that.

PR #5528, opened 2025-02-28
Fix glitch of DBCS character at the end of a line

maron2000:
When printing out messages in translation files, the current code is supposed to set the first byte of a DBCS character not to be located at the end of line, but by default the feature is disabled.
This PR fixes so that the feature is enabled by default.

image

### DIFF
--- a/include/programs.h
+++ b/include/programs.h
@@ -106,7 +106,7 @@ public:
 	bool EraseEnv(void);
 	virtual void WriteOut(const char *format, const char * arguments);
 	void WriteOut(const char * format,...);				//! Write to standard output 
-	virtual int WriteOut_NoParsing(const char * format, bool dbcs = false); //! Write to standard output, no parsing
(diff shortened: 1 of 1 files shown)


1abcd:
Hi @maron2000 

I found a similar problem when I typed or pasted these texts.

a十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅
a十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅十卅
a〸〺〸〺〸〺〸〺〸〺〸〺〸〺〸〺〸〺〸〺〸〺〸〺〸〺〸〺〸〺〸〺〸〺

maron2000:
Thank you and yes, I was aware of that as well, but haven't figured out how to fix it.

1abcd:
未命名


https://github.com/joncampbell123/dosbox-x/commits/master/?since=2021-12-11&until=2021-12-12&author=Wengier
There are some changes that are relative to the DBCS problem.
But I can not understand that. That might be helpful.

PR #5532, opened 2025-03-03
Add more missing translations 

maron2000:
Add missing translations for various commands.
With this PR, messages of DOS commands in dos_programs.cpp should be able to be translated.


1abcd:
Hi @maron2000 
This message seems not used.
`MSG_Add("PROGRAM_IMGMOUNT_GEOMETRY_ERROR", "Unable to detect geometry\n");`

PR #5543, opened 2025-03-07
[WIP] A more universal fix for glitch of DBCS character at the end of a line

maron2000:
PR #5528 fixed the glitch of DBCS character at the end of a line for only limited cases.
This PR patches the CON output for a more universal fix.

Test on Traditional Chinese
image

Test on Japanese
image

PR #5574, opened 2025-03-20
More fixes for glitch of DBCS character at the end of a line

maron2000:
Some additional fixes since PR #5543, mostly regarding editing command history.
Fixes include display corruption in command line when pressing up, down, left, right, delete, and backspace keys.

Examples of corruptions can be found in the discussions in PR #5543.

Edit: Additionally fixed Home and End keys.

### DIFF
--- a/src/dos/dev_con.h
+++ b/src/dos/dev_con.h
@@ -1050,8 +1050,9 @@ std::string log_dev_con_str;
 bool logging_con = false;
 bool DOS_BreakTest(bool print);
 void DOS_BreakAction();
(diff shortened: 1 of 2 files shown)


1abcd:
It seems not stable when I insert some characters.

https://github.com/user-attachments/assets/54f1f36d-8b90-41c3-bdfc-a4f0af3a3e3c

Another different problem can be found in old versions of dosbox-x.

https://github.com/user-attachments/assets/2f12f689-24ff-4852-9500-52ea52a33145

PR #5576, opened 2025-03-20
Fix creation of overlay special files

### LINKED ISSUE #4354: Crash due to .DBOVERLAY_DEL_ invalid file name concatenation
### Describe the bug

When trying to create the `.DBOVERLAY_DEL_` file, the original and overlay paths are erroneously concatenated together. This results in an invalid file path which then results in a crash.

Linux:
```
$ mkdir a
$ mkdir b
$ touch a/FILE.TXT
```

DOSBox-X:
```
Z:\> MOUNT C a
Z:\> MOUNT C b -t overlay
Z:\> C:
C:\> del FILE.TXT
```

### Expected behavior

`.DBOVERLAY_DEL_FILE.TXT` is created in the proper location on the host resulting in the file being 'deleted' in the guest.

Ubuntu 22.04 (KDE Neon)

2023.05.01, commit e62dd921669e8321b190e9e5ffcc3444322c1e64

### Used configuration

_No response_

### Output log

```text
LOG: E_Exit: Failed creation of b/a/.DBOVERLAY_DEL_FILE.TXT
E_Exit: Failed creation of b/a/.DBOVERLAY_DEL_FILE.TXT
```

### Additional information

_No response_

zyebytevt:
Closes #4354

Might be a naive fix, but changing the variable from `tname` to `name` seems to fix the crashing issue. Tested it for a few hours, including with Windows 3.1, and everything seems fine.

### DIFF
--- a/src/dos/drive_overlay.cpp
+++ b/src/dos/drive_overlay.cpp
@@ -1499,7 +1499,7 @@ void Overlay_Drive::add_deleted_file(const char* name,bool create_on_disk) {
 		strcat(tname,temp_name);
 	if (!is_deleted_file(tname)) {
 		deleted_files_in_base.emplace_back(tname);
-		if (create_on_disk) add_special_file_to_disk(tname, "DEL");
+		if (create_on_disk) add_special_file_to_disk(name, "DEL");
 	}
 }
 


joncampbell123:
Before I merge this, can you please figure out why the original code crashes and why your change fixes it?

It's not clear to me what the difference is.

zyebytevt:
From what I could figure out, `add_special_file_to_disk` takes the DOS name as it's first argument, i.e. the file path inside the virtual file system. The other two invocations for the function (`RMD` and `ATR`) do pass the correct path, the `DEL` invocation however passed `tname` instead of `name`, which is the already concatenated host path to the overlay directory.

This caused the old code to concatenate the path twice, resulting in an invalid path and crashing. Passing just `name` seems to fix it, as that variable only holds the DOS name.

joncampbell123:
> From what I could figure out, `add_special_file_to_disk` takes the DOS name as it's first argument, i.e. the file path inside the virtual file system. The other two invocations for the function (`RMD` and `ATR`) do pass the correct path, the `DEL` invocation however passed `tname` instead of `name`, which is the already concatenated host path to the overlay directory.
> 
> This caused the old code to concatenate the path twice, resulting in an invalid path and crashing. Passing just `name` seems to fix it, as that variable only holds the DOS name.

Ah, got it.

PR #5590, opened 2025-03-24
Initial port to OS/2

josch1710:
Added things to get an OS/2 port running. TBD is support of Tiny File Dialogs.


joncampbell123:
Awesome!

I don't have the ability to compile for OS/2 (I assume Warp?) and I won't be able to test my code on it nor ensure that changes I make won't break the OS/2 version, nor can I answer any questions about DOSBox-X on OS/2.

it would be a good idea to check in from time to time if anyone has questions about the OS/2 version of DOSBox-X. We have a Discord server as well.

joncampbell123:
Your patch breaks the "lowend" builds for XP and lower because those do not have inet_pton and inet_ntop, those didn't appear until Vista.

https://github.com/joncampbell123/dosbox-x/actions/runs/14039223423/job/39307845650?pr=5590

joncampbell123:
The second issue I have is that your patch is really large.

If you're going to update OS/2 support can you please split the patch into smaller more manageable pull requests? It's OK to do your work gradually.

PR #5595, opened 2025-03-25
Fix insertion of characters in command history

maron2000:
As reported in the conversation of PR #5574, insertion and overwriting of characters in a DBCS string was not working as intended.
This PR fixes the insertion/overwriting of characters.

Before fix
image

After fix
image


1abcd:
Cursor movement is not consistent, if set autodbcs=false.

https://github.com/user-attachments/assets/4f537429-d45e-4735-9ebc-898f82a34f1e

maron2000:
This PR has nothing to do with the behavior of cursor keys, so remove your comment in this PR.
Also, I didn't change the difference of the behavior of cursor keys between `autodbcs=true` and `autodbcs=false` in my recent fixes.
The code did not, and will not consider DBCS characters at all when set to false.

PR #5598, opened 2025-03-26
Renamed variable ulimit to mlimit (bios.cpp) and class COLOR to COLORPGM (dos_programs.cpp)

josch1710:
Renamed variable ulimit to mlimit (bios.cpp) and class COLOR to COLORPGM (dos_programs.cpp), because both symbols are always defined in OS/2 EMX.

To be able to port DOSBOX-X to OS/2, I need to solve the conflict with these pre-existing symbols.
Should be other names be more suitable, please say so.

Should not break anything.

### DIFF
--- a/build-os2.cmd
+++ b/build-os2.cmd
@@ -1,0 +1,4 @@
+echo off
+bash autogen.sh
+dash configure
+make -j4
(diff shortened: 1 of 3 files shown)


josch1710:
Well, I used CLion on Windows to create the PR, so it's possible that some Linux line endings where converted to Windows ones. I wondered, why Github was showing diffs, where CLion showed none.

maron2000:
Some git clients have settings to ignore differences if it concerns white spaces only.
Usually you can switch it on or off.

PR #5599, opened 2025-03-26
Int21h AH=0x0a Keep cursor position if keyboard buffer full

maron2000:
Int21h AH=0x0a function (Buffered input) beeps when buffer is full.
This PR fixes so that cursor position is unchanged when it beeped.

According to PC-DOS 7.0, echoed beep can be redirected, but of course not included in the output buffer.

### DIFF
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -1293,7 +1293,13 @@ static Bitu DOS_21Handler(void) {
                     }
                     if (read == free && c != 13) {      // Keyboard buffer full
                         uint8_t bell = 7;
+                        uint8_t page = real_readb(BIOSMEM_SEG, BIOSMEM_CURRENT_PAGE);
+                        uint8_t col = CURSOR_POS_COL(page);
+                        uint8_t row = CURSOR_POS_ROW(page);
+                        BIOS_NCOLS;
                         DOS_WriteFile(STDOUT, &bell, &n);
+                        if(CURSOR_POS_COL(page) > col)
+                            INT10_SetCursorPos(row, col, page); // stay where we were
                         continue;
                     }
                     DOS_WriteFile(STDOUT,&c,&n);


maron2000:
Try your program on a real DOS, and show us comparison the actual result.
I tried a simple program my self and the result was identical.

maroonbells:
Unless there's a way to install 'real' DOS inside DOSBOX-X

Here's another A86 file that creates a better INPUT.BIN to use as redirected input, which zeroes in on the special characters at the front of the ASCII alphabet. It creates a 255+8=263 byte file containing 8 'Z' preceding the terminating 0x0D. The number of 'Z' in the buffer matches the number of bytes stripped from the DS:DX buffer. In the main release of DOSBOX-X I am seeing 3 'Z' because these 3 bytes are missing from the input buffer, and if that's the way 'real' DOS works, then that's great:

2 = The 0x08 byte is handled as if the ^H destructive backspace were pressed, resulting in the 0x68 and 0x08 both stripped from the DS:DX buffer.
1 = The 0x0A byte is stripped from the DS:DX buffer.

Also, can you verify that the 2nd byte 0x00 is being handled the same way as by 'real' DOS?
thanks

radix 16
mov	dx,offset filename
sub	cx,cx
mov	ah,3c
int	21h
jc	bad
xchg	bx,ax
mov	di,dx
mov	cl,0ff
mov	al,0
L1:
mov	ah,al
and	ah,0f
add	ah,60h
mov	[di],ah
inc	di
stosb
L2:
inc	ax
cmp	al,0d
jz	L2
cmp	al,20h
jae	L3
loop	L1
mov	al,40h
L3:
stosb
inc	al
and	al,4f
or	al,40h
cmp	di,offset filename + 0fdh
jbe	L3
mov	ax,"ZZ"
mov	cl,4
rep	stosw
mov	al,0d
stosb
mov	cx,di
sub	cx,dx
mov	ah,40
int	21h
bad:
int	20h

filename:db "input.bin",0

PR #5607, opened 2025-03-29
Some changes for porting OS/2

josch1710:
Some changes for porting OS/2. 
There are some more changes to be commited, which would potentially break the XP build, so they will PR separately.

Building an OS/2 executable


joncampbell123:
I'm going to merge your change and then make the quick change to the help text so it's still shown as COLOR not COLORPGM

PR #5614, opened 2025-04-01
More OS/2 porting

josch1710:
Porting to OS/2. More changes. DOSBOX-X should now build against SDL 1 and be usable.
SDL 2 and Tiny File Dialogs support is coming.

Potentially, it breaks the XP build. This time around, I changed the #ifdef's.

### DIFF
--- a/README.video-debug-overlay.md
+++ b/README.video-debug-overlay.md
@@ -5,7 +5,7 @@
 The video debug overlay is an extension to the video output of DOSBox-X that puts VGA state information on the bottom and right sides of the DOS screen. The information updates constantly so that state changes are immediately visible.
 
 ## Why do the scalers and video mode affect it?
-The video debug overlay is rendered as part of the DOS screen so that, when enabled, it appears in your screen captures and video recordings as well, making it easy to provide screenshots for video debugging purposes, and a video recording where it is possible to analyze the information frame-by-frame in a video editor.
+The video debug overlay is rendered as part of the DOS screen so that, when enabled, it appears in your screen captures and video recordings as well. This makes it easy to provide screenshots for video debugging purposes, and to analyze the information frame-by-frame in a video player or editor.
 
 ## What is provided in the video debug overlay?
 The exact information provided depends entirely on what video hardware is being emulated. Even when registers and state are common across hardware, exact state varies from video type to video type.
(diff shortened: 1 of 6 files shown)


josch1710:
Heads up: I'm trying to compile the lowend build, to find a way to get it building.

josch1710:
So, I can't replicate the issue. What would be your advice on testing build the lowend versions?

joncampbell123:
> So, I can't replicate the issue. What would be your advice on testing build the lowend versions?

You'll need Windows XP and the original MinGW project. The newer MinGW project cannot target anything below Windows Vista.

The error in question is the use of inet_pton and inet_ntop. Those functions didn't exist in the WINSOCK library until Windows Vista.

maron2000:
As clearly mentioned in the build instructions, you need to setup a Windows XP / HX-DOS compatible environment.

The build environment itself in bundled in the source code tree  if you don't have one.
As an example, Release and nightly builds are built by replacing the latest environment with the XP / HX-DOS compatible one.
You can refer to the CI build procedure to see what I mean.
https://github.com/joncampbell123/dosbox-x/blob/master/.github/workflows/vsbuild_xp.yml

josch1710:
> > So, I can't replicate the issue. What would be your advice on testing build the lowend versions?
> 
> You'll need Windows XP and the original MinGW project. The newer MinGW project cannot target anything below Windows Vista.
> 
> The error in question is the use of inet_pton and inet_ntop. Those functions didn't exist in the WINSOCK library until Windows Vista.
Yes, I understood that, but that part is guarded by # !defined(__MINGW32__) in misc_util, isn't it? On OS/2, there is no enet.h, either. That's why expanded this guards, which clearly makes the build break.
Well, I will fetch an old nettop, I have lying around, and install XP and the MingW. Is this the one, that gets downloaded in the action workflow file?

maron2000:
It isn't a must to install the old MinGW on a XP machine.
In fact, the Releases and the nightly builds are built on Windows Server 2022 with the old MinGW installed.

josch1710:
> It isn't a must to install the old MinGW on a XP machine. In fact, the Releases and the nightly builds are built on Windows Server 2022 with the old MinGW installed.
> 
> Edit: Just in case the link to the MinGW is as below. https://github.com/joncampbell123/dosbox-x/blob/master/build-scripts/mingw/lowend-bin/i686-7.3.0-release-posix-dwarf-rt_v5-rev0%2Bnasm.7z

Yes, I tried that, but it didn't get my a build error.
To verify: I installed MSYS2, downloaded the old MingW package, unpacked it, and put it in the path. then I ran the script. This should be the correct way to do the build?

maron2000:
If it runs on XP, then it would be correct.
If so, you can modify the CI build script to follow your steps so that we can build the Release and nightly as well.

josch1710:
I reverted the breaking changes in misc_util. The checks are now working.
Perhaps, the sensible thing to do, would be to test `inet_ntop `and `inet_pton `in configure, and then guard against something like
`#if def HAS_INET_NTOP` or whatever.

PR #5636, opened 2025-04-14
OS/2: SDL 2 builds and run

josch1710:
- SDL2 build now works
- Debug build now works
- Added build scripts for OS/2

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -576,6 +576,10 @@ case "$host" in
        AC_DEFINE(OS2, 1, [Compiling on OS/2 EMX])
        dnl FEATURE: Whether to support direct serial port pass-through
        AC_DEFINE(C_DIRECTSERIAL, 1, [ Define to 1 if you want serial pass-through support (Win32, Posix and OS/2).])
+       dnl If we have SDL2, we need to link the multimedia extensions.
+       if test "x$SDL2_LIBS" != "x"; then
+         LIBS="$LIBS -lmmpm2"
+       fi
        ;;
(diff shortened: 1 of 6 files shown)


josch1710:
> In order to prevent bitrot of the OS/2 changes - can you please create build instructions and/or even better - Github Action [.yml file](https://github.com/joncampbell123/dosbox-x/pull/5590#issuecomment-2750455751)?
Build instructions are a TODO, but I'm not finished, so they are on a back burner. I hve to look through the files, whether I wrote something, when I did the original port of DOSBOX proper to OS/2. 
But actions, as much as I would like that, are not possible with a good cross compiler. Also, I have no intentions to make it run with OpenWatcom.
 
> What are the minimum requirements for the OS/2 binary to run? I assume you didn't change CPU, FPU and RAM requirements, but what about:
> 
>     * OS/2 version?
> 
>     * video, sound, mouse - anything compatible with the OS/2 version is OK or more strict requirements?
see above 
 
> Does the OS/2 build have the DOSbox debugger?
> 
> > Added build scripts for OS/2
> 
> I assume [here](https://github.com/joncampbell123/dosbox-x/pull/5646)
I didn't really tested it, yet, but I have build script with debug enabled and they are building. I remember from the original port of DOSBOX to OS/2, that it was working.
 
> Are you're interested in adapting DOSbox-X with [SDL 1.2 for Win3.1](https://github.com/libsdl-org/SDL-1.2/issues/885) (probably will also work on OS/2)?
Sorry, no.

PR #5712, opened 2025-05-20
Fix potential memory leaks in drive_cache.cpp

maron2000:
Fix potential memory leaks due to initializing pointers to null but not releasing the memory in drive_cache.cpp.


maron2000:
Unfortunately this induced a regression.
Apologies for the confusions.

Issue #5584, opened 2025-03-24
Dynamic core can freeze on restart


henk717:
### Describe the bug

When using the dynamic core DOSBox-X may hang during certain reboots that would originally crash DOSBox-X instead.

### Steps to reproduce the behaviour

Install Windows 98 or in some cases install a driver on 98. The final reboot will hang, most reboots are fine.

### Expected behavior

Previously DOSBox-X would crash, ideally the restart workaround does not break.
Alternatively this can be solved by implementing DOSBox-Daum's workaround which restarts the program for dynamic core restarts.

### What operating system(s) this bug have occurred on?

Windows 10 21H2

### What version(s) of DOSBox-X have this bug?

DOSBox-X 2025.02.01 2025-02-01 13:48 (Tested on LowEnd SDL1 Win32)

### Used configuration

```ini
#Windows 98 compatible VirtualPC config by Henk717
#No drivers or partitioning required, just install and go! :D
#This is the DOSBox-X version of this config and may not work properly with other versions of DOSBox.

# A: mounts to A.IMG
# C: mounts to C.IMG (Created if not supplied)
# D: mounts to D.IMG (Created if not supplied)
# E: mounts to E.ISO (You can use other formats like BIN/CUE, simply rename them and it will work)
# F: mounts to F.ISO (You can use other formats like BIN/CUE, simply rename them and it will work)
# You can also make a folder called mount to pass trough files, if done the CD-rom drives shift up one letter.
# If you use this option it is recommended to do so prior to installing Windows.

[sdl]
aspect=false
fullscreen=false
fulldouble=true
fullresolution=desktop
output=opengl
autolock=true
autolock_feedback=beep
priority=higher,higher
showdetails=true

[render]
aspect=true
aspect_ratio=-1:-1
scaler=none

[dosbox]
title=VirtualPC Mode
memsize=128

[video]
vmemsize=8
vesa modelist width limit=0
vesa modelist height limit=0
allow high definition vesa modes=true
allow unusual vesa modes=true

[dos]
ver=7.1
hard drive data rate limit=0
floppy drive data rate limit=0

[cpu]
cputype=ppro_slow
#core=normal
core=dynamic_x86

[pci]
voodoomem=max

[fdc, primary]
int13fakev86io=true

[ide, primary]
int13fakeio=true
int13fakev86io=false

[ide, secondary]
int13fakeio=true
int13fakev86io=false
cd-rom insertion delay=4000

[ne2000]
ne2000=true
[...]

PR #5718, opened 2025-05-21
Change workaround for the dynamic core

henk717:
The existing workaround for the dynamic core hangs the process instead of crashing, it made the workaround worse than the original situation. This PR undoes the problematic commit and replaces it will a full restart command automatically relaunching the process.

#5584

No

The entire software is now restarted, this does not break for me but I can't rule this out on more exotic configurations.

## Additional information

I totally get it if you'd rather not merge this quick workaround, but as this Daum style restart solves the issue for me I wanted to offer the changes to motivate the issue being fixed.

### DIFF
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -3195,36 +3195,6 @@ bool CPU_CPUID(void) {
 	return true;
 }
 
-/* this is used by dynamic core as a workaround for sending the reset signal into the emulator
- * because C++ exceptions don't work from within dynamically generated code. */
-int reset_decode_signal = 0;
-Bits Reset_Decode(void) {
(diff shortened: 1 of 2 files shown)


joncampbell123:
Odd, I've been using the dynamic core code in question every time Windows 98 wants to reboot the system without any problems, Linux x86_64.

joncampbell123:
The idea with how DOSBox-X does it is that it signals the desire to reset and then lets the CPU core finish executing instructions.

Throwing a C++ exception with dynamic core may happen to work in Win32 but for most platforms, the C++ runtime cannot unwind the stack properly because it doesn't recognize anything that was dynamically generated.

If that method is causing DOSBox-X to hang, then additional work needs to be done so that on the CPU reset signal the core or normal loop needs to make sure it exits the core loop and then throws the desired exception.

Issue #5759, opened 2025-07-02
FPU emulation for FILD qword is broken in MacOS Arm


asemarafa:
### Describe the bug

Noticed a bug with FPU emulation on macOS M1. The FILD instruction with a 64-bit integer (qword) gives incorrect result.

### Steps to reproduce the behaviour

Compile and run this C code with Borland C++ inside DOSBox-X:

```c
#include <stdio.h>
#include <dos.h>

struct {
    unsigned long low;
    unsigned long high;
} my_qword = {123L, 0L};

long my_dword = 123L;

long double result_val;

void main(void)
{
    printf("This program directly tests the FILD instruction.\n");
    printf("It will load the integer 123 and store it back.\n\n");
    printf("Press Enter to begin...\n");
    getchar();

    printf("--- Testing 64-bit FILD (qword) -> 80-bit ST(0) -> 80-bit FSTP (tbyte) ---\n");

    asm {
        finit
        fild    qword ptr [my_qword]
        fstp    tbyte ptr [result_val]
    }

    printf("  Expected result: 123.000000\n");
    printf("  Actual result:   %Lf\n\n", result_val);

    printf("--- Testing 32-bit FILD (dword) for comparison ---\n");

    asm {
        finit
        fild    dword ptr [my_dword]
        fstp    tbyte ptr [result_val]
    }
    printf("  Expected result: 123.000000\n");
    printf("  Actual result:   %Lf\n", result_val);
}
```

Observed Output:

```
--- Testing 64-bit FILD (qword) -> 80-bit ST(0) -> 80-bit FSTP (tbyte) ---
  Expected result: 123.000000
  Actual result:   9223372036854776320.000000

--- Testing 32-bit FILD (dword) for comparison ---
  Expected result: 123.000000
  Actual result:   123.000000
```

The issue also reproducable by simply using functions like atof() and scanf() while dealing with doubles.

```c
#include <stdio.h>
#include <stdlib.h>
#include <dos.h>

void main(void)
{
	double d_val;
	char* str_to_convert = "123";
	asm { int 3 }
	d_val = atof(str_to_convert);
        asm { int 3 };
	printf("Result of atof(\"%s\"): %lf\n", str_to_convert, d_val);
}
```

This will also print 9223372036854776320.000000

I created inturrputs so I can get CPU Logs, that I attached.

Notable is 

```assembly
...
; Borland's internal code converts the string "123" to the integer 0x7B (123)
[...]

Issue #119, opened 2016-01-01
Normal core FPU emulation is incompatible with "Fast" Pentium memcpy trick


joncampbell123:
Ref: http://collaboration.cmc.ec.gc.ca/science/rpn/biblio/ddj/Website/articles/CUJ/1996/9612/durham1/durl1.htm

The normal core attempts to use the host floating point support to emulate Intel x87 instructions, but it does so by typecasting everything to "double", even 80-bit extended formats.

Unfortunately, this typecasting causes data corruption for any demo or game that uses FILD+FISTP (64-bit integer load and store) as a "faster" method of copying memory on Pentium-class (pre-MMX) hardware. On an actual processor, the 64-bit integer fits normally into the mantissa of the 80-bit FPU register without losing bits, while in DOSBox-X's normal core, the typecast to "double" loses bits and causes data corruption. In most demos, since the fast memcpy trick is used to copy from a system buffer into video memory, this corruption is visible as black vertical bars on the screen that sometimes shift or blend into the picture content depending on how the 64-bit int -> double conversion truncates.

joncampbell123:
Example: "Toontown" demo
![toontown_corrupted](https://cloud.githubusercontent.com/assets/6245486/12070858/7fea5480-b042-11e5-8348-b280cfb87973.png)

ghoul8765:
Reproduced your very interesting finding! PCem seems to correctly handle the above conversion in its "normal core"; I think below may be (abbreviated) hints.

x87.h:
/*Hack for FPU copy. If set then ST_i64 contains the 64-bit integer loaded by FILD*/
#define TAG_UINT64 (1 << 2)

x87.c:
uint16_t x87_gettag()
{
...
        for (c = 0; c < 8; c++)
                if (tag[c] & TAG_UINT64)
                        ret |= 2 << (c*2);
                else
                        ret |= (tag[c] << (c*2));

x87_ops.h:
static inline void x87_st_fsave(int reg)
{
        reg = (TOP + reg) & 7;
        if (tag[reg] & TAG_UINT64)
        {
        	writememl(easeg, eaaddr, ST_i64[reg] & 0xffffffff);
        	writememl(easeg, eaaddr + 4, ST_i64[reg] >> 32);
        	writememw(easeg, eaaddr + 8, 0x5555);
        }
        else
                x87_st80(ST[reg]);

static inline void x87_ld_frstor(int reg)
{
        reg = (TOP + reg) & 7;
[...]

PR #5760, opened 2025-07-02
Fix: Correct FILD emulation on platforms without long double

asemarafa:
This pull request fixes the FILD qword instruction, which was producing incorrect floating-point values on platforms with doesnt have an 80bit long double like M1 Mac.

The old code in FPU_FLD_I64 didn't do a proper conversion from a 64-bit integer to an 80-bit float.

This patch replaces the broken logic with a correct conversion. It first converts the integer to a double and then constructs a proper 80-bit long double from it, reusing the logic from FPU_ST80

#5759

### DIFF
--- a/src/fpu/fpu_instructions.h
+++ b/src/fpu/fpu_instructions.h
@@ -196,12 +196,30 @@ static void FPU_FLD_I64(PhysPt addr,Bitu store_to) {
 	FPU_Reg blah;
 	blah.l.lower = mem_readd(addr);
 	blah.l.upper = (int32_t)mem_readd(addr+4);
-	fpu.regs[store_to].d = static_cast<double>(blah.ll);
-	// store the signed 64-bit integer in the 80-bit format mantissa with faked exponent.
-	// this is needed for DOS and Windows games that use the Pentium fast memcpy trick, using FLD/FST to copy 64 bits at a time.
-	// I wonder if that trick is what helped spur Intel to make the MMX extensions :)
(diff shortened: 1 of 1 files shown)


joncampbell123:
I appreciate the fix, however please consider why the code stores to both the double and to the 80-bit long double format.

https://github.com/joncampbell123/dosbox-x/issues/119

In the 486 and Pentium era, before MMX, a "fast memcpy" hack was somewhat common where FILD and FSTP were used as a fast memcpy (64 bits at a time). It happens to work when the 64-bis are loaded into an 80-bit long double on x86 hardware because the mantissa is big enough to hold all the bits.

asemarafa:
Hi Jon,

Thanks for the explanation. You were right about the toontown demo in issue #119, my fix caused a regression.

I noticed dosbox staging handles this fast memcpy trick and dosent have the bug #5759 im trying to fix. They solve it with a shadow register. FILD_I64 saves the raw integer and FST_I64 checks if math happened, if not it uses the saved value for the memcpy trick otherwise it does a proper conversion.

I tried to copy this logic into my branch. Its a very simple attempt and i kept the changes in one file cause im not familiar with the codebase and didnt want to risk breaking anything else. but this approach works, it solves the fpu bug without the regression.

You can see the changes here.
https://github.com/joncampbell123/dosbox-x/compare/master...asemarafa:dosbox-x:fix/fpu

Let me know if you would be interested in merging this instead.

Torinde:
@asemarafa, since you tested it and it works OK for both cases I think it'll be good if you make a PR with it.

PR #5798, opened 2025-08-20
Fix mapper not rendering anything on SDL2/Windows

aybe:

Since my last contribution, the mapper would stubbornly render as a black screen.
After debugging, `DOSBox_SetMenu` is what makes the SDL surface become invalid.
Not sure why, maybe it's Windows' fault... But initializing it after fixes the problem.


src/gui/sdl_mapper.cpp
```diff
@@ -5308,6 +5311,18 @@ void update_all_shortcuts() {
         if (ev != NULL) ev->update_menu_shortcut();
 }
 
+void UpdateMapperSurface()
+{
+    mapper.surface = SDL_GetWindowSurface(mapper.window);
```
maron2000, line 5316: This function shall do nothing for SDL1 builds in order to fix the build failure.

PR #5799, opened 2025-08-20
Scale mapper according display resolution

aybe:

Eye comfort upgraded!

Before:

Screenshot 2025-08-21 010526

After:

Screenshot 2025-08-21 010542

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -46,6 +46,7 @@ Next version
   - PC-98: Redraw the function keys after updating them (bobsayshilol)
   - Do not carriage return with a single LF('\n') (maron2000)
   - Fix mapper not rendering anything on SDL2/Windows (aybe)
+  - Scale mapper according display resolution (aybe)
 
 2025.05.03
   - Show TURBO status in title bar. (maron2000)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -45,6 +45,7 @@ Next version
   - Fix errors when initializing fluidsynth (maron2000)
   - PC-98: Redraw the function keys after updating them (bobsayshilol)
   - Do not carriage return with a single LF('\n') (maron2000)
+  - Fix mapper not rendering anything on SDL2/Windows (aybe)
 
 2025.05.03
   - Show TURBO status in title bar. (maron2000)
--- a/src/gui/menu.cpp
+++ b/src/gui/menu.cpp
@@ -1868,6 +1868,9 @@ void SDL1_hax_SetMenu(HMENU menu) {
 extern "C" void SDL1_hax_SetMenu(HMENU menu);
 #endif
 
+/**
+ * NOTE: this function can make a SDL_Surface become invalid (e.g. mapper, Windows)
+ */
 void DOSBox_SetMenu(DOSBoxMenu &altMenu) {
 #if DOSBOXMENU_TYPE == DOSBOXMENU_SDLDRAW
     /* nothing to do */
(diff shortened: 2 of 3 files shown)


aybe:
By the way, if one looks closely, the scaling wasn't right before this PR anyway,

Here's a zoomed view of the initial behavior:

Screenshot 2025-08-21 022927

It turns out that the actual culprit is the menu with its height not being accounted for. If one tries to fix it by using Win32 menu-related functions to retrieve menu height, it fixes the mapper but in turn breaks the normal mode.

After looking at the code, it's pretty messy with windows dimensions being saved, reused and adjusted in multiples places.

The whole thing needs to be reworked, there shouldn't be that many things fiddling with window dimensions. (for another PR)

Torinde:
It seems you already support 2 of the 3 [Brazil keys](https://en.wikipedia.org/wiki/IBM_PC_keyboard#/media/File:Brazilian_104-key_ABNT2_keyboard.jpg) (numpad comma and one next to each Shift).

Can you please add the third button (left of the right Shift, which can be shortened to the same size as Enter)?

Related:
- https://github.com/dosbox-staging/dosbox-staging/pull/2209

aybe:
> By the way, if one looks closely, the scaling wasn't right before this PR anyway,
> 
> Here's a zoomed view of the initial behavior:
> 
> Screenshot 2025-08-21 022927
> It turns out that the actual culprit is the menu with its height not being accounted for. If one tries to fix it by using Win32 menu-related functions to retrieve menu height, it fixes the mapper but in turn breaks the normal mode.
> 
> After looking at the code, it's pretty messy with windows dimensions being saved, reused and adjusted in multiples places.
> 
> The whole thing needs to be reworked, there shouldn't be that many things fiddling with window dimensions. (for another PR)

After thinking for a bit, I don't think it's worth trying to fix the problem:

- no one wants a microscopic 1:1 mapper on today's screens like 4K
- thus, scaling will always be > 1, therefore, the problem vanishes by itself
- too many structural changes I'm not comfortable to assess and test

aybe:
> It seems you already support 2 of the 3 [Brazil keys](https://en.wikipedia.org/wiki/IBM_PC_keyboard#/media/File:Brazilian_104-key_ABNT2_keyboard.jpg) (numpad comma and one next to each Shift).
> 
> Can you please add the third button (left of the right Shift, which can be shortened to the same size as Enter)?
> 
> Related:
> 
> * [Handle OEM102 and ABNT1 keyboard scancodes dosbox-staging/dosbox-staging#2209](https://github.com/dosbox-staging/dosbox-staging/pull/2209)

Nope, sorry.
There appears to be more things being done in that PR than its title, and it drifts from what's in here, e.g. no oem102 here, etc.
The author likely knew what he was doing, not my case, I'd rather not screw something else in the process.

Torinde:
Oh, I'm not asking to adopt that PR, just gave it as example.

The key defined is as follows:
- ABNT1 - Key to the left of Right Shift on ABNT layouts
- "ABNT_C1" at [kbdlayout.info](https://kbdlayout.info/KBDBR/overview+scancodes?arrangement=ABNT)
- SDL scancode 135 (0x87) maps to first ABNT key with scancode 115 (0x73)
- SDL_SCANCODE_INTERNATIONAL1 = 135
- INT 3 at http://www.quadibloc.com/comp/scan.htm

So, only to make the connection of SDL_SCANCODE_INTERNATIONAL1 (0x87) to scancode 0x73 and then carve out the graphic in the mapper to the left of Right Shift.

aybe:
Nope, sorry.

PR #5814, opened 2025-08-31
Save Kanji CG access mode setting, restrict output accordingly

OPNA2608:
With an option for disabling this behaviour, since some machines don't seem to run into this issue when playing certain translation patches.

Playing the current translation patches for the following games on real hardware reveals an issue with the way text is drawn: It mostly consists of white blocks.

- Aegean Kai no Shizuku (https://github.com/Svipur/Aegean-Kai-No-Shizuku-PC98-English/issues/1)
- Azusa 999
- Rance 4.\*

These are caused by failing character bit reads from the Kanji Character Generator. Retrieving the bitmap of an ANK character (ASCII & Kana), or a Kanji while in Code Access mode, via port reads returns invalid values if a VSync is not currently in progress (Undocumented 9801/9821 says it's "not possible").

These translations were tested against emulators, which don't seem to implement this detail. So I'm implementing this here, in hopes of avoiding late surprises in future projects.

The newly introduced `pc98->pc-98 chargen vsync-limited access` option, which is default-enabled, will make reads within certain character ranges & access types from port `0xA9` return `0xFF` outside of VSync. This matches the behaviour found on various, but seemingly not all, models of 9801s and 9821s.

image

This will "break" some existing translation patches, in the same way that they're currently broken on real hardware. Until those patches are fixed, the option can be disabled in the settings when necessary.

I'm not sure if any other software that exists out in the wild relies on this behaviour working this way, either in emulators or on models that don't seem to have this restriction (whichever those are, only know that the blocks don't seem to appear on a PC-9821As3).

## Additional information

[`pc-98_translation_discussion`](https://discord.com/channels/231947749862539264/231949854463623168/1411356509056532581) channel on the PC-98 discord server, where we slowly worked out all the details of this.

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -2436,6 +2436,11 @@ void DOSBOX_SetupConfigSections(void) {
                     "at driver startup AND when INT 33h AX=0 is called. This is NEC MOUSE.COM behavior and default\n"
                     "enabled. To emulate other drivers like QMOUSE that do not follow this behavior, set to false.");
 
+    Pbool = secprop->Add_bool("pc-98 chargen vsync-limited access",Property::Changeable::WhenIdle,true);
+    Pbool->Set_help("If set, reading pixels from the character generator while in Code Access mode (or always, for ANK\n"
+                    "characters) will be invalid. Some models (i.e. PC-9821As3) dont seem to have this limitation, but\n"
+                    "many others do.");
+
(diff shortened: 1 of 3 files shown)


maron2000:
~I don't think it is good to hack the emulator to avoid a bug in the game or patch.~

Edit:
Sorry, I misunderstood your explanation. Maybe good to make this an option.

Issue #5836, opened 2025-09-09
Sync with SVN r4491-r4494


Torinde:
### What you want

There are new SVN commits:
- https://sourceforge.net/p/dosbox/code-0/4491/ Add floppy disk image support for more extended-size formats.
- https://sourceforge.net/p/dosbox/code-0/4492/ As verified on real hardware, the MPU-401 and compatibles do not clear their message queues on reset. Fixes Roland music in Best of the Best Championship Karate.
- https://sourceforge.net/p/dosbox/code-0/4493/ Handle EOF character in batch files as actual end of file. Fixes install batch file of Crime Does Not Pay.
- https://sourceforge.net/p/dosbox/code-0/4494/ Use 4-byte Partial Page Mapping data as EMM386 and other memory managers do. Fixes Kyrandia II: Hand of Fate with ems=true.

Please check if those are applicable. @Allofich

### Additional information

- previous #5801

### Have you checked that no similar feature request(s) exist?

- [x] I have searched and didn't find any similar feature request.

### Code of Conduct & Contributing Guidelines

- [x] I agree to follow the code of conduct and the contributing guidelines.

Allofich:
Added a PR for r4492, r4493 and r4494.

r4491 will need filling in the missing fields, as DOSBox-X has additional geometry fields not in regular DOSBox.

This happened previously with https://github.com/joncampbell123/dosbox-x/issues/1472.

Allofich:
r4491 adds support for:

	{1520, 19, 2, 80, 2},	// DS/HD 5.25" (XDF)
	{1720, 21, 2, 82, 4},	// DS/HD 3.5"  (DMF)
	{1840, 23, 2, 80, 4},	// DS/HD 3.5"  (XDF)

The 1520 KB and 1840 KB XDF formats are already supported in DOSBox-X. As for the 1720 KB DMF format, I think this might be a mistake. At https://forum.winworldpc.com/discussion/9191/extract-1-72mb-diskettes-from-images-windows-95b-installer, someone wrote:

> DMF size is exactly 1.68MB (1,720,320 Bytes.)
> 
> *1720320 / 1024 = 1.68
> 
> A few of user think DMF size is 1.72MB , but 1.68MB is correct exactly.

So I'm considering this issue taken care of by the linked PR now.

maron2000:
[...]

PR #5843, opened 2025-09-12
SVN r4492, r4493, r4494

Allofich:
Adds changes from:
https://sourceforge.net/p/dosbox/code-0/4492/
https://sourceforge.net/p/dosbox/code-0/4493/
https://sourceforge.net/p/dosbox/code-0/4494/

Closes https://github.com/joncampbell123/dosbox-x/issues/5836

### DIFF
--- a/src/ints/ems.cpp
+++ b/src/ints/ems.cpp
@@ -624,7 +624,7 @@ static uint8_t EMM_GetPagesForAllHandles(PhysPt table,uint16_t & handles) {
 }
 
 static uint8_t EMM_PartialPageMapping(void) {
-	PhysPt list,data;uint16_t count;
+	PhysPt list,data;uint16_t count,page;
 	switch (reg_al) {
 	case 0x00:	/* Save Partial Page Map */
 		list = SegPhys(ds)+reg_si;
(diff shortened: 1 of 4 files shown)


src/ints/ems.cpp
```diff
@@ -634,37 +634,40 @@ static uint8_t EMM_PartialPageMapping(void) {
 		for (;count>0;count--) {
 			uint16_t segment=mem_readw(list);list+=2;
 			if ((segment>=EMM_PAGEFRAME) && (segment<EMM_PAGEFRAME+0x1000u)) {
-				uint16_t page = (unsigned int)(segment-EMM_PAGEFRAME) / (unsigned int)(EMM_PAGE_SIZE>>4u);
```
joncampbell123, line 637: Does this fix any expanded memory (EMS) bugs? Just asking.

It seems from the standard and RBIL that the call defines only that you can save/restore the data. It doesn't define what the data structure actually is, except that you can query the size, save, and restore it.

Allofich: The original commit message is "Use 4-byte Partial Page Mapping data as EMM386 and other memory managers do. Fixes Kyrandia II: Hand of Fate with ems=true." I haven't tested in DOSBox-X.

I'm fine with dropping it from the PR if you want.

maron2000: I tried to launch Kyrandia II with the original code, and it did start with ems=true.
I don't know how far I should proceed in the game to reproduce the bug.

Allofich: Presumably the DOSBox change matches more closely what EMM386 and compatibles do.

Allofich: > I tried to launch Kyrandia II with the original code, and it did start with ems=true. I don't know how far I should proceed in the game to reproduce the bug.

Thanks for testing. I can't find reference to the original problem that DOSBox supposedly fixed. If you have some time, can you try testing it a bit further, like the following:

1. Enable EMS (ems=true).
2. Start Kyrandia II with a sound card selected.
3. Watch the entire intro cinematic.
4 .Save the game in the first area, then load it.
5. Play long enough to leave the starting area.

Allofich: I tested myself. The only difference I can find before and after this PR is that, before this PR, if running the game with MT-32 for music (not SoundBlaster) and you start a new game or load a game and then try to exit the program, it hangs on exit. With this PR applied, it doesn't hang. I tested multiple times and it seems to be consistent.

joncampbell123: Huh. Before the PR, quitting the game caused a storm of INT 6 invalid opcode exceptions. After the PR, it exists normally.

PR #5854, opened 2025-09-20
Debugger: Warn single-stepping may not work as expected with Dynamic core

### LINKED ISSUE #5488: Print a warning in the debugger, at least once, if you break into the debugger while using dynamic core
### Is your feature request related to a problem? Please describe.

Debugging is more difficult with dynamic core because of the way that the JIT groups instruction handling together, and some breakpoints don't work.

### What you want

Print a warning at least once letting the user know that for full debugging, they should switch to normal core.

### Describe alternatives you've considered

_No response_

### Additional information

_No response_

### Have you checked that no similar feature request(s) exist?

- [x] I have searched and didn't find any similar feature request.

### LINKED ISSUE #5848: Debugger appears to jump over many instructions running in protected mode
### Describe the bug

I'm trying to debug a DOS protected mode program that runs with DOS/4GW. At first I thought the bug was that breakpoints I set weren't being hit, but I think the real problem is that the debugger just jumps over a bunch of instructions (well they're executed, but you can't seem to stop on them in the debugger). Here's just one example of the debugger jumping over one jne instruction when I hit F11; it usually jumps over more than 1 though:

**Before**
Image

**After**
Image

1. Run a protected mode dos EXE in DOSBox-X.
2. Enter debugger.
3. Hit F11 to step to next instruction.

### Expected behavior

Steps to next instruction.

Windows 10

2025.05.03

### Used configuration

```ini

```

### Output log

```text

```

### Additional information

_No response_

maron2000:
Warn users not to use Dynamic core if single stepping matters when debugging.

Fixes #5488
Fixes #5848

image

### DIFF
--- a/src/debug/debug.cpp
+++ b/src/debug/debug.cpp
@@ -186,6 +186,8 @@ void DEBUG_EndPagedContent(void);
 Bitu MEM_PageMaskActive(void);
 Bitu MEM_TotalPages(void);
 Bitu MEM_PageMask(void);
+void DEBUG_WarnDynamic(void);
+int CPU_IsDynamicCore(void); // cpu.cpp 0=normal, 1=dynamic, 2=dynamic core with assembler
 
 static void LogEMUMachine(void) {
     DEBUG_BeginPagedContent();
@@ -313,6 +315,8 @@ static bool debug_running = false;
 static bool check_rescroll = false;
 
 static FPU_rec oldfpu;
+static bool warn_dynamic = false;
+
 
 void VGA_DebugRedraw(void);
 
@@ -4306,7 +4310,12 @@ uint32_t DEBUG_CheckKeys(void) {
 				break;
 		case KEY_F(10):	// Step over inst
 				DrawRegistersUpdateOld();
-				if (StepOver()) {
+                if(!CPU_IsDynamicCore()) warn_dynamic = false;
+                else if(!warn_dynamic) {
+                    DEBUG_WarnDynamic();
+                    warn_dynamic = true;
+                }
+                if (StepOver()) {
 					mustCompleteInstruction = true;
 					inhibit_int_breakpoint = true;
 					ret = DEBUG_Run(1,false);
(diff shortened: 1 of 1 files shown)


jez9999:
@joncampbell123 If `warn_dynamic` is a flag indicating whether the dynamic warning logic has been executed (which it seems to be), shouldn't it get set to true here?

```
if(!CPU_IsDynamicCore()) warn_dynamic = false;
```

Alternatively, if the flag means "you should do the dynamic warning" then this code seems wrong:

```
if(!warn_dynamic) {
    DEBUG_WarnDynamic();
    warn_dynamic = true;
}
```

maron2000:
This is not to warn many tImes. If warn_dynamic is true, it will not warn again.
Or do you want it to warn every step?

jez9999:
@maron2000 So it sounds like it's a "has warned" flag.  In which case:

```
if(!CPU_IsDynamicCore()) warn_dynamic = false;
```

Doesn't seem to make sense because don't you wanna say "CPU isn't dynamic, no need to warn, so consider 'has warned' to be true"?

maron2000:
It resets the "has warned" when you switch to normal core, so that if you switch to dynamic, it will remind you again just In case.

I tried it so that it works as intended.
Did you find something wrong?
If so, in which case?

jez9999:
> It resets the "has warned" when you switch to normal core, so that if you switch to dynamic, it will remind you again just In case.

So it's like a `done_debug_warning` flag.  What I don't understand is:

```
if(!warn_dynamic) {
    DEBUG_WarnDynamic();
    warn_dynamic = true;
}
```

Because this code doesn't look like it's just supposed to execute when you have a dynamic core.  So if this is executing when you have a normal core, it'll display the warning message.  Is that intended?

maron2000:
Please explain an actual procedure that doesn't work as intended.
I'm now strongly discourage to deal with you, that you say "incorrect" or "doesn't make sense" without a concrete procedure.

PR #5883, opened 2025-10-04
Add UnifontEX

stgiga:

UnifontEX is basically the evolved form of the Unifont TTF I contributed many years ago during the start of PC-98 support. It is quite a lot more compatible and the glyph shapes follow Unicode 15, not 2017 Unicode. AND for anyone doing stuff like Lotus Multi-Byte Character Set, DOS/V, and such, it's better. ALSO, TrueType-wise, there's more tables. I also contributed BDF and FontX2 versions because of DOSBox-X supporting those in DOS/V mode. So there is that going for it. Also of note is that this isn't the end. I'm projected to run out of codepoints when Unicode 18 or 19 drop (it depends on how many Plane 0 glyphs the latter adds), plus a year to wait for Unifont upstream to finish said release. 

It is a significant update over what was here.

If there are any related issues, reference them here. Ex. #⁠1234, Fixes #⁠1234, Closes #⁠1234, etc.

If yes, describe the feature(s) introduced.

WELL, there's a LOT more characters, and the glyph shapes are newer, and importantly, the TrueType now has a LOT of tables that improve font renderer behavior compared to what was previously there.

## Additional information

This change is designed to make DOS/V, JEGA, FM Towns, and PC-98 support better, but it can handle WAY more than that.


maron2000:
Great, but what are those files not related to UnifontEX?
I don't see the requirement of including np2 for Wii (18MBytes!) or those FREECG98.BMP files.

stgiga:
> Great, but what are those files not related to UnifontEX? I don't see the requirement of including np2 for Wii (18MBytes!) or those FREECG98.BMP files.

Oh those are from 2017, and were used to create the first version of FREECG98 I did for DOSBox-X back then, when there was no such thing as TrueType support and an ANEX86.BMP replacement was involved. I had also thrown in that ancient UnifontEX version before it was even known as UnifontEX. So they aren't junk. I don't know why when I was in HS I included the Wii homebrew version of NP2 but it was the only version of NP2 that generates ANEX86 BMPs from TTFs that still works and on dev Dolphin it works fine. Not that i expect merging it. I just forgot to clean it out. Don't worry. Also I've been in the Discord for DOSBox-X since 2024 so I'm not a bot.

maron2000:
I'm not suspecting you a bot, but you have to remove the unrequired files.
You are now trying to include a bloat for all.

maron2000:
By the way, I see alphabets, numbers, half-width kanas are kind of fancy in FREECG98.BMP that you included in this PR.
I don't feel its suitable for font in console.
You can see that your BMP, the characters are kind of round and is like handwriting.

Your BMP  ------->   Current BMP

image  
image


I think it is a double-byte half-width characters.
PC-98 uses slightly different glyphs for single-byte and double-byte half-width characters, so using different glyphs itself makes sense.
image

Edit: single-byte 円, 年, 月, 日, 時, 分, 秒 (characters on the right side) are fancy as well
image

maron2000:
The current BMP wasn't bug free.
It also had some fancy characters.

image<br>

image

maron2000:
UnifontExMonoAnk.fontx2 file has some flaws in box-drawing character?

Sorry for the lots of posts. I just wanted to make things better.

image

Changing the `fontxsbcs` option to `GONHN16X.tlf` works fine.
https://web.archive.org/web/20241128212629/http://silo.la.coocan.jp/FONTV/VGON16.LZH

image

joncampbell123:
Is this pull request ready to merge?

stgiga:
Look, I accidentally forgot to remove the 2017 stuff, including that revision of FREECG98.BMP, so pay it no mind. It's the TTF, BDF, and FontX2 that matters.

maron2000:
You can also delete the existing Unifont file in your PR.
We don't need two similar files and your TTF is an upper compatible one isn't it.

stgiga:
> You can also delete the existing Unifont file in your PR. We don't need two similar files and your TTF is an upper compatible one isn't it.

Yeah. Honestly my experience with Git is a bit messy. I just don't know how to safely clean.

stgiga:
Sorry for all of the confusion, and I wasn't trying to create duplicates. That being said if this only wants to be added *once*, I guess I could wait until Unicode 18/19 when I get to 65535 and go through the server again if that works. I just wanted to let you know that non-TTF DOS/V versions exist when they previously didn't on the inferior version. The BDF will get updates into the max, but the good news is the FontX2 files are stable. So if we do wait you can just use those as a stopgap. My apologies for this mess.

maron2000:
I can assist you on making a simple PR if you wish.
Add the BDF, fontx2, and replace the TTF and FREECG98.BMP is what you want to do?

stgiga:
> I can assist you on making a simple PR if you wish. Add the BDF, fontx2, and replace the TTF and FREECG98.BMP is what you want to do?

Actually if you're going to replace FREECG98.BMP, https://stgiga.github.io/UnifontEX/UFEXPC98.BMP is it. 

# HOWEVER

DOSBox-X contributors *already* improved FREECG98.BMP in ways NP2 can't do. So while I'm not against the work being repeated on the newer base, strictly speaking you `don't` *have* to do it. Now, the BDF, FontX2, and TTF are the ones that matter here. The 2017 TTF was not ideal, and it's missing some helpful tables like BASE and VDMX that can help with rendering. It also has far better (and improving) Unicode support, and its glyph shapes are better, especially the Korean, Japanese, and fullwidth characters, but not just those. It is a direct upgrade.

maron2000:
So PR #5884 just adds the BDF, FontX2, and TTF files, which are the ones you want to add.
You can close this PR.

joncampbell123:
Will do. Thank you.

Issue #3214, opened 2022-01-11
TSR Trainer functionalities like Game Wizard


Torinde:
> > @bastetfurry I've heard that Action Replay and Game Genie intercept reads (writes?) to certain addresses and replace the byte value if they match, correct?
> 
> Yes, these work like that or they simply force a value to be set in stone. Sometimes they even patch the code so that a DEC $lives gets turned into a bunch of NOPs.
> 
> The working ones for DOS tough needed to be trained to a game every time the game is started, tough. So you start, for example, Sim City, start a new game, tell your cheat tool to scan the memory for 10000, it will create a list of all addresses that have said value, go back to the game and buy something, search which memory addresses changed their value to the new one, rinse and repeat until you have your address where the ingame money is stored.

Game Wizard was great. To have its functions built-in will be amazing!
As @bastetfurry explained, and if I recall correctly - it worked by searching for a user provided value in memory, logging where it's found and allowing the user to modify/freeze it. Depending on the game you had to search for the value as number (e.g. 5) or characters (e.g. the 0x35 representing the character "5").

Nice feature will be to log also ANY changed values (I don't think GW did that).

>> What do you anticipate the interface should look like?

[Vogons disucussion including screenshots](https://www.vogons.org/viewtopic.php?t=30322)
[Another discussion about it](https://misterfpga.org/viewtopic.php?t=1429)
[Download latest Game Wizard](https://anapan.525lines.moe/GameWizard.htm)

Maybe an extension of Debug or Capture menus? Or a separate Trainer menu?

If needed, I can try to run GW32 and capture more screenshots.

_Originally posted by @Torinde in https://github.com/joncampbell123/dosbox-x/issues/1184#issuecomment-1009978031_

PR #5901, opened 2025-10-10
Add debug feature memory find

Enmet:

I was documenting addresses in an old game some time ago, but I couldn't find a way to directly search for values in RAM. Many emulators have some sort of RAM search (also known as cheat finder) that allows the user to search for specific values with the addition of searching for changes over time (e.g. player moves forward and the x-position has increased). Initially, I used MEMDUMPBIN coupled with a python script but that got old fast, so I figured I'd try implement the feature directly.

This PR introduces two new debug commands: **MEMFIND** and **MEMS**.

**MEMFIND** - Starts a memory search instance within the specified segment, offset and range, with byte size being an optional parameter. This creates two arrays of the given size: One that is a direct copy of the memory at that point and the other being a truth table for said value. The truth table starts with all values set as "_true_".

**MEMS** - This is used to iterate through the search instance started with the former command. The user may choose to match values directly or relative to an operator. Valid operators are "**EQUAL TO** (**=**), **NOT EQUAL TO** (**!**), **LESS THAN** (**<**), **GREATER THAN** (**>**), **GREAT THAN OR EQUAL TO** (**>=**), **LESS THAN OR EQUAL TO** (**<=**)". No operator defaults to **EQUAL TO**.

Every time **MEMS** is used, the values in the array are updated. This means that the compared value is always relative to what the value was during the last search, or **MEMFIND** initialization if no **MEMS** command was entered previously in the instance.

Non-matching values will be set to "_false_". This process of elimination means that the user can keep sorting out uninteresting values until a manageable amount is left. The remaining addresses that are "_true_" can then finally be listed using "**MEMFIND L**". Only one search instance can be active at a time. The current search instance can be cancelled with "**MEMFIND !**". If there are no more matching values, the search instance automatically ends.

### Example use case

Figuring out values such as in-game lives are fairly straight forward. For example, if the player has **5** lives, "**MEMS 5**" can be used. Then after a life has been lost, "**MEMS 4**", etc.

However, trying to figure out values that are not explicitly shown is a bit more tricky, and this is where this feature really shines. Let's say that the user wants to find the address for the x-position coordinates:

"**MEMFIND 00:010000 FFFF W**" - This starts a search instance between 010000 - 01FFFF. 2 byte value specified.
"**MEMS ! 0**" - We know the position is _not_ 0 at least. Move the player slightly to the right after this.
"**MEMS >**" - Given that the game uses common 2D positioning, the value should increase after moving right. This line should therefor match because the value was increased when the player moved. Let's move the player to the left before doing another search.
"**MEMS <= 5D**" - Since we moved the player to the left, the value is now smaller than the value at the previous search. Let's also assume we know that the value is smaller than or equal to hex 5D.

... We can keep doing this while trying to rule out other factors such as screen/border x-pos, game obstacles, etc. Then, when there are only a few addresses left matching we can finally list them.

"**MEMFIND L**" - These addresses can now be viewed or poked to confirm.
"**MEMFIND !**" - Let's end the search instance before we want to find a new address.

No.

### Note:

While this has been working well for my own purposes, I'm not too confident with C++ so some scrutiny should probably be applied, especially since dynamic arrays are involved.

The maximum array size is set to 16 MB after the default VM memory size. This could probably be changed to the user defined size, or some other appropriate value.

There is a small bug if the user does not specify a range. The search instance will then be of size "0D", it is not breaking but I couldn't quite figure this one out.

### DIFF
--- a/src/debug/debug.cpp
+++ b/src/debug/debug.cpp
@@ -281,6 +281,21 @@ public:
 };
 #endif
 
+typedef struct MEMFinder {
+	bool usePreviousValue = false;
+	uint8_t opType = 0;
+	uint8_t size = 1;
+	uint16_t iterations = 0;
+	uint16_t seg = 0;
+	uint32_t ofs = 0;
+	uint32_t range = 0;
+	uint32_t value = 0;
+	uint32_t matches = 0;
+	vector<uint8_t> tableValue;
+	vector<bool> tableTruth;
+}MEMFinder;
+
+MEMFinder* MEMFINDInstance = NULL;
 
 class DEBUG;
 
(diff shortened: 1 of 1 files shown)


Torinde:
Related:
- #3214

Issue #5974, opened 2025-12-03
Cannot Remove Save State Files (File Lock Prevents Deletion)


srcdabada:
### Describe the bug

The save state files created by DOSBox-X are continuously locked by the emulator when it is running. This file lock prevents the built-in "Remove state in slot" function from working, causing the error message "Failed to remove the state in the save slot." every time.

* Core Issue: DOSBox-X locks existing .sav files immediately upon startup. This lock is only released when the emulator is fully closed. Because of this, manual deletion via the OS file explorer is impossible while DOSBox-X is open.
* Important: Saving a new state (overwriting the old one) and Loading a state both work fine. Only the removal/deletion process fails because of the file lock.

### Steps to reproduce the behaviour

1. Start DOSBox-X and load a DOS program/game.
2. Create a save state file in a slot (e.g., slot 1).
3. [Key Scenario] Quit DOSBox-X, then relaunch it. (The file handle is clearly closed during shutdown, then reopened upon relaunch.)
4. Try to remove the state file through the DOSBox-X menu.
5. Result: The error message "Failed to remove the state in the save slot" appears.
6. (Optional) While DOSBox-X is still running, try to delete the state file manually from your computer's file system.
7. (Optional) The operating system says the file is in use and cannot be deleted until DOSBox-X is closed.

### Expected behavior

When I choose to "Remove state in slot", DOSBox-X should:
1. Release the file lock and successfully delete the file.
2. Finish the process without showing an error message.
* Suggestion: It is likely that DOSBox-X keeps files locked to show accurate menu information. However, I suggest that the .sav file should only be opened for reading or writing when the Save/Load state action is specifically started, and then closed right away. Allowing the user file management freedom (even if it causes the menu information to be inaccurate) and showing a simple 'file not found' error upon loading is the preferred approach.

### What operating system(s) this bug have occurred on?

Windows 10 Pro (64-bit) 22H2 19045.6575

### What version(s) of DOSBox-X have this bug?

dosbox-x-vsbuild-win64-20251201165421.zip

### Used configuration

```ini

```

### Output log

```text

```

### Additional information

<img width="1354" height="1803" alt="Image" src="https://github.com/user-attachments/assets/9665eea1-b3f1-45d0-a29b-4c3aeba86a31" />

### Have you checked that no similar bug report(s) exist?

- [x] I have searched and didn't find any similar bug report.

### Code of Conduct & Contributing Guidelines

[...]

PR #5975, opened 2025-12-03
Fix: File handle leak prevents save state file removal

srcdabada:
Hello,

This Pull Request addresses a critical bug where save state files (.sav) could not be removed or deleted while DOSBox-X was running. This issue led to "Failed to remove the state" errors.

The Problem

The function SaveState::getName reads metadata (such as Program Name and Timestamp) from the .sav file for menu display. Since the file is a ZIP archive, it is opened using the C-based minizip function unzOpen2_64. The code was missing the corresponding C cleanup function unzClose(zf). This omission caused a persistent Operating System File Handle Leak on the file, preventing external deletion.

The Fix

The fix involves adding a single line to ensure the file handle is properly released after the metadata is read.

#5974

### DIFF
--- a/src/misc/savestates.cpp
+++ b/src/misc/savestates.cpp
@@ -924,6 +924,7 @@ std::string SaveState::getName(size_t slot, bool nl) const {
 			if (length != 0) ret += nl?"Remark: "+(!strlen(buffer1)?"-":std::string(buffer1))+"\n":" - "+std::string(buffer1);
 		}
 	}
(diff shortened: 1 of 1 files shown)


joncampbell123:
Ah.

The changes I made some months ago to improve performance seems to have caused it :(
Prior to my change, it basically opened and closed the ZIP archive for every single file it added.
Thank you for catching this.

PR #6009, opened 2025-12-21
Fix build errors for platforms not supporting xBRZ

maron2000:
Fix build errors since commit 0a3cdef, for platforms not supporting xBRZ.


joncampbell123:
Ah, thank you.

PR #6026, opened 2025-12-30
Fix build error by replacing alloca() with std::vector<uint8_t>

maron2000:
Fix HX-DOS build failure by replacing alloca() with std::vector<uint8_t>.
This also reduces the risk of excessive stack allocation.


joncampbell123:
Ah, beat me to my own TODO :)

Tbank you!

PR #6047, opened 2026-01-08
Convert all bash-dependent shell files to POSIX-shell compliant

pandasoli:
I replaced all GNU Bash-dependent shell files to POSIX-shell compliant code allowing it to be compiled in very minimal systems like ones that use Busybox or Dash and the only shell removing Bash from the list of compile time dependencies.

No.

No.

## Additional information

I'd like someone to review these specific files that included arrays. I am not sure if I converted correctly and I didn't know a way to test them.
```
./vs/sdl2/build-scripts/androidbuild.sh
./mtavi2wav.sh
./vs/libpng/ci/ci_verify_cmake.sh
./vs/libpng/ci/ci_verify_makefiles.sh
```

The ones related to compilation are fine, I tested compiling the project on my machine at least (Linux x86_64 musl).

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -2,7 +2,7 @@
 
 # If an error occurs, quit the script and inform the user. This ensures scripts
 # like ./build-macos and ./build-macos-sdl2 etc. Don't continue on if Autotools isn't installed.
-function finish() {
+finish() {
   if [ "${success}" -eq 0 ]; then
     echo 'autogen.sh failed to complete: verify that GNU Autotools is installed on the system and try again'
   fi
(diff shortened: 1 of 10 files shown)


autogen.sh
```diff
@@ -1,8 +1,8 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # If an error occurs, quit the script and inform the user. This ensures scripts
 # like ./build-macos and ./build-macos-sdl2 etc. Don't continue on if Autotools isn't installed.
-function finish {
+function finish() {
```
maron2000: Remove `function` ?

PR #6051, opened 2026-01-10
Use MACH_PORT_NULL instead of kIOMainPortDefault for old macOS

maron2000:
`kIOMainPortDefault` is not supported on old macOS (< 12.0), so replace it with its synonym `MACH_PORT_NULL` to support it.

### DIFF
--- a/build-macos-sdl2
+++ b/build-macos-sdl2
@@ -65,14 +65,6 @@ if [ -n "${x86_64_brew_cmd}" ] && [ "${universal}" -eq 1 ]; then
     x86_64_brew_cmd="/usr/bin/arch -x86_64 ${x86_64_brew_cmd}"
 fi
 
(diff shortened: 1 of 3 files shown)


joncampbell123:
It is possible to use macros defined by Mac OS to determine the API version at compile time. For a good example, look at my Apple Touch Bar support in src/gui/menu_macos.mm.

The macro used is MAC_OS_X_VERSION_MAX_ALLOWED.

Please update the code to use that constant only if MAC_OS_X_VERSION_MAX_ALLOWED is low enough to represent a version of Mac OS that needs MACH_PORT_NULL.

maron2000:
The one you mentioned works if you build on old macOS.
It is already included in the SDL2 code.
The fix is to make binary built on latest macOS run on old macOS as well.
I will fix as you mentioned soon.

PR #6067, opened 2026-01-15
Fix savestate loading crash related to dynamically allocated TempLine in VGA system

### LINKED ISSUE #6048: Savestate consistent crash
### Describe the bug

It seems the savestate system is a bit unstable now. Actually kinda unusable with the game i tried, which worked in the past ( though it was like 2021 and with the SDL1 build ).

I used it a few years ago for the character creation in a certain game.
Back then i had only issues with the SDL2 build, while the SDL1 build worked fine.
Now the crashing happens on both.

When i use it for the same game, crashes happen after like 2 loads of the state.

1. Start Realms of Arkania
2. Go into character creation
3. Save state
4. Do some stuff
5. Load state
8. Dosbox X crashes after 2 or so loads ( the first load succeeds )

### Expected behavior

No crash, savestate loaded

Win 10

2026.01.02

### Used configuration

```ini
Doesn't seem to matter, issue happens with default config too.
```

### Output log

```text

```

### Additional information

_No response_

rebb:
Fix savestate loading crash related to dynamically allocated TempLine in VGA system.

Previous code didn't account for TempLine becoming dynamically sized.

Fixes #6048

## Possible breaking changes

Since this changes how the VGA part in the savestate is written, it probably breaks savestate compatibility with previous versions.

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -8403,7 +8403,8 @@ void POD_Save_VGA_Draw( std::ostream& stream )
 
 
 	// - pure data
-	WRITE_POD( &TempLine, TempLine );
+	WRITE_POD( &TempLineSize, TempLineSize );
+	WRITE_POD_SIZE( TempLine, TempLineSize );	
 
 
 	// - system data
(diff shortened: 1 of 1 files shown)


rebb:
Actually i'm unsure if the reading code does the right thing. Would like some comments.

Maybe reusing the TempLine address if the Size is the same is not a good idea, and it should always reallocate ? 🤔
Then again it might be fine. Late night doubts, sorry 😄

joncampbell123:
> Actually i'm unsure if the reading code does the right thing. Would like some comments.
> 
> Maybe reusing the TempLine address if the Size is the same is not a good idea, and it should always reallocate ? 🤔 Then again it might be fine. Late night doubts, sorry 😄

Let me know when you're ready for me to merge this code.

rebb:
I think it should be fine. Do you see anything that might be an issue ?

I've been playing with this change for a while in a game where i use a lot of savestating, and so far i haven't noticed additional issues compared to the previous DosBox version that had a static TempLine.

I also just tested it with Terra Nova, and saving/loading states between different resolutions seems to work well.

Canalboy:
Hello there.   I have just noted the same problem with save states closing DOSBox-X 2026.01,02 (VS SDL1 64-bit) Jan 02 2026 via Windows 10 22H2.  Is the bug fixed version available anywhere as I'm not too used to navigating Github.  Cheers.

PR #6099, opened 2026-02-04
D3D11: Workaround for blanked screen in CGA/EGA/Tandy/Hercules modes

maron2000:
The experimental D3D11 output option somehow fails to display text & graphics on CGA/EGA/Tandy/Hercules modes.
This PR bypasses the existing renderer using a simple renderer as a workaround while figuring out the reason.
Due to the above workaround, some scalers/shaders are not working but better than a blank screen.

image

image

### DIFF
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -326,11 +326,21 @@ static void RENDER_ScalerLineHandler(const void * s) {
 	}
 }
 
+void D3D11_DrawLine_8bpp(const void* src);
+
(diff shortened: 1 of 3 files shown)


joncampbell123:
The MDA/CGA/EGA/Tandy/etc modes render 8bpp with a color palette.Does Direct3D 11 have any consideration for anything using a color palette?

maron2000:
Th current code requires pallete to be converted to 32bpp.
The problem is that nothing is sent to the buffer so the screen is totally blanked.

PR #6104, opened 2026-02-07
D3D11: Consider unsupported OS

maron2000:
Direct3D11 is not available (or limited support) on old OS, so add check for availability to avoid crashes.


joncampbell123:
There you go. :+1: Just like the support for Windows 7 taskbar enhancement interfaces in the code!

PR #6132, opened 2026-02-27
Experimental: Implement Metal rendering (SDL2 only)

maron2000:
Add experimental Metal output backend for macOS.
Since Apple considers OpenGL deprecated, this adds a Metal-based rendering option to prepare for future compatibility.

image

### DIFF
--- a/src/gui/menu.cpp
+++ b/src/gui/menu.cpp
@@ -401,7 +401,7 @@ static const char *def_menu_video_output[] =
 #if C_GAMELINK
     "output_gamelink",
 #endif
-#if ß && C_METAL
+#if defined(MACOSX) && defined(C_SDL2) && C_METAL
     "output_metal",
 #endif    
     "--",


joncampbell123:
Apparently Visual Studio, at least for Visual Studio, doesn't like unicode characters in preprocessor tokens. Can you change this please? Perhaps instead of using the "Beta" character, name it BETA.

image

src/gui/menu.cpp
```diff
@@ -401,6 +401,9 @@ static const char *def_menu_video_output[] =
 #if C_GAMELINK
     "output_gamelink",
 #endif
+#if ß && C_METAL
```
joncampbell123: Why is there a "ß" here?

Is this a mistake?

If you're trying to block this out, use #if 0 && C_METAL instead.

maron2000:
Yeah, I think it's eszett of German but I don't know how to type that letter so strange why this happened.
Fixed.

PR #6244, opened 2026-04-28
Set int 13 enable 48-bit LBA to true in configuration

BridgeHeadland:
I set 48-bit LBA enabled as default, that's what at least prevents Windows XP from crashing while the cycles are over max and the VHD fills up to (as I see it) around 7-8 GB.


BridgeHeadland:
I forgot that I had to edit src/dos/drives.cpp as well, something I don't have access to, so I'm closing this now. Sorry so much.

PR #6253, opened 2026-05-02
TTF mode: Limit window dimensions only on windowed mode

maron2000:
Commit b67984c limited the windows size of TTF output to a reasonable size, however, the fontsize of fullscreen mode was limited as well.
This PR disables the limitation when in fullscreen mode.

Tested on VS x64 SDL2 and macOS SDL1.


joncampbell123:
Good catch :+1:

PR #6261, opened 2026-05-06
CD Audio: Fix track not advancing at end of playback

maron2000:
This PR fixes a flaw in CD Audio that when a track reaches to an end, it didn't advance to the next track.
(Possibly when the each audio tracks are independent files.)


maron2000:
The nightly build somehow failed, but it all succeeded i  my repository so it is reeady to be merged.

PR #6281, opened 2026-05-11
Fix logging of 'Packed file is corrupt' message detection

### LINKED ISSUE #6280: "Fix detection of EXEPACK failure" breaks certain writes
### Describe the bug

Commit https://github.com/joncampbell123/dosbox-x/commit/6e3ecfaa7405c104790e3a32cb4d1e1c742736bc does not function correctly. It currently matches any substring of `Packed file is corrupt` as a valid error message, including simply outputting the letter `P` by itself.

1. Start DOSBox-X
2. Type `dos4gw` and press enter
3. Notice the title text is wrong, `DOS/4G rotected Mode Run-time`

### Expected behavior

`DOS/4G Protected Mode Run-time`

Arch Linux

all commits from 6e3ecfaa7405c104790e3a32cb4d1e1c742736bc to current

### Used configuration

```ini

```

### Output log

```text

```

### Additional information

_No response_

maron2000:
Fixed matching 'Packed file is corrupt' message was not working as intended.

Fixes #6280

image

image

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -79,6 +79,7 @@ Next version
     FAT filesystem driver, and then a drive letter. At this time, this
     is enough for Microsoft RAMDRIVE.SYS to load and appear as a working
     drive letter in DOSBox-X (joncampbell123)
+  - Fixed logging of 'Packed file is corrupt' message detection (maron2000)
 
 2026.05.02
   - DOSBox-X will now check both the current directory and the .config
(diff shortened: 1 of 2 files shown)


joncampbell123:
A commit in the last two months adds code in DOS_Execute() to detect and directly handle EXEPACK compressed EXE files. When enabled, it decompresses it directly, which avoids the Packed File is corrupt message entirely.

Perhaps in a future commit, if that EXEPACK code is enabled, emulation should bypass this check entirely?

maron2000:
Thanks for your comment.
Yes, the log is for a "just in case"; so the EXEPACK code is mature it's a good idea to bypass logging.

PR #6298, opened 2026-05-19
IMGMOUNT: Attach CD-ROM to IDE controller

maron2000:
This PR fixes that the mounted CD-ROM sometimes not attached to an IDE slot, which can result in cases where the software cannot find the drive.

For example, if CD-ROM drive is mounted by dropdown menu, Windows 2000 setup fails to find the drive and results in BSOD.

### DIFF
--- a/src/dos/dos_programs.cpp
+++ b/src/dos/dos_programs.cpp
@@ -7285,7 +7285,7 @@ class IMGMOUNT : public Program {
 		}
 
 	public:
-		bool MountIso(const char drive, const std::vector<std::string> &paths, const signed char ide_index, const bool ide_slave) {
+		bool MountIso(const char drive, const std::vector<std::string> &paths, signed char ide_index, bool ide_slave) {
 			//If mounting while a guest OS is active, you may ONLY replace an ISO with an ISO! You may not mount a new drive!
 			if (dos_kernel_disabled) {
 				if (Drives[drive - 'A']) {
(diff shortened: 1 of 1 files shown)


joncampbell123:
Is this intended for mounting a CD-ROM image **before** or **after** BOOTing into a guest OS?

IDE devices aren't hot swappable while the system is running. You can't just plug in a CD-ROM drive while the system is running and expect it to work even if you do if before the guest OS starts up.

I don't mind if that's done in the native DOSBox-X DOS environment, but once you start Windows 95, removing and adding IDE devices should never happen (except for future PCMCIA IDE emulation, but that's a future discussion).

joncampbell123:
Consider that PCMCIA CF adapters common in the early 2000s were basically an IDE controller on a card that directly connected to a CompactFlash adapter, and Windows 95 seems to support those quite well.

Another example of PCMCIA IDE was an old laptop I found in the early 2000s at a PC Recycle that had a PCMCIA CD-ROM drive, which turned out to be an IDE drive in an enclosure connected to a PCMCIA card.

maron2000:
The code is intended for mounting BEFORE booting the guest.
I thought that mounting a new drive on a booted guest is rejected before hand, from the following comment.
Corrext me if I'm wrong.

```
//If mounting while a guest OS is active, you may ONLY replace an ISO with an ISO! You may not mount a new drive!
```

maron2000:
The code is about `bool MountIso()` function, so I really don't know what PCMCIA cards are related here.

joncampbell123:
> The code is about `bool MountIso()` function, so I really don't know what PCMCIA cards are related here.

I was rambling on about PCMCIA, sorry.

maron2000:
No problem, we can discuss anything to improve.

I can add an additional condition to ensure not to attach a new drive to the controller when a guest OS is booted.

PR #6319, opened 2026-06-01
Replace Nuked-OPL3 with Nuked-OPL3-fast

tgies:
Swap the vendored Nuked-OPL3 (nukedopl.cpp/.h, upstream 730f8c2) for [Nuked-OPL3-fast](https://github.com/tgies/Nuked-OPL3-fast/), a performance fork with bit-for-bit identical output that runs 1.4x-2.2x faster. ~~Also advances the core algorithm to upstream cfedb09, which includes two fixes to envelope inaccuracies that upstream 730f8c2 lacked.~~ Disregard previous sentence: [nukedopl3.cpp has a misleading comment](https://github.com/joncampbell123/dosbox-x/blob/7056e1099dd58bdb29a2b96df2278384eefd9452/src/hardware/nukedopl.cpp#L32) that led me to believe you had vendored an older version.

N/A

No

~~Depends on your definition: API is strictly the same, behavior strictly the same as upstream, but dosbox-x was missing 2 fixes from upstream Nuked-OPL3 that materially changed envelope behavior, so output will differ very, very slightly from the version of Nuked-OPL3 dosbox-x had been using.~~ Disregard prior sentence: this was a misread.

## Additional information

To be perfectly clear: this is not a quality-vs-performance tradeoff, this is a pure performance gain by way of a better optimized path through the same algorithm.

Nuked-OPL3-fast is at https://github.com/tgies/Nuked-OPL3-fast/

Already adopted by [dosbox-staging](https://github.com/dosbox-staging/dosbox-staging/pull/4910), [86box](https://github.com/86Box/86Box/pull/7214), and [libADLMIDI](https://github.com/Wohlstand/libADLMIDI/commit/6528dafce641e81ded5d62809445267b82dc0d8d).


maron2000:
Although not important, which commit of the original code do you refer to as missing?
The last related update is PR #5085 which should have updated to the latest commit.

joncampbell123:
Should I merge this now or wait until after release?

tgies:
> Although not important, which commit of the original code do you refer to as missing? The last related update is PR #5085 which should have updated to the latest commit.

Because [the current code in dosbox-x](https://github.com/joncampbell123/dosbox-x/blob/7056e1099dd58bdb29a2b96df2278384eefd9452/src/hardware/nukedopl.cpp#L32) claims in a comment that it was sourced from that older commit, and I trusted it and didn't look closely enough to notice it was actually latest upstream. So: there will be zero output changes. Apologies for confusion.

tgies:
Forcepushed solely to update misleading commit message now that it's been pointed out to me that dosbox-x was not in fact using an older pre-envelope-fixes version of Nuked. Thanks @maron2000.

> Should I merge this now or wait until after release?

I assume you're conferring with other maintainers, not asking me, but just in case: from my end, it's merge-ready.

PR #6326, opened 2026-06-03
Fix PMULLW signedness issues

bachimusprime:
PMULLW is signed through and through, as found by cross-referencing with PCBox's code.


bachimusprime:
The CI failures aren't related to this change, it seems.

PR #6327, opened 2026-06-04
Fix FPU_Reg_80 size 

cimarronm:


joncampbell123:
I'm sorry, I just made some changes that conflict badly with your pull request.

Are the changes still needed?

cimarronm:
I'm not 100% sure this is required to ensure the padding is at the very end on little and big endian systems but it does avoid any confusion on the actual size of the 80-bit FPU register.

PR #6345, opened 2026-06-10
Fix build error of menu related changes

maron2000:
Fix build error since commits around 6926a40.

Seems the menu shows as intended after the fix.
image


joncampbell123:
Sorry about that. The hide() method only works on Linux at this time.

PR #6346, opened 2026-06-10
Use windows-2022 image for ARM64 build

maron2000:
The latest CI runner image is now under deployment and availability of VS build tool v143 is unstable, making building ARM64 builds fail.
This PR switches to windows-2022 image at the moment, where we can stably assume v143 to be available.

The problem is already reported to github CI runner devs, so that once it is solved, we will resume using latest runner.
https://github.com/actions/runner-images/issues/14215


maron2000:
#6345 has to be merged first or the non-ARM Windows related builds will fail.

PR #6364, opened 2026-06-19
Fix graphical printouts corrupted

maron2000:
Fixed graphical data not to be handled as printing command or text characters. 

Figure below is a printout test using JWCAD. 

page6


maron2000:
Also fixed a misalignment of DBCS characters

page11

PR #6411, opened 2026-07-12
Add support for nested subdirectories in Z drive

maron2000:
This PR enables support for nested subdirectories in Z drive.

## What issue(s) does this PR address?
Fixes #6383

image

A deeper subdirectory
image

PR #6413, opened 2026-07-13
Fix build error of XP VS builds

maron2000:
Fix build error of XP Visual Studio builds, since PR #6411.
The builds are successful for all other platforms.


maron2000:
The build completed successfully in my repository, so this appears to have been an intermittent CI failure rather than an issue with the changes themselves.

https://github.com/maron2000/dosbox-x-1/commit/46eff0d691277246bc38927201bc02b15bd2ab40

PR #6421, opened 2026-07-20
Fix an exception in drive_virtual.cpp

zhongjianhua163:
我很困惑， 为什么第236行在对current_normalized_dir的尾部取值前进行了空字符串的判断，而在第235行中对头部进行取值时却没有判断空字符串？？

我在首次测试中就遇到此处触发的异常。由于DEBUG版在遇到此异常时无法继续，所以我添加了空字符串的判断。

### DIFF
--- a/src/dos/drive_virtual.cpp
+++ b/src/dos/drive_virtual.cpp
@@ -232,7 +232,7 @@ void VFILE_Register(const char* name, uint8_t* data, uint32_t size, const char* 
     static unsigned int last_resolved_onpos = 0;
 
     std::string current_normalized_dir = dir_str;
-    if(current_normalized_dir.front() == '/' || current_normalized_dir.front() == '\\') current_normalized_dir.erase(0, 1);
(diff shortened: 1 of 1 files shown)


joncampbell123:
Google translate:

"I'm confused. Why does line 236 check an empty string before taking a value for the current_normalized_dir's tail, but not an empty string when taking the head value in line 235? I encountered an exception triggered here during my first test. Since the DEBUG version cannot continue when encountering this exception, I added an empty string check."

zhongjianhua163:
> Google translate:
> 
> "I'm confused. Why does line 236 check an empty string before taking a value for the current_normalized_dir's tail, but not an empty string when taking the head value in line 235? I encountered an exception triggered here during my first test. Since the DEBUG version cannot continue when encountering this exception, I added an empty string check."

“but not an empty string ...” -> "but not check an empty string ..."   maybe, im not sure, haha

PR #6490, opened 2026-08-28
Fix for OpenGL, fullscreen and modeswitching positional offsets.

DieSkaarj:

When going fullscreen with modesetting enabled the surface would resize to the reported size from SDL. This created issues where the size of the surface was often larger than the resolution requested. This PR fixes this by calling the new width and height from the vga_crtc device. It also adds a fix for the menu bar height being added to the surface size too.


maron2000:
May I ask why there are two output_opengl.cpp‎ files?

DieSkaarj:
My apologies it was a skill issue, with git.

maron2000:
I think you can just delete the one in the root directory.
Or create a new branch from the current original `master` and commit only the required change and re-submit the PR.
The latter us cleaner since you don't add and delete the unnecessary file.

Issue #6491, opened 2026-08-29
IMGMOUNT crashes DOSBox-X


GaMartinelli:
## Environment
- DOSBox-X: 2026.08.02-1.el9.x86_64
- OS: Rocky Linux 9
- Architecture: x86_64
- GDB: 16.3-3.el9
- glibc: 2.34
- SDL2: 2.26.0
- DOSBox-X installed as /usr/bin/dosbox-x via RPM package (Copr repo for DOSBox-X owned by rob72)
- Configuration file: any

## Problem
`IMGMOUNT` crashes DOSBox-X with: `*** buffer overflow detected ***: terminated`

This happens even when IMGMOUNT is invoked with no arguments, so the image file itself is not involved.

### Steps to reproduce the behaviour

1. Start DOSBox-X 2026.08.02 on Rocky Linux 9.
2. At the DOS prompt run: `IMGMOUNT`
3. DOSBox-X aborts with: `*** buffer overflow detected ***: terminated`

### What operating system(s) this bug have occurred on?

Rocky Linux 9

### What version(s) of DOSBox-X have this bug?

2026.08.02

### Used configuration

```ini

```

### Output log

```text
LOG: Early LOG Init complete
LOG: "" is not a valid value for variable: configuration tool theme.
It might now be reset to the default value: 
LOG: "" is not a valid value for variable: configuration tool theme.
It might now be reset to the default value: 
LOG: CONFIG: Loaded config file: /XXX/.config/dosbox-x/dosbox-x-2026.08.02.conf
LOG: DOSBox-X's working directory: /XXX
LOG: Logging init: beginning logging proper. This is the end of the early init logging
LOG: Logging: No logfile was given. All further logging will be discarded.
LOG: DOSBox-X version 2026.08.02 Commit e522642 (Linux SDL2 64-bit)
LOG: Prevent capture: 0
LOG: SDL: version 2.26.0, Video x11, Audio pulseaudio
LOG: Host keyboard layout is now  ()
LOG: Mapper keyboard layout is now  ()
LOG: SDL2 reports desktop display mode 3440 x 1440
LOG: The default output for the video system: opengl
LOG: OpenGL Version : 4.6 (Compatibility Profile) Mesa 25.2.7
LOG: Configured windowposition: -
LOG: SDL: Current window pixel format: SDL_PIXELFORMAT_RGB888
LOG: SDL: You are running in 24 bpp mode, this will slow down things!
LOG: Releasing keyboard
[...]

PR #6492, opened 2026-08-29
dos_execute: fix potential non-null-terminated string passed to SetFileName()

cimarronm:
Fixes a bug in DOS_Execute() where the 8-character MCB filename buffer (stripname) was not guaranteed to be null-terminated: if the stripped name was exactly 8 characters with no '.' or embedded null, memset(&stripname[index], 0, 8-index) was a no-op, leaving the buffer without a terminator. DOS_MCB::SetFileName() takes a const char * and calls strlen() on it, so this could read past the end of the 8-byte stack buffer into adjacent stack memory.


cimarronm:
Relates to and should fix #6491

PR #6499, opened 2026-08-31
parport+serialport: local named-pipe / AF_UNIX transports (extlpt, nullmodem sock:2)

WinLin97:
Adds two low-latency, same-host ways to connect a DOS guest port to an
external program, without a TCP hop or any network stack: a Windows named
pipe or a POSIX AF_UNIX stream socket.

The use case is talking to a device that DOSBox-X does not model itself -
custom or uncommon hardware on the LPT or COM lines, or an emulation of it -
from a helper process that implements whatever protocol the guest software
expects. Useful for homebrew work and for testing DOS software against such
a device with no physical hardware.

## parallel1=extlpt - raw LPT register passthrough

Every guest access to the DATA/STATUS/CONTROL registers becomes a
synchronous 4-byte request / 4-byte response to the peer; the guest's next
instruction does not run until the peer has answered (or a 2s timeout
elapses, after which the port falls back to the printer engine / 0xFF).
Blocking is deliberate - the guest's device protocol assumes electrical
immediacy, and a write that returned early would let the guest CPU outrun a
non-real-time helper and read stale input lines. The backend has no
knowledge of any device protocol; the helper makes sense of the bytes.

```
parallel1=extlpt transport:pipe pipe:/tmp/mydev
parallel1=extlpt transport:tcp host:127.0.0.1 port:5800
```

Options: `transport:pipe` (default) | `transport:tcp`; `server` (DOSBox-X
creates the endpoint instead of connecting to one); `registers:` to wire
only some of DATA/STATUS/CONTROL; `maskdata:`/`maskcontrol:` to skip the
round-trip when no bit the device cares about changed. A one-frame
version handshake gates the connection. The wire protocol is summarised at
the top of `extlpt.cpp`.

`transport:tcp` reuses the serialport socket layer, so it is only built
with modem support (`C_MODEM`); `transport:pipe` has no such dependency.

## serial1=nullmodem sock:2 - named-pipe transport for nullmodem

A third `sock:` transport alongside TCP (`0`) and ENet (`1`). The endpoint
is a path, not host:port:

```
serial1=nullmodem sock:2 pipe:/tmp/mycom      (listen on the path)
serial1=nullmodem sock:2 server:/tmp/mycom    (connect to the path)
```

`NamedPipeClientSocket` / `NamedPipeServerSocket` implement the existing
`NETClientSocket` / `NETServerSocket` interface; `+SOCK` on softmodem
rejects it (no host to dial).

The new socket types are a byte transport only - the RTS/CTS/DTR/DSR/DCD/RI
and break handling in `CNullModem` (the `0xFF`-escaped handshake protocol,
and the `transparent:` switch that disables it) sits above the socket layer
and is unchanged, so `sock:2` carries the modem control lines exactly as
`sock:0` does.

## Shared / supporting changes

- **`src/misc/localpipe.{cpp,h}`** (new) - the pipe/socket connection dance
  (create, connect-with-retry, one-client accept, unlink), plus
  non-blocking recv/send building blocks. Used by both features above.
- **`printer_if.h`** - `PRINTER_StrobeByte()`, the five-register STROBE
  pulse that every `LPTn:`/`PRN` redirection `Putchar()` does, factored out
  so `extlpt` and `printer_redir.cpp` share one copy (same behaviour).
- **`misc_util.cpp`** - `TCP_NODELAY` and a 15s `SO_SNDTIMEO` on the native
  TCP socket. Every caller here (nullmodem, softmodem, `extlpt
  transport:tcp`) is a small request/response protocol: Nagle only adds
  delayed-ACK stalls, and a peer that stops reading should fail the send
  rather than block the emulator core.

## Testing

Full build on Linux (`./configure && make`) and Windows (MSVC / the `vs/`
solution); both pass the checks below.

- **extlpt** - a headless DOSBox-X running a small DOS `.COM` that writes
  and reads 378h/379h/37Ah, with a stub server on the other end, produces
  exactly the expected write/read frame sequence over both `transport:pipe`
  (named pipe / AF_UNIX) and `transport:tcp`; the version handshake and
  clean disconnect work.
- **nullmodem sock:2** - with `CTTY COM1`, the DOS shell's stdin/stdout
  round-trip in both directions over the pipe (this run uses
  `transparent:1`, so it covers the data path, not the handshake-line
  escaping - that code is untouched and shared with `sock:0`); the socket
  file is removed on shutdown.

I can attach the (stdlib-only, cross-platform) Python test scripts as a
follow-up comment if useful.


WinLin97:
Test scripts (Python):
[test_extlpt.py](https://github.com/user-attachments/files/31637099/test_extlpt.py)
[test_serial_pipe.py](https://github.com/user-attachments/files/31637101/test_serial_pipe.py)

joncampbell123:
This is a new feature, not a bugfix, and a release of DOSBox-X will come out in the next 24 hours, so I will hold merging this until after release. Is that OK?

WinLin97:
Yes, it's okay 😃 In this time I'll try repair the HX-DOS build.

WinLin97:
A reference peer (extlpt_peer.py, stdlib-only, protocol + --selftest) is attached. 
Happy to commit it somewhere if you'd rather have it in-tree.

[extlpt_peer.py](https://github.com/user-attachments/files/31676616/extlpt_peer.py)

```
usage: extlpt_peer.py [-h] [--pipe PATH | --tcp [HOST:]PORT | --selftest] [--connect] [--device {latch,sniffer}] [--once] [-v]

extlpt_peer.py - minimal reference peer for the DOSBox-X `extlpt` parallel-port
backend. Copy this file, subclass `Device`, done.

DOSBox-X's `parallel1=extlpt` turns every guest access to the LPT registers
(DATA / STATUS / CONTROL) into a synchronous 4-byte request / 4-byte response
RPC to an external process. This file is that process, boiled down to the
protocol and nothing else - no hardware, no framework. The wire rules never
change, so whatever device you need to fake (a programmer, a transfer cable,
a lab instrument) is just a `Device` subclass.

---------------------------------------------------------------------------
Wire protocol
---------------------------------------------------------------------------
Reliable ordered stream (TCP or a local pipe). No length prefix: every frame
is exactly 4 bytes.

    byte0  opcode
    byte1  register    0 = DATA (base+0), 1 = STATUS (base+1), 2 = CONTROL (base+2)
    byte2  value
    byte3  sequence    echoed back verbatim; a framing tripwire only

    opcode   0x01 WRITE   0x02 READ   0x03 RESPONSE   0x04 ACK
             0x05 HELLO   0xFF ERROR

Exchange:
    DOSBox-X -> OP_HELLO  ver caps 0     (very first frame, once per connection)
    peer     -> OP_HELLO  ver caps 0     ver must match or the connection is refused

    DOSBox-X -> OP_WRITE  reg val seq    guest wrote `val` to `reg`
    peer     -> OP_ACK    reg P   seq    P = 1 if you now have a response queued
                                         (lets DOSBox-X decide whether a following
                                         read belongs to you or its printer engine)

    DOSBox-X -> OP_READ   reg 0   seq    guest is reading `reg`
    peer     -> OP_RESPONSE reg val seq  `val` is handed to the guest

    peer     -> OP_ERROR  reg code seq   fails that transaction (diagnostic only)

DOSBox-X blocks the guest CPU on each exchange, bounded by ~2 s; answer
promptly. Keep PROTOCOL_VERSION in lockstep with extlpt.cpp.

---------------------------------------------------------------------------
Quick start
---------------------------------------------------------------------------
    python extlpt_peer.py --selftest                # no DOSBox-X: exercise the Device
    python extlpt_peer.py --tcp 5700 -v             # listen, decode every frame
    python extlpt_peer.py --pipe \\.\pipe\dosbox_lpt -v
    python extlpt_peer.py --tcp 5700 --device sniffer -v   # log what a DOS program does

On startup it prints the matching `dosbox-x.conf` line for the other side.
It keeps serving reconnects until Ctrl+C (use --once for a single connection).

Topology  (who creates the endpoint)
    DOSBox-X config                                    run this as
    parallel1=extlpt transport:pipe pipe:<p>           --pipe <p>            (listen)
    parallel1=extlpt transport:pipe pipe:<p> server    --pipe <p> --connect
    parallel1=extlpt transport:tcp port:<n>            --tcp [host:]<n>      (listen)
    parallel1=extlpt transport:tcp port:<n> server     --tcp host:<n> --connect

options:
  -h, --help            show this help message and exit
  --pipe PATH           named pipe (\\.\pipe\name) / AF_UNIX socket path
  --tcp [HOST:]PORT     TCP endpoint
  --selftest            drive the Device with a scripted sequence, no DOSBox-X
  --connect             connect to an endpoint DOSBox-X created (DOSBox-X started with the extlpt 'server' option); default: create it and wait
  --device {latch,sniffer}
                        which Device to run (default: latch)
  --once                serve one connection then exit (default: keep serving)
  -v, --verbose         log every raw 4-byte frame, decoded, with handler latency
```

joncampbell123:
@WinLin97 Please do. It's better that the documentation exist somewhere, including in-tree, than disappear into the internet and potentially get lost. Make it a separate pull request please.

PR #6519, opened 2026-09-04
MCP Protocol

caiiiycuk:
## Summary

This adds a small, optional control channel to the built-in debugger so an external
process can drive it: `src/debug/debug_mcp.cpp` (~700 lines) plus three integration
points in `debug.cpp` / `debug_gui.cpp`.

When the debugger starts, DOSBox-X opens an outbound TCP connection to
`127.0.0.1:58991` and speaks a line-oriented text protocol:

```
REQ <id> PING                  ->  BEGIN <id> OK
                                   PONG
                                   END <id>

REQ <id> BREAK                 ->  enters the debugger

REQ <id> EXEC <debugger cmd>   ->  BEGIN <id> OK
                                   <captured debugger output>
                                   END <id>
```

`EXEC` runs the string through the very same `ParseCommand()` used by the
interactive debugger prompt, and the resulting `DEBUG_ShowMsg` output is captured
and returned. 

The counterpart is a separate project, [dosbox-x-mcp-server](https://github.com/caiiiycuk/dosbox-x-mcp-server), which exposes this to LLM agents as MCP tools.

### How to use it

`dosbox-x-mcp-server` is a Rust program that works on all major platforms. Just register it as an MCP server in OpenCode and start a session.

From that point on, whenever you launch DOSBox-X, it automatically connects to the MCP server, allowing you to interact with and control the running DOSBox-X instance directly through OpenCode.

I originally wrote this for the js-dos project to help diagnose and fix issues in games that were difficult to solve manually. It works really well — in my cases, it did exactly what I wanted it to do.

The same approach also works in browser environments.

---

Below, you can read an overview and comparison with the recently merged PR. I think the approaches are very similar, but I don't know Chinese well enough to fully understand how to use that implementation.

Also, for my particular use case, only this PR allows me to control a js-dos instance running in the browser.

I think it makes sense to keep both options.

---

## Design philosophy

The guiding idea is that **the emulator should ship a transport, not an API.**

- **No new debugger semantics.** There is no session model, no state machine, no
  capability negotiation, no schema. Whatever the interactive debugger can do
  today, an MCP server can do today; whatever gets added to the debugger tomorrow
  works over this channel with no changes here. The maintenance cost does not grow
  as the debugger grows.

- **Policy belongs outside the emulator.** Tool descriptions, JSON schemas, retry
  and timeout policy, prompt-facing naming, and everything else that is really
  *LLM tooling* lives in the MCP server. That layer changes on a weekly cadence;
  an emulator does not. Keeping it out of tree means DOSBox-X never has to track
  the churn of the agent ecosystem.

- **Text is a feature, not a shortcoming.** The debugger's output is what a human
  reverse engineer reads, and reading exactly that kind of text is what language
  models are unusually good at. A per-field JSON contract is something a *program*
  needs; a model does not need `AX` in its own key to understand `AX=0000`. Where
  a client does want structure, it can parse on its side, at its own risk, without
  the emulator freezing a format into a public contract.

- **Debuggable by hand.** The wire format is plain lines: `nc 127.0.0.1 58991`
  and you can drive the whole thing manually. No pipes, no framing, no codegen.

- **The emulator connects out; it never listens.** DOSBox-X is the TCP *client*.
  There is no listening socket to firewall, no port to secure, no unauthenticated
  server inside the emulator process. If nothing is listening, DOSBox-X just keeps
  running and retries in the background.

- **Small enough to actually review.** One file, three integration points, all of
  it compiled out unless `C_DEBUG` and SDL_net are both available.

## What changed

| File | Change |
| --- | --- |
| `src/debug/debug_mcp.{cpp,h}` | New. I/O thread, reconnect loop, request parsing, output capture. |
| `src/debug/debug.cpp` | `DEBUG_ExecuteCommand()` wrapper around `ParseCommand()`; start/stop with the debugger; `ControlServer_Poll()` inside `DEBUG_Loop()`. |
| `src/debug/debug_gui.cpp` | Capture hook in `DEBUG_ShowMsg`; pager suppressed while capturing. |
| `README.debugger` | Protocol documentation. |
| `src/debug/Makefile.am`, `vs/dosbox-x.vcxproj*` | Build files. |

Threading: all socket I/O happens on a dedicated thread; commands are only ever
executed on the emulation thread (via a tick handler, and via `DEBUG_Loop()` while
the debugger is stopped). Queues are bounded, the receive buffer is capped, and
`\r`/`\n` are sanitized out of protocol lines.

Suppressing the pager during capture matters in practice: without it, `HELP`,
`GDT`, `LDT`, `IDT` and the DOS memory listings block waiting for a keypress and
hang the channel.

Resume commands (`RUN`, `RUNWATCH`, `VRT`) are acknowledged *before* DOSBox-X
leaves the debugger loop, so the client gets an accepted-response rather than a
timeout. That response is an acknowledgement, not a post-resume state snapshot.

## Relationship to the agent RPC interface (#6512)

#6512 landed a much larger structured interface. The two solve the same problem
from opposite ends, and this section is an honest comparison rather than an
argument that one should replace the other.

**This PR — pros**

- ~700 lines in one file, three integration points; nothing added to `Normal_Loop`,
  the shell, or `control.h`.
- Works everywhere DOSBox-X works: SDL_net is available on every target, and the
  feature compiles out cleanly where it is not.
- Full debugger command surface immediately, including commands added later.
- No listening socket, no new command-line options, no config file format.
- Wire format inspectable with `nc`/`telnet`.

**This PR — cons**

- The client parses human-readable debugger text; that format is not a stable
  contract and can change upstream at any time.
- Port `58991` is hardcoded — no `dosbox.conf` setting or command-line switch yet.
- Synchronous request/response only. There is no operation/`wait` model, no
  timeout negotiation, and no asynchronous breakpoint-hit event (the send-event
  helper exists but is not wired up yet).
- "One request in flight" is a convention the client must honour, not an
  invariant enforced by the code.
- No bundled client library or test suite; the client lives in a separate repo.

**#6512 — pros**

- A real structured contract: registers as an object, `memory.read` reporting a
  failing offset, breakpoint handles, trace pages with cursors and per-step
  register diffs.
- Asynchronous model with `operation_id` + `execution.wait`, and a debugger-stop
  listener, so the client is not polling.
- `EmulationThreadQueue` is a cleaner marshalling primitive than a tick handler.
- Named pipe transport, per-request limits, config file, and a `--agent-self-test`
  mode.
- Ships a Python client, unit tests, integration scripts, and extensive docs.
- Genuinely useful debugger-core additions (`CBreakpoint::DeleteBreakpoint`,
  `ConsumeLastTriggered`, agent step-over, resume-after-terminate).

**#6512 — cons**

- Windows only in practice: `UnixSocketTransport::Start()` returns *"outside the
  Windows v1 scope"*, and `C_DOSBOX_AGENT` is defined only in
  `vs/dosbox-x.vcxproj` — the autotools build compiles `libagent.a` but every
  integration point stays disabled.
- ~10.5k lines across 72 files, with `agent_server.cpp` alone at 3136 lines and a
  hand-rolled JSON parser.
- Wider blast radius: `Normal_Loop`, `shell_cmds.cpp`, `control.h`,
  `include/debug.h`, plus unrelated `PlatformToolset v142 -> v143` bumps in the
  bundled zlib/libpng/SDL/freetype projects.
- Rigid session model: a single session, a fixed stop at entry, DOS-filename-only
  targets, a controlled mount — attaching to an already-running machine is not
  possible.
- `trace.*` requires `C_HEAVY_DEBUG`, and register data still comes from parsing a
  strict textual form of the debugger output.
- Primary documentation is in Chinese, which is a barrier for most reviewers.

## Why the two can coexist

They occupy different layers and different build configurations, and this branch
is already rebased on top of #6512 with both paths live:

- Build flags are independent: the agent interface is gated on `C_DOSBOX_AGENT`
  (MSVC only today), this channel on `C_DEBUG` + SDL_net. Either, both, or neither
  can be enabled.
- The one place they truly overlap is debugger output capture, and both capture
  paths now sit side by side in `DEBUG_ShowMsg`, `DEBUG_BeginPagedContent` and
  `DEBUG_EndPagedContent`.
- In `DEBUG_Loop()`, `AGENT_BridgePump()` runs first, then `ControlServer_Poll()`.
- Neither owns global debugger state exclusively; they are separate transports in
  front of the same debugger.

They also serve different users. The agent interface is the right tool for a
deterministic, scripted, reproducible analysis pipeline on Windows. This channel
is the right tool for an LLM that wants the debugger as a human uses it, on any
platform, with nothing to keep in sync as the debugger evolves.

## Known limitations / possible follow-ups

- Make the port configurable (`dosbox.conf` option and/or a command-line switch),
  with `0` meaning disabled.
- Emit asynchronous events on breakpoint hits, using the existing
  `ControlServer_SendEvent()` helper.
- Optionally reject overlapping requests explicitly instead of relying on client
  discipline.

## Testing

- Builds clean on Linux (autotools, GCC, SDL2 + SDL2_net) with no new warnings.
- Exercised end to end against
  [dosbox-x-mcp-server](https://github.com/caiiiycuk/dosbox-x-mcp-server):
  `PING`, `BREAK`, and `EXEC` of the paged commands (`HELP`, `GDT`, `LDT`, `IDT`),
  plus resume commands (`RUN`, `VRT`) and reconnect after the server restarts.

### DIFF
--- a/README.debugger
+++ b/README.debugger
@@ -96,7 +96,7 @@ Debugger MCP control protocol
 
 When the debugger is enabled, DOSBox-X also starts a minimal TCP control
 client for external debugger tooling. This is used by the companion
-dosbox-x-mcp-server project.
+[dosbox-x-mcp-server](https://github.com/caiiiycuk/dosbox-x-mcp-server) project.
 
 DOSBox-X connects to:
 
--- a/README.debugger
+++ b/README.debugger
@@ -142,6 +142,10 @@ interactive debugger prompt. For example:
 returns the same CPU state text that the interactive CPU command would write
 to the debugger output window.
 
+Resume commands (`RUN`, `RUNWATCH`, and `VRT`) are acknowledged before
+DOSBox-X leaves the debugger loop. Their response is only a successful
+acceptance of the command, not a post-resume state snapshot.
+
 Output is captured from DEBUG_ShowMsg while an MCP command is running. Paged
 debugger output is not paused for terminal input during this capture, so
 commands such as HELP, GDT, LDT, IDT, and DOS memory listings can be returned
--- a/README.debugger
+++ b/README.debugger
@@ -151,6 +151,11 @@ debugger output is not paused for terminal input during this capture, so
 commands such as HELP, GDT, LDT, IDT, and DOS memory listings can be returned
 to the MCP client without hanging on the debugger pager.
 
+The MCP control queue is polled both from the normal timer path and directly
+from the active debugger loop. This lets MCP clients issue commands after a
+BREAK has stopped emulation in the debugger; concurrent MCP tool calls are
+still serialized by the MCP server before they reach DOSBox-X.
+
 Unknown commands return ERR and include the normal debugger
 "command not recognized" text. Commands behind build options, such as
 C_HEAVY_DEBUG-only commands, may be unavailable in some builds. MCP clients
--- a/README.debugger
+++ b/README.debugger
@@ -91,6 +91,69 @@ DEBUGBOX [command] [options]
 You can also type DEBUGBOX without a parameter to start the
 debugger.
 
+Debugger MCP control protocol
+-----------------------------
+
+When the debugger is enabled, DOSBox-X also starts a minimal TCP control
+client for external debugger tooling. This is used by the companion
+dosbox-x-mcp-server project.
+
+DOSBox-X connects to:
+
+  127.0.0.1:58991
+
+The external MCP server must be started first and listen on that address.
+If the server is not available, DOSBox-X keeps running normally and retries
+the connection in the background. If the connection is lost, DOSBox-X will
+try to reconnect.
+
+The control channel is intentionally line-oriented text, not JSON. The MCP
+server sends one request line:
+
+  REQ <id> PING
+  REQ <id> BREAK
+  REQ <id> EXEC <debugger command>
+
+DOSBox-X replies with:
+
+  BEGIN <id> OK
+  <zero or more output lines>
+  END <id>
+
+or:
+
+  BEGIN <id> ERR
+  <one or more error lines>
+  END <id>
+
+Only one debugger command is expected to be pending at a time. The MCP
+server serializes calls before sending them to DOSBox-X.
+
+PING returns PONG and is used to check the active connection.
+
+BREAK enters the built-in debugger, equivalent to invoking the debugger
+break action.
+
+EXEC runs one existing debugger command through the same parser used by the
+interactive debugger prompt. For example:
+
+  REQ 7 EXEC CPU
+
+returns the same CPU state text that the interactive CPU command would write
+to the debugger output window.
+
+Output is captured from DEBUG_ShowMsg while an MCP command is running. Paged
+debugger output is not paused for terminal input during this capture, so
+commands such as HELP, GDT, LDT, IDT, and DOS memory listings can be returned
+to the MCP client without hanging on the debugger pager.
+
+Unknown commands return ERR and include the normal debugger
+"command not recognized" text. Commands behind build options, such as
+C_HEAVY_DEBUG-only commands, may be unavailable in some builds. MCP clients
+should call HELP for the live command list from the running DOSBox-X build,
+or use their own static command catalog as a hint before falling back to raw
+EXEC.
+
 Debugger keyboard shortcuts
 ---------------------------
 
--- a/README.debugger
+++ b/README.debugger
@@ -94,18 +94,24 @@ debugger.
 Debugger MCP control protocol
 -----------------------------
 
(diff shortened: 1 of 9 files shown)


joncampbell123:
Can the user disable the server from dosbox.conf?

Better yet, is the server off by default?

Let's not listen to ports by default.

caiiiycuk:
> Can the user disable the server from dosbox.conf?
> 
> Better yet, is the server off by default?
> 
> Let's not listen to ports by default.

I agree, will do.

joncampbell123:
Let me know when this is ready to merge.

caiiiycuk:
@joncampbell123 done

caiiiycuk:
@joncampbell123 let me know if I need to fix also this failing tests, thank you

PR #6531, opened 2026-09-07
Fix build error of Windows low-end and HX-DOS builds

maron2000:
This PR fixes build errors on Windows Low-end and HX-DOS by preventing RPC agent-related files from being compiled when the feature is disabled.

The RPC agent code introduced in PR #6512 is currently incompatible with these two platforms.

Nightly builds for all platforms are currently built with the RPC agent feature disabled. Re-enabling the feature will be attempted once all platforms are buildable again.

### DIFF
--- a/src/agent/bridge/agent_bridge.cpp
+++ b/src/agent/bridge/agent_bridge.cpp
@@ -1,3 +1,4 @@
+#if defined(C_DEBUG) && defined(C_DOSBOX_AGENT)
 #include "agent/agent_bridge.h"
 
 #include <atomic>
@@ -259,3 +260,5 @@ bool AGENT_RunQueueSelfTest(std::string* error)
 }
 
 } // namespace dosbox_agent
+
+#endif // defined(C_DEBUG) && defined(C_DOSBOX_AGENT)
--- a/src/agent/debugger/debugger_adapter.cpp
+++ b/src/agent/debugger/debugger_adapter.cpp
@@ -1,3 +1,4 @@
+#if defined(C_DEBUG) && defined(C_DOSBOX_AGENT)
 #include "dosbox.h"
 #include "agent/debugger_adapter.h"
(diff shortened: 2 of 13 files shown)


maron2000:
Since the RPC feature uses CancelIoEx() which is only supported on Vista and after, I think the feature needs to be always disabled for MinGW Low-end and HX-DOS builds.

joncampbell123:
There's a merge conflict with the master branch that I'm not certain how to resolve properly, please look into it.

maron2000:
It was a race condition with #6519, maybe I need help from @caiiiycuk.

caiiiycuk:
@maron2000 Hi, I believe MCP Protocol (#6519) can work just fine with HX-DOS and Windows low-end. I resolved conflict but don't know how to make a PR into your branch (without merging whole master). 

Resolved files: 
[debug.cpp](https://github.com/user-attachments/files/31953116/debug.cpp)
[debug_gui.cpp](https://github.com/user-attachments/files/31953120/debug_gui.cpp)

The fix is simple, just split `if` conditions with both agent/mcp into two `if` blocks, agent should be under define, but MCP part is valid, so no define needed

maron2000:
@caiiiycuk 
Thanks for looking into this. And fixed the conflicts considering your suggestion.

I was just wondering that some places the MCP related functions are called without checking whether it is enabled by `mcp_server` option.
e.g. `ControlServer_Poll()` in `DEBUG_loop()` etc.
I haven't looked into it so deeply but just in case.

caiiiycuk:
Yeah, when I was developing it, I thought about this, but decided it was a normal part of debugging and didn’t need an extra flag. Anyway, we can wrap it behind a flag if needed.

maron2000:
Thanks for your comment. If you decide to fix something, please raise another PR.
This PR is to fix the build errors of HX-DOS and Windows Low-end builds.

PR #6570, opened 2026-09-18
Pro Audio Spectrum, Pro Audio Spectrum Plus and Pro Audio Spectrum 16 support

firevaran:
Pro Audio Spectrum, Pro Audio Spectrum Plus and Pro Audio Spectrum 16 support


LeonG-ZA:
Awesome changes, I've picked up a few things:

pas.cpp is in src/hardware/Makefile.am but not in vs/dosbox-x.vcxproj / .filters, so the MSVC build fails with unresolved PAS_Init / PAS_ShutDown.

With sbtype=pas, anything that only uses 388h/389h (the plain AdLib driver, ADL.DRV in the Sierra games) plays FM music in the left channel only. 
Correct usage for the original PAS FM split switch in bit 7 of B88h (the audio mixer control register):
bit 7 set: FM split off. 388h/389h go to both OPL2s, so the card is AdLib compatible (mono in both channels).
bit 7 clear: FM split on. 388h drives the left OPL2 and 38Ah the right (stereo).

The FM music is a bit soft, I'm not sure if it's normal or not.

firevaran:
> Awesome changes, I've picked up a few things:
> 
> pas.cpp is in src/hardware/Makefile.am but not in vs/dosbox-x.vcxproj / .filters, so the MSVC build fails with unresolved PAS_Init / PAS_ShutDown.
> 
> With sbtype=pas, anything that only uses 388h/389h (the plain AdLib driver, ADL.DRV in the Sierra games) plays FM music in the left channel only. Correct usage for the original PAS FM split switch in bit 7 of B88h (the audio mixer control register): bit 7 set: FM split off. 388h/389h go to both OPL2s, so the card is AdLib compatible (mono in both channels). bit 7 clear: FM split on. 388h drives the left OPL2 and 38Ah the right (stereo).
> 
> The FM music is a bit soft, I'm not sure if it's normal or not.

Thank you for reviewing and testing my changes. All fixed.

Issue #6574, opened 2026-09-19
The 'Enable PS2 mouse event' change make it unable to move the mouse in Windows 3.1x


BridgeHeadland:
### Describe the bug
The mouse is difficult to move from 'Enable PS2 mouse events'/'Merge pull request #6529 from cyberfox1/feature/ps2aux_activate'.

### Steps to reproduce the behaviour

Run one of the latest nightly builds, ranging from 'Enable PS2 mouse events'/'Merge pull request #6529 from cyberfox1/feature/ps2aux_activate' and onward, run Windows 3.1x from there, and move the mouse.

### Expected behavior

I expect to be able move the mouse in Windows 3.1x without issues.

### What operating system(s) this bug have occurred on?

Windows 11 Pro 25H2

### What version(s) of DOSBox-X have this bug?

 'Enable PS2 mouse events'/'Merge pull request #6529 from cyberfox1/feature/ps2aux_activate' and onwards.

### Used configuration

```ini

```

### Output log

```text

```

### Have you checked that no similar bug report(s) exist?

- [x] I have searched and didn't find any similar bug report.

### Code of Conduct & Contributing Guidelines

- [x] I agree to follow the code of conduct and the contributing guidelines.

PR #6577, opened 2026-09-20
Fix Windows 3.1 PS/2 mouse regression

cimarronm:
## Summary

Revert 0c94870d855 ("pass through ps2 aux events and enable irq12 cb"), which regresses the Windows 3.1 PS/2 mouse driver.

## Why this is necessary

DOSBox-X implements two distinct routes for PS/2 mouse input:

- **BIOS PS/2 callback route:** a guest registers an INT 15h, AX=C2h callback. Windows 3.1 uses this route and expects mouse reports through its registered callback.
- **Raw 8042 AUX route:** a guest explicitly enables the keyboard controller AUX interface, reads PS/2 packets from port 60h, and handles IRQ12 itself. This is required by operating systems that access the controller directly.

Before the reverted change, enabling `biosps2` installed the BIOS services and IRQ12 callback but did not enable raw AUX routing. Host mouse input was therefore queued as a mouse event and delivered to the registered BIOS callback.

0c94870d855 calls `KEYBOARD_SetAUXActive(true)` during BIOS PS/2 startup. That sets both `keyb.auxactive` and the controller IRQ12-enable bit. It causes host input to enter `KEYBOARD_AUX_Event()`, which writes raw three- or four-byte PS/2 packets to the keyboard-controller output buffer and triggers IRQ12. The same commit then also queues the BIOS callback event when `useps2callback` is set.

Consequently, a single host mouse action is delivered through both routes. Windows 3.1 does not consume the raw AUX stream when using its BIOS driver, so it receives extra IRQ12s and leaves unsolicited AUX packet bytes pending in the controller buffer. This makes the mouse unresponsive and can destabilize or crash Windows.

The additional callback forwarding is unsafe even when AUX is enabled later by the guest: it duplicates and can reorder input between the raw packet and BIOS callback paths. The controller IRQ12 bit should likewise remain controlled by the guest's keyboard-controller command byte.

## Fix

Restore the previous mutually exclusive routing:

- `biosps2=true` provides BIOS callback support without forcing raw AUX activation.
- Raw AUX packets are produced only after the guest explicitly activates the AUX interface.
- An input event goes through either raw AUX or the BIOS callback path, not both.


cimarronm:
Addresses #6574

cyberfox1:
@cimarronm I didn't realize this change impacted Win 3.1, under DOS the old code path did not work with CuteMouse which expects C2 to activate the AUX path and the INT12 handling, before it reads 60h directly. Maybe there needs to be an explicit configuration option for a CuteMouse-style C2 handler so PS/2 under DOS can work as well.

joncampbell123:
Perhaps in a future change, the aux interface would always be enabled and the default IRQ 12 handler would read from the aux device and process it the same as any BIOS would to make it all work together. That way, if the guest OS takes over IRQ 12 and handles it by itself, it still works properly.

